PR #82, opened 2013-11-19
add support for builtin login to the authorization form

edemocracy:
add support for builtin login to the authorization form
and add AuthorizationLoginView which enables it


outsbart:
Thanks for your pull request. Altough this is an interesting way of solving the problem, I believe we should adopt a more configurable and already proven-to-be-good approach instead of using a non-standard one.

PR #88, opened 2013-12-18
Feature/add list of urls without auth

TomNaessens:
Sometimes, in controlled environments, we want to automaticly authorize people (even if their scopes have changed). This pull request allows the admin to add a list of callback URIs that get automaticly authorized, without any user interaction.


JensTimmerman:
it seems like this was fixed in https://github.com/evonove/django-oauth-toolkit/pull/157 thanks!

PR #146, opened 2014-09-01
Add an index on AccessToken.token to speedup token verification

tchaumeny:
A report on our PostgreSQL database use showed that the most consuming query was this one :

``` SQL
SELECT "oauth2_provider_accesstoken"."id", "oauth2_provider_accesstoken"."user_id", "oauth2_provider_accesstoken"."token", "oauth2_provider_accesstoken"."application_id", "oauth2_provider_accesstoken"."expires", "oauth2_provider_accesstoken"."scope" FROM "oauth2_provider_accesstoken" WHERE "oauth2_provider_accesstoken"."token" = '';
```

This pull request adds an index on the `AccessToken.token` field so that this very frequent query can run much faster.

### DIFF
--- a/oauth2_provider/migrations/0002_auto__add_index_accesstoken_token.py
+++ b/oauth2_provider/migrations/0002_auto__add_index_accesstoken_token.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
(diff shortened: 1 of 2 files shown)


n2ygk:
I'm guessing that if indexeing was manually added to a migration back a long time that it got reverted if it's not in the model definition proper. I see the #1312 addresses this by adding `db_index = True` in the model definition.

Issue #154, opened 2014-10-03
Allow custom schemes in rediect_uri


RodneyRichardson:
The current implementation only supports http, https, and ftp as valid URI schemes. (The validate_uri function has been updated to allow others, but it fails a check by django.http.HttpResponseRedirect later on).

I would suggest adding a new setting (ALLOWED_REDIRECT_URI_SCHEMES), which defaults to to the existing value of ['http', 'https'].

I will create a branch and a pull request for this.
https://github.com/RodneyRichardson/django-oauth-toolkit/tree/allow-custom-uri-schemes

PR #155, opened 2014-10-03
Allow custom uri schemes

RodneyRichardson:
As per [#154](https://github.com/evonove/django-oauth-toolkit/issues/154).


Zopieux:
I started working on my own PR before realizing there was already some efforts being made for this issue (and related ones). Thanks! If this could be merged I could take care of refactoring this regex nightmare into something based solely on `urlparse`, as long as there is no specific reason to use that(?).

PR #157, opened 2014-10-05
Configurable authorization page in AuthorizationView

DavidMuller:
This pull request adds the attribute `skip_authorization_completely` to the AuthorizationView class.  `skip_authorization_completely` defaults to False.  However, if `skip_authorization_completely` is True, the authorization page is skipped and a token is immediately granted/returned.  

For example the following url would always skip the authorization page whenever granting tokens to users:
```python
url(r'^o/authorize_skip_authorization_page/$', AuthorizationView.as_view(skip_authorization_completely=True), name="no_authorization_page"),
```
This PR includes a new test in `test_implicit.py` for this behavior.

### DIFF
--- a/oauth2_provider/tests/test_implicit.py
+++ b/oauth2_provider/tests/test_implicit.py
@@ -1,12 +1,14 @@
 from __future__ import unicode_literals
 
+import mock
+
 from django.test import TestCase, RequestFactory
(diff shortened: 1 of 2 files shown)


masci:
Moved the skip_authorization flag to Application model, your code was merged

PR #168, opened 2014-11-18
Allowed RedirectURIs fixed

dmitry-krasilnikov:
1. There is a problem with allowed redirect URIs check. Redirect URI must start with one of the allowed URIs, it does not have to be fully equal. It is generally accepted practice, for example Slack uses such approach when authorizes clients with OAuth2 protocol. 
2. The second problem is that redirect URIs can contain query strings after '?' symbol, some consumers could use query strings for some sort of reasons. It is better to compare paths without query strings.

I've stumbled on these when I was writing backend for django-social-auth which is used by Sentry.

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -93,7 +93,10 @@ class AbstractApplication(models.Model):
 
         :param uri: Url to check
         """
-        return uri in self.redirect_uris.split()
+        for allowed_uri in self.redirect_uris.split():
+            if uri.startswith(allowed_uri):
+                return True
+        return False
 
     def clean(self):
         from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


oauth2_provider/models.py
```diff
@@ -148,7 +151,7 @@ def is_expired(self):
         return timezone.now() >= self.expires
 
     def redirect_uri_allowed(self, uri):
-        return uri == self.redirect_uri
+        return uri.split('?')[0] == self.redirect_uri.split('?')[0]
```
Zopieux: You might want to use the `urlparse` (or Py3 `urllib.parse`) to do this kind of extraction.

synasius:
This feature is added with the following commit: d17d3ea81beff69871e5d9faa06e6bdaf233c58b

dmitry-krasilnikov:
Thanks! But what about Grant.redirect_uri_allowed method? It has exactly the same bug, because of full URI comparison including query string

synasius:
That check is correct. It only applies to the Authorization Code Grant when the Access Token is requested. See http://tools.ietf.org/html/rfc6749#section-4.1.3

> redirect_uri
>         REQUIRED, if the "redirect_uri" parameter was included in the
>         authorization request as described in Section 4.1.1, and their
>         values MUST be identical.

Issue #151, opened 2014-09-11
get_application_model incompatible with Django 1.7


edemocracy:
get_application_model works fine with Django 1.6.5, but fails with 1.7.
Application = get_application_model() is executed at the top level of the application's model.py

    Traceback (most recent call last):
      File "manage.py", line 12, in <module>
        execute_from_command_line(sys.argv)
      File "<pythonpath>/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
        utility.execute()
      File "<pythonpath>/site-packages/django/core/management/__init__.py", line 354, in execute
        django.setup()
      File "<pythonpath>/site-packages/django/__init__.py", line 21, in setup
        apps.populate(settings.INSTALLED_APPS)
      File "<pythonpath>/site-packages/django/apps/registry.py", line 108, in populate
        app_config.import_models(all_models)
      File "<pythonpath>/site-packages/django/apps/config.py", line 197, in import_models
        self.models_module = import_module(models_module_name)
      File "<pythonpath>/python2.7/importlib/__init__.py", line 37, in import_module
        __import__(name)
      File "<projpath>/myapp/models.py", line 35, in <module>
        Application = get_application_model()
      File "<pythonpath>/site-packages/oauth2_provider/models.py", line 241, in get_application_model
        app_model = get_model(app_label, model_name)
      File "<pythonpath>/site-packages/django/apps/registry.py", line 199, in get_model
        self.check_models_ready()
      File "<pythonpath>/site-packages/django/apps/registry.py", line 131, in check_models_ready
        raise AppRegistryNotReady("Models aren't loaded yet.")
    django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.

masci:
@edemocracy could you provide a snippet of code causing the error?

ilGuccio:
Still this bug in django>=1.9.1
This happens when try to use get_application_model() as foreign key
`models.ForeignKey(get_application_model(), null=True, blank=True) `

PR #186, opened 2015-01-13
Django 1.7 compatibility and SchemedHttpResponseRedirect

Zopieux:
This PR fixes two things:

- With Django 1.7 it becomes impossible to have root-level functions that return a model class, as they're not yet loaded. Thus, the code below will fail. As a temporary workaround, I used monkey-patching within the application registry to make it work. This is quite hacky but I don't see any other way. See: #151 
```python
class MyForm(ModelForm):
  class Meta:
    model = some_func()
```
- HttpResponseRedirect has a very restrictive list of allowed schemes (http, https, ftp) and fails with HTTP 400 for everything else. This prevents mobile (eg. Android) OAuth applications to use custom handlers eg. `my-app-callback://oauth/`. This PR adds a `SchemedHttpResponseRedirect` that allows all the schemes extracted from `redirect_uris`. As such, it remains safe.

The code has been live-tested but there are no unit tests yet.

Feel free to make comments I you want me to change things.

### DIFF
--- a/oauth2_provider/__init__.py
+++ b/oauth2_provider/__init__.py
@@ -3,3 +3,5 @@ __version__ = '0.7.2'
 __author__ = "Massimiliano Pippi & Federico Frenguelli"
 
 VERSION = __version__  # synonym
+
+default_app_config = 'oauth2_provider.apps.OAuth2ProviderConfig'
--- a/oauth2_provider/apps.py
+++ b/oauth2_provider/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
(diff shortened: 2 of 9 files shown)


Zopieux:
Well, I did not notice there were two other PR and one issue for the "redirect" issue.
- Issue #154 is covered by this PR; I don't use any new setting, just allow whitelisted URI schemes – seems legit
- PR #155 is an implementation of the said unwanted new setting
- PR #168 seems a bit too wide (accepts everything that begins with a whitelisted URI) but maybe we can merge it too?

Zopieux:
Okay let's do this the proper way.

PR #187, opened 2015-01-14
Allow arbitrary schemes within HttpResponseRedirect()

Zopieux:
`redirect_uri` is supposed to handle arbitrary schemes as per the RFC; this is especially useful for mobile app handlers eg. `mobile-app://my/callback?code=...`.

To this end, this PR implements `UnsafeHttpResponseRedirect` in `oauth2_provider/views/http.py` that does the same thing as `HttpResponseRedirect` except it does not check the scheme, as it is already done in the various validators.
It's a shame there is no base class in `django.http` to do the Location header stuff without the check, but anyway.

A test has been added to check this bugfix.

See #154 and #155 for discussions regarding the validation of the `redirect_uri` field iself.

### DIFF
--- a/oauth2_provider/tests/test_authorization_code.py
+++ b/oauth2_provider/tests/test_authorization_code.py
@@ -36,7 +36,7 @@ class BaseTest(TestCaseUtils, TestCase):
 
         self.application = Application(
             name="Test Application",
-            redirect_uris="http://localhost http://example.com http://example.it",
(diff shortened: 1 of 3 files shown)


Zopieux:
One Travis job failed because of a Django1.6 tarball wrong checksum.

synasius:
#155 it's now merged!

PR #203, opened 2015-03-08
Make 0.7.2 header the same level as 0.7.0

svetlyak40wt:


svetlyak40wt:
Thank you, @synasius!

PR #213, opened 2015-03-24
use all fields instead of a hard coded list, so that extended models don't require extending the form

### LINKED ISSUE #208: bug and/or documentation issue on extending model - doesn't explain how to utilize extended model
In your advanced topics, you explain how to extend the existing model, which is great!  Docs here:

https://django-oauth-toolkit.readthedocs.org/en/latest/advanced_topics.html

However, doing this breaks the registration form because it's missing these two new fields, and the fields are hard-coded in the RegistrationForm class. Submitting the form therefore fails due to empty fields (since they're not even rendered)

It would be great if the documentation went a step further and explained how to ensure the new fields show up in the RegistrationForm.

Alternatively, it could potentially be a bug.  Since we can set the application model as a setting, perhaps the RegistrationForm could/should derive the fields automatically from the class in that setting?

honestbleeps:
I'm not sure why the fields are hard-coded here, but I suggest using \_\_all\_\_. This way, end users don't necessarily have to extend/override RegistrationForm when using their own extension of AbstractApplication

fixes #208

### DIFF
--- a/oauth2_provider/forms.py
+++ b/oauth2_provider/forms.py
@@ -25,4 +25,4 @@ class RegistrationForm(forms.ModelForm):
     """
     class Meta:
         model = get_application_model()
-        fields = ('name', 'client_id', 'client_secret', 'client_type', 'authorization_grant_type', 'redirect_uris')
+        fields = '__all__'


honestbleeps:
I now see the error of my ways, closing this request. Sorry :-)

PR #214, opened 2015-03-26
add 'Authorization' to 'Vary' header in the middleware

renanivo:
When `OAuth2TokenMiddleware` is used with django's `UpdateCacheMiddleware` and `FetchFromCacheMiddleware`, the response is cached independently of the access token validation.

For example: When the first request has a correct access token, the following requests will retrieve the same response, (Even when the `Authorization` header is removed from the request).

It would happen to a reverse proxy too, depending on its configuration.

This PR adds the `Authorization` to the `Vary` header in the `OAuth2TokenMiddleware`. So that the django cache middleware (or a reverse proxy) can use the `Authorization` header in its cache key.

### DIFF
--- a/oauth2_provider/middleware.py
+++ b/oauth2_provider/middleware.py
@@ -1,4 +1,5 @@
 from django.contrib.auth import authenticate
+from django.utils.cache import patch_vary_headers
 
 
 class OAuth2TokenMiddleware(object):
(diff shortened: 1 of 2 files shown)


renanivo:
I forgot to mention that django's `SessionMiddleware` already adds `Cookie` to the `Vary` header in every response which accessed sessions.

https://github.com/django/django/blob/14f28f82330b3d2ed7f16fa5de1596e50f2a3952/django/contrib/sessions/middleware.py#L36-L37

Issue #219, opened 2015-03-30
0.7.2 to 0.8.0


jairhenrique:
The `skip_authorization` field from `AbstractApplication` is not present on migrations. When I upgrade `django-oauth-toolkit`, I receive this error when try to get token.

```
OperationalError at /oauth/token/
(1054, "Unknown column 'oauth2_provider_application.skip_authorization' in 'field list'")
```

synasius:
That is correct since Application is a swappable model and you should keep those migrations in your app

dustinfarris:
That doesn't make sense.  Why are there migrations at all then?

jairhenrique:
@dustifarris :+1:

synasius:
Ok, I changed my mind on that
the migration has been added here 62847af5c322e74075ff454fffcd824a2ebb2b57

PR #224, opened 2015-04-02
Update migrations

dustinfarris:
A first stab at aligning migrations with recent versions of Django.

Existing migrations are moved to south_migrations.  This means South users will need to be >= 1.0.0 which [looks for this directory](http://south.readthedocs.org/en/latest/releasenotes/1.0.html#library-migration-path), or they will have to add to their settings.py:

```python
SOUTH_MIGRATION_MODULES = {
    'oauth2_provider': 'oauth2_provider.south_migrations',
}
```

I have not extensively tested this on existing projects, but it should be enough to get the conversation started.

Also a great blog post here on supporting new Django migrations in third-party apps: http://treyhunner.com/2014/03/migrating-to-django-1-dot-7/

__Note:__ if you need this right now (for instance if Django is blowing up because you have a custom AUTH_USER model that needs to migrate _before_ oauth2_provider, e.g. if you're seeing things like `django.db.utils.ProgrammingError: relation "myapp_user" does not exist` when you run your tests), you can replicate the new [migrations folder](https://github.com/dustinfarris/django-oauth-toolkit/tree/django-migrations/oauth2_provider/migrations) from my fork into your app as something like `myapp/oauth2_provider_migrations/` and then add to your settings.py:

```python
MIGRATION_MODULES = {
    'oauth2_provider': 'myapp.oauth2_provider_migrations',
}
```

Assuming you've been using oauth2_provider for a while, you will need to fake the migration on your regular db so it doesn't complain about the table already being there in the future:

```
./manage.py migrate --fake oauth2_provider
```


dustinfarris:
__Note:__ if you need this right now (for instance if Django is blowing up because you have a custom AUTH_USER model that needs to migrate _before_ oauth2_provider), you can replicate the new [migrations folder](https://github.com/dustinfarris/django-oauth-toolkit/tree/django-migrations/oauth2_provider/migrations) from my fork into your app as something like `myapp/oauth2_provider_migrations/` and then add to your settings.py:

```python
MIGRATION_MODULES = {
    'oauth2_provider': 'myapp.oauth2_provider_migrations',
}
```

dustinfarris:
__Note:__ if you need this right now (for instance if Django is blowing up because you have a custom AUTH_USER model that needs to migrate _before_ oauth2_provider), you can replicate the new [migrations folder](https://github.com/dustinfarris/django-oauth-toolkit/tree/django-migrations/oauth2_provider/migrations) from my fork into your app as something like `myapp/oauth2_provider_migrations/` and then add to your settings.py:

```python
MIGRATION_MODULES = {
    'oauth2_provider': 'myapp.oauth2_provider_migrations',
}
```

jcassee:
This change makes django-oauth-toolkit work for my Django 1.7 project. :+1:

Thanks @dustinfarris!

dustinfarris:
Be careful if you are simultaneously upgrading to DOT 0.8 as an extra attribute was added without doing a proper migration. (https://github.com/evonove/django-oauth-toolkit/commit/0656670d1d6cc71b34ab6e55523822791c69833e)

You will begin seeing errors, like:

```
ProgrammingError: column oauth2_provider_application.skip_authorization does not exist
```

To truly migrate, you will have to:

```
./manage.py migrate oauth2_provider zero
./manage.py migrate oauth2_provider
```

_This will result in a loss of data._

I am not sure why the maintainers explicitly chose to forego a migration here.  See #219.  I can potentially update the new migrations for Django >= 1.7 in this PR to accommodate a smoother transition here, but I'd like to know more about why the decision was made to skip this migration in the first place.

synasius:
Thanks for your work!
Please also read #227.

@dustinfarris 
I added south migration for the `skip_authorization` field here 62847af5c322e74075ff454fffcd824a2ebb2b57

alexbw:
Without using south migrations, how can we avoid data loss?

PR #231, opened 2015-04-13
Fixed #230 authentication does not pass in request

### LINKED ISSUE #230: Package authentication does not pass in request
I have a custom ``authentication`` backend which I use alongside django-oauth-toolkit at ``token/``. But,  I need to have access to the request object, however because you don't pass it in here:

https://github.com/evonove/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L344

I cannot access it in my own authentication. Currently I have to do some nasty stuff or add it manually to the package where it calls  ``u = authenticate(username=username, password=password, request=request)``.

Would it be possible to pass in the request object during your own authenticate so that we can have access to it in our own  authentication systems? :) I can create a pull request for this but need to know if this is sensible ok etc. Thanks

glynjackson:
django-oauth-toolkit does not passing client or request to Django's ``authenticate``.  See issue: https://github.com/evonove/django-oauth-toolkit/issues/230

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -325,7 +325,7 @@ class OAuth2Validator(RequestValidator):
         """
         Check username and password correspond to a valid and active User
         """
-        u = authenticate(username=username, password=password)
+        u = authenticate(username=username, password=password, client=client, request=request)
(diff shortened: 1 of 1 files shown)


synasius:
See #230

Issue #127, opened 2014-06-11
Getting { "error": "unsupported_grant_type" }


affo:
Hi,
I am writing an AngularJS application that consumes mi REST Framework APIs.  
I am using DOT as oAuth2 provider.  
The problem is that the same request sent with cURL is successful, once it is sent with an XHR request I get:

    { "error": "unsupported_grant_type" }

I suspected an encoding problem, so I tried to change headers putting:

    'Content-Type': 'application/x-www-form-urlencoded;charset=utf-8'

But it doesn't work...  
For precision's sake, XHR requests are sent with Restangular.

Thank you in advance

palazzem:
Hi (and sorry for taking so long for an answer),
the default behavior is to enforce the use of ``application/x-www-form-urlencoded`` Content-Type.
The ``OAuthLibCore`` implementation extracts the request body using ``request.POST`` dictionary, which is empty for a request that uses ``application/json`` Content-Type.

Anyhow, from your settings, you can set the ``JSONOAuthLibCore`` class that supports ``application/json``. The following example shows how to use the new backend class:

```
OAUTH2_PROVIDER = {
    # other OAUTH2 settings
    'OAUTH2_BACKEND_CLASS': 'oauth2_provider.oauth2_backends.JSONOAuthLibCore'
}
```

This feature will be released in the milestone ``0.9.0`` and is [already merged][1] in master.
Regards.

[1]: https://github.com/evonove/django-oauth-toolkit/pull/234

sacgup:
Hi, Isn't it possible that OAUTH2_BACKEND_CLASS is used based on the content type of the request.
I actually want to support 'Content-Type': 'application/x-www-form-urlencoded;charset=utf-8' as well as 'application/json'. Is there any way i can do it ?

theromis:
+1 I want conditional content type selector as well

ygag-basil:
> JSONOAuthLibCore

Just create a custom oauth2_backends 

```
import json

from oauth2_provider.oauth2_backends import OAuthLibCore

class JSONOAuthLibCoreCustom(OAuthLibCore):
    """
    Extends the default OAuthLibCore to parse correctly application/json requests
    """

    def extract_body(self, request):
        try:
            body = json.loads(request.body.decode("utf-8")).items()
        except (AttributeError, ValueError):
            body = request.POST.items()

        return body
```

and 

```
OAUTH2_PROVIDER = {
    # other OAUTH2 settings
[...]

PR #234, opened 2015-04-19
oauthlib_backend_class is now pluggable through Django settings

palazzem:
Developers can configure the ``oauthlib_backend_class`` used by ``OAuthLibMixin``. DOT generic views retrieve the ``oauthlib_backend_class`` instance through the ``OAUTH2_BACKEND_CLASS`` settings which is defaulted to ``oauth2_provider.oauth2_backends.OAuthLibCore``.

In this way, the RFC compliant ``OAuthLibCore`` class is used as a default, while developers can create or extend new implementations.

The ``oauth2_provider.oauth2_backends.JSONOAuthLibCore`` is provided and using it as a backend class, solves the issue #127. This approach accepts POST request using the ``application/json`` ``Content-Type`` for all token endpoints.

Before merging this PR we should:
* code review provided tests
* choose in which part of the documentation add the ``JSONOAuthLibCore`` example

### DIFF
--- a/oauth2_provider/oauth2_backends.py
+++ b/oauth2_provider/oauth2_backends.py
@@ -30,13 +30,12 @@ class OAuthLibCore(object):
 
         return urlunparse(parsed)
 
-    def _extract_params(self, request):
+    def _extract_headers(self, request):
         """
-        Extract parameters from the Django request object. Such parameters will then be passed to
-        OAuthLib to build its own Request object
+        Extracts headers from the Django request object
+        :param request: The current django.http.HttpRequest object
+        :return: a dictionary with OAuthLib needed headers
         """
-        uri = self._get_escaped_full_path(request)
-        http_method = request.method
         headers = request.META.copy()
         if 'wsgi.input' in headers:
             del headers['wsgi.input']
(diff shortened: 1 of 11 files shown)


oauth2_provider/oauth2_backends.py
```diff
@@ -44,7 +45,26 @@ def _extract_params(self, request):
...
+
+    def _extract_params(self, request):
+        """
+        Extract parameters from the Django request object. Such parameters will then be passed to
+        OAuthLib to build its own Request object. The body should be encoded using OAuthLib urlencoded
+        """
+        uri = self._get_escaped_full_path(request)
+        http_method = request.method
+        headers = self._extract_headers(request)
+        body = urlencode(self._extract_body(request))
```
palazzem: This makes the use of oauthlib ``urlencode`` a requirement

oauth2_provider/oauth2_backends.py
```diff
@@ -136,6 +156,24 @@ def verify_request(self, request, scopes):
...
+    def _extract_body(self, request):
+        """
+        Extracts the JSON body from the Django request object
+        :param request: The current django.http.HttpRequest object
+        :return: provided POST parameters "urlencodable"
+        """
+        try:
+            body = json.loads(request.body.decode('utf-8')).items()
+        except ValueError:
+            body = ""
```
palazzem: The issue #127 is caused because the extracted body was empty for a wrong request.POST decoding in ``application/json`` requests. In this case I maintain the previous behavior while we can raise an exception if the request is not correctly formatted as a JSON. Removing this try-except, leads to an internal server error instead of a bad request.

oauth2_provider/tests/test_oauth2_backends.py
```diff
@@ -1,7 +1,56 @@
...
+    def test_form_urlencoded_extract_params(self):
+        payload = "grant_type=password&username=john&password=123456"
+        request = self.factory.post("/o/token/", payload, content_type="application/x-www-form-urlencoded")
+
+        uri, http_method, body, headers = self.oauthlib_core._extract_params(request)
+        self.assertIn("grant_type=password", body)
+        self.assertIn("username=john", body)
+        self.assertIn("password=123456", body)
+
+    def test_application_json_extract_params(self):
```
palazzem: This test enforces that the default backend class doesn't accept an ``application/json`` Content-Type

oauth2_provider/oauth2_backends.py
synasius: The only thing I would change is remove the _ in the ``_extract_headers`` and ``_extract_body`` methods. The underscore prefix usually mean that the method is "private" and one shouldn't override it.

Maybe we need to define which is the public API of the class and set the name accordingly.

palazzem: +1

As we've discussed via hangout, it's better to promote ``extract_body`` and ``extract_headers`` as a public APIs

PR #239, opened 2015-04-27
wrong exception is sent due to separator being hard coded to "?" when raising an error

### LINKED ISSUE #238: redirect uri on error (e.g. AuthCanceled) may contain two question marks, thus sending the wrong exception
I'm using django-oauth-toolkit along with python-social-auth and of course oauthlib which is a requirement of both, and I've run into an issue when the user clicks the "cancel" button. 

The wrong exception may occur because of a redirect problem.  When the user clicks "cancel", if the redirect_uri contains any URL parameters, an incorrect state error is given instead of `AuthCanceled`, because the URI ends up something like:

`http://example.com?redirect_state=SOME_STATE_TOKEN?error=access_denied`

Note the two question marks here. This is due to an issue in mixins.py - a pull request that fixes this is forthcoming.

honestbleeps:
This checks for the presence of a question mark in the URI and changes the separator accordingly so as to not put two questions marks in the URI.

fixes #238

### DIFF
--- a/oauth2_provider/views/mixins.py
+++ b/oauth2_provider/views/mixins.py
@@ -157,8 +157,11 @@ class OAuthLibMixin(object):
         oauthlib_error = error.oauthlib_error
 
         separator = '?'
-        if '?' in oauthlib_error.redirect_uri:
-            separator = '&'
+        try:
+            if '?' in oauthlib_error.redirect_uri:
+                separator = '&'
+        except:
+            pass
 
         error_response = {
             'error': oauthlib_error,
--- a/oauth2_provider/views/mixins.py
+++ b/oauth2_provider/views/mixins.py
@@ -155,9 +155,14 @@ class OAuthLibMixin(object):
         :param error: :attr:`OAuthToolkitError`
         """
         oauthlib_error = error.oauthlib_error
+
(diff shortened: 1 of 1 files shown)


masci:
Changes look good to me. Do you mind to fix the tests?

honestbleeps:
sorry about that, I will check out the tests today!

honestbleeps:
tests are running now, should be fixed!

oauth2_provider/views/mixins.py
```diff
@@ -155,9 +155,17 @@ def error_response(self, error, **kwargs):
         :param error: :attr:`OAuthToolkitError`
         """
         oauthlib_error = error.oauthlib_error
+
+        separator = '?'
+        try:
+            if '?' in oauthlib_error.redirect_uri:
+                separator = '&'
+        except:
+            pass
```
masci: sorry for being pedantic but I don't like `except: pass` very much ;)
What about something like:
```
redirect_uri = oauthlib_error.redirect_uri or ""
separator = '&' if '?' in redirect_uri else '?'
```
Also, I guess a test case checking the double ? problem (a test that fails without your patch) should be added

PR #246, opened 2015-05-11
Update models fields and Meta

davidfischer-ch:


masci:
Hi and thanks for your code. Where/when does exactly happen and which is the warning? Which versions of django?

Defining the `app_label` field on the `Meta` class solves a different problem, thus the fix seems a little bit hackish... 
The field itself is not needed anymore from Django 1.7 so I expect it'll be deprecated soon.

synasius:
Hi @davidfischer-ch 
thanks for your PR but this cannot be merged:

- we couldn't find the warning you mentioned
- PR is not focused on a specific issue/feature
- all fields should be editable by default, so that they can be edited from the admin. 
- the default registration form should expose only the fields listed in the `Meta.fields`, because it is a good default that guarantees the compliance with RFC. If you want to customize how the Registration form behaves at the moment you have to write your own form and view.

thanks

Issue #196, opened 2015-02-07
Application Creation


acenario:
Is there any reason why any registered user can create an application? I do not want users creating resource-owner based applications. I would like to limit application creation and parameters allowed. Only "developers" should be allowed to create applications and they should be granted specific scopes and only follow grant types that I allow.

What steps do you suggest I take?

maryokhin:
+1 Was very surprised when I found that anyone can use the application creation endpoint.

acenario:
I manually add the urls and create a custom page for application creation. However, there should be a way to control this behavior within oAuth2_provider.  

    from oauth2_provider.views import base
    #     url(r'^oAuth2/', include('oauth2_provider.urls', namespace='oauth2_provider')),
    url(r'^oAuth2/authorize/$', base.AuthorizationView.as_view(), name="authorize"),
    url(r'^oAuth2/token/$', base.TokenView.as_view(), name="token"),

RodneyRichardson:
I too manually added the urls when I discovered this (accidentally). 

I would suggest:
1) Remove the application creation URLs from `oauth2_provider.urls`
2) Add the application creation URLs to, say, `oauth2_provider.app_urls`, so people can use the current web pages if they want.

KodeMaven:
+1 doesn't seem very secure to allow any user to create applications.

KodeMaven:
Wrapping the oauth2 url include with a decorator as per these examples works well and lets you do fine grain access to urls. 

http://stackoverflow.com/questions/2307926/is-it-possible-to-decorate-include-in-django-urls-with-login-required

morganbohn:
+1 It seems to be no secure.

nryoung:
:+1:, fine tuned permission controls would be a great addition.

davidsidlinger:
This really needs to change. We will likely use the workaround of URL registration, but that seems unacceptably fragile in the long run. I'd even take a simple setting to disable the Applications UI.

valerymelou:
I noticed this the first time I browsed the doc. I'm starting a new project using it and no way to continue without fixing it. But I think I will implement what @scenario said.

poswald:
I agree with @RodneyRichardson's suggestion on this. A quick fix would be to simply remove the suggestion in the documentation to users that they should add all urls to their app via `oauth2_provider.urls`.

poswald:
Can someone please check the documentation I added in #379 here? It's based roughly on @dotmobo's solution.

[...]

PR #248, opened 2015-05-18
Issue #196 Add permission checking to application registration

trbs:
Add an addition setting called APPLICATION_REGISTRATION_PERMISSIONS
which controls which permissions must be set on a user to allow the
creation of applications.

By default (in this PR) this is the 'add_application' permission which
regretfully is not backwards compatible but it does seem to make the
most sense from a security perspective as a default.

Reverting to the old behaviour is easily done by using None for the
settings key.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -40,12 +40,29 @@ this value if you wrote your own implementation (subclass of
 
 APPLICATION_REGISTRATION_PERMISSIONS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Which permissions are required to be able to register an application. By default
-only superusers or users with ```oauth2_provider | application | Can add application```
-permission are allowed to register applications. If this setting is set to ```None```
-then all authenticated users are allowed to register applications. For more information
-about setting more complex permissions see
+Defines which permissions are required to be able to register an application.
+By default every authenticated user is able to register applications.
+
+A common setup for none public sites might be to only allow superusers and users
+with the ```oauth2_provider | application | Can add application``` permission to be allowed
+to register applications.
+
+This can be done with the following configuration::
+
+    OAUTH2_PROVIDER = {
+        ...
+        'APPLICATION_REGISTRATION_PERMISSIONS': {
+            'all': ('oauth2_provider.add_application', ),
+        },
+        ...
+    }
+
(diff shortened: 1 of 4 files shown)


synasius:
HI @trbs,
great work! 

I'd prefer to leave the default behavior unchanged, so the app is backwards compatible.

Can you add some docs for the settings APPLICATION_REGISTRATION_PERMISSIONS with an example and some specifc tests?

Thanks a lot for your effort

trbs:
@synasius the default is now backwards compatible and I put the old settings as the example in the documentation. Two extra tests now test creating an application with and without the correct permission.

docs/settings.rst
```diff
@@ -38,6 +38,32 @@ The import string of the class (model) representing your applications. Overwrite
 this value if you wrote your own implementation (subclass of
 ``oauth2_provider.models.Application``).
 
+APPLICATION_REGISTRATION_PERMISSIONS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Defines which permissions are required to be able to register an application.
+By default every authenticated user is able to register applications.
+
+A common setup for none public sites might be to only allow superusers and users
```
masci: s/none/non/

oauth2_provider/settings.py
```diff
@@ -41,6 +41,10 @@
     'AUTHORIZATION_CODE_EXPIRE_SECONDS': 60,
     'ACCESS_TOKEN_EXPIRE_SECONDS': 36000,
     'APPLICATION_MODEL': getattr(settings, 'OAUTH2_PROVIDER_APPLICATION_MODEL', 'oauth2_provider.Application'),
+    'APPLICATION_REGISTRATION_PERMISSIONS': None,
```
masci: key is duplicated

pySilver:
Isn't it better to have a configurable callback that tests if user has rights to create apps? Originally it won't be defined and would do nothing but in real life experience one can supply own callback which will receive lets say `User` instance and will be expected to return boolean. So, implementation would be up to the developer.

synasius:
@pySilver
 
> Isn't it better to have a configurable callback that tests if user has rights to create apps?

Definitely! That way we're not bound to how django handles permissions. I'm closing this PR.
@trbs thanks for the great work but it's not what we're looking for.

Issue #148, opened 2014-09-02
Delete expired tokens from database


shvechikov:
I don't see any purging mechanism for expired tokens.
I think it's worth to have a management command to delete expired tokens.
Am I missing something here?

alewisohn:
I came up with the following logic for cleaning up tokens. In our app, it runs as a periodic task, but it could easily be converted to a management command.

```python
AccessToken.objects.filter(refresh_token__isnull=True, expires__lt=datetime.now()).delete()
```

Basically, we can delete any access tokens that have expired and that do not have a corresponding refresh token, i.e., the refresh token for that access token has been used. Unused refresh tokens and their related access tokens will still hang around. 

Without and expiration on refresh tokens, there's no way to clean them up without invalidating the related access tokens. A management command could be written that accepts an explicit expiration time for refresh tokens though.

trbs:
Why not (at least) configure a static REFRESH_TOKEN_EXPIRE_SECONDS which indicated how long we want to keep the access_token + refresh_token pair after the access_token has expired.

The cleanup mechanism could then be:
```
with transaction.atomic():
    if REFRESH_TOKEN_EXPIRE_SECONDS:
        refresh_expire_date = timezone.now()-timedelta(seconds=REFRESH_TOKEN_EXPIRE_SECONDS)
        RefreshToken.objects.filter(access_token__expires__lt=refresh_expire_date).delete()
    AccessToken.objects.filter(refresh_token__isnull=True, expires__lt=datetime.now()).delete()
```

Futher steps might be to be able to override the REFRESH_TOKEN_EXPIRE_SECONDS per application. So that one might have unlimited refresh_tokens on some applications (ios/android apps?) and not on others (web api's).

SilviyaRepo:
Could you tell me elaborately ow you achieved this???

bforchhammer:
We use the oauth toolkit with a quite a large user-base and are getting to a point where the access tokens table is getting quite large; we have >100M tokens with only a small percentage active. I would love to see a way to purge expired ones, without having to delete any refresh tokens.

The cleanup script above (and in the related issue/pull request) looks good, but it only deletes access tokens for which the refresh token has been used already. Could we instead make the relationship between access and refresh token optional and delete *all* expired access tokens?

synasius:
Closed by #249

silviyaranirepo:
[...]

PR #249, opened 2015-05-18
Issue #148 cleanup of expired tokens

trbs:
Simple implementation of cleanup strategy for expired tokens.

This cleanups expired ```access_token```'s (when there is no ```refresh_token```) and expired grant entries. When ```refresh_token_expire_seconds``` is set it will expire ```refresh_token```'s as well.

Happy to write documentation if we decide to move forward with this or similar PR.
But I first wanted to submit this PR so we can get the discussion on how to expire token and cleanup the database periodically going.


masci:
:+1: very nice!

trbs:
To clarify cause I think it's only clear after looking at the code :)

In this PR the idea is that a refresh_token can expire ```REFRESH_TOKEN_EXPIRE_SECONDS``` _after_ the access_token has expired. This way no changes to model are necessary and it seems like a logical thing to do :)

I would like to see if we can/should tie this in with the custom models for tokens and applications 
PR. So expire cleanup is flexible enough for people to easily write their own custom expire strategies.

For example; Being able to use a custom refresh_token model that has an ```expires``` field (configurable via REFRESH_TOKEN_EXPIRE_FIELDNAME ?) or an Application model that defines a ```refresh_token_expires_after``` field or similar.

I would assume that different applications with different expire strategies could be fairly common. Like having a part of the api for IOS/Android apps that rarely (or never) expire and a public user/development api which should in expire in minutes or hours.

pySilver:
I like this PR. Refresh tokens should not be immortal and should be removed at some point, they just lasts more longer than access tokens.

synasius:
I want to merge this PR :)
Can you please add some docs for the django command and the `REFRESH_TOKEN_EXPIRE_SECONDS` setting??

Also add yourself to contributors.

trbs:
Documentation is only for ```REFRESH_TOKEN_EXPIRE_SECONDS``` at the moment. Information on the management command still needs to be written. 

What would be the most reasonable default for ```REFRESH_TOKEN_EXPIRE_SECONDS``` ? Leave it set to None which is backwards compatible but will fill up the database at some point... or set it too some multiple of ```ACCESS_TOKEN_EXPIRE_SECONDS``` ?

I will try to look at that and the conflicts as soon as I can.

synasius:
Merged as cd462990e132de1e50328b74a6f324ef77543805. 
We need to add documentation for the `cleartokens` command too. I opened #324 to keep track  of it.

PR #252, opened 2015-05-22
New feature: Configurable AccessToken and RefreshToken

dirkmoors:
Because I want to be able to implement custom AccessToken and RefreshToken models, I've implemented a customizable AccessToken and RefreshToken, in the same way that the Application model was already configurable via the settings


masci:
:+1: Looks good to me

pySilver:
any news on this?

dfeinzeig:
Seems this could be used to implement JWT as the access tokens?

clintonb:
This would greatly help those of use who want to use JWT access tokens. See #368 and #343.
@dirkmoors @masci @palazzem @synasius what can we do to get this released?

menecio:
Already implemented on https://github.com/evonove/django-oauth-toolkit/pull/467 and landed on master

@jleclanche I think this one can be closed also.

duaneking:
Just to double check: Was anything a part of this what was not covered by #467 that would be good for #397 or JWT support in general?

PR #261, opened 2015-06-23
Fix the migrations to be two-step and allow upgrade from 0.7.2

OdyX:
This splits the Django (non-South) migrations to ease the upgrade from 0.7.2 to 0.8.1+.

With this in place, one can fake the 0001 migration and then migrate to 0002 correctly, resulting in a correctly migrated app.


bforchhammer:
This looks good to me! I have not updated the package yet because of the migration process; this would definitely help! :+1:

synasius:
Thanks @OdyX for your fix!
I just released a 0.8.2 on PyPi with this fix included

Issue #277, opened 2015-07-23
The oauthlib-1.0.0 breaks django-oauth-toolkit-0.8.2


abarto:
oauthlib-1.0.0 was released recently and it seems to break the current version of django-oauth-tookit. The site works just fine with oauthlib-0.7.2. Here's a traceback showing the issue:

Internal Server Error: /o/token/
Traceback (most recent call last):
  File "venv/lib64/python3.4/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "venvlib64/python3.4/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "venv/lib64/python3.4/site-packages/braces/views/_forms.py", line 22, in dispatch
    return super(CsrfExemptMixin, self).dispatch(*args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "venv/lib64/python3.4/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "venv/lib64/python3.4/site-packages/oauth2_provider/views/base.py", line 170, in post
    url, headers, body, status = self.create_token_response(request)
  File "venv/lib64/python3.4/site-packages/oauth2_provider/views/mixins.py", line 124, in create_token_response
    return core.create_token_response(request)
  File "venv/lib64/python3.4/site-packages/oauth2_provider/oauth2_backends.py", line 106, in create_token_response
    headers)
[...]

PR #278, opened 2015-07-23
Proposal for a fix of issue #277

abarto:
Use gettattr instead of accessing the request's fields directly in client_authentication_required.


masci:
fixed with #245, thanks for your time anyway!

PR #285, opened 2015-08-05
fixed South migrations for custom User model

jensadne:
The current South migrations fails in cases where the User model is inherited from another model which in turn inherits from AbstractBaseUser, because they except there to be a column named "id" in the User model's table. This patch fixes that by changing from "id" to User._meta.pk.column, which gives the correct value.


jleclanche:
South migrations are gone, this PR should be closed.

PR #297, opened 2015-09-23
Fixing doc about SessionAuthenticationMiddleware

lzen:
As discussed in https://github.com/evonove/django-oauth-toolkit/issues/295.


lzen:
@masci Thanks!

PR #308, opened 2015-10-16
Use compact syntax for tox envs

dulaccc:
I came across this syntax minutes ago, it can be easier to read and avoid annoying typo mistakes.

Cheers


synasius:
:+1:

PR #309, opened 2015-10-16
Add the py35-django19 env to travis but allow it to fail

dulaccc:
Be ready for the integration of Python 3.5.
This way, we keep an eye when looking for ci builds, and we'll only need to remove the `allow_failures` key in `travis.yml`.


synasius:
thnx @dulaccc

dulaccc:
always a pleasure !

synasius:
@dulaccc next time add yourself to contributors list.

Issue #296, opened 2015-09-23
JSONOAuthLibCore breaks multipart form-data request for django-rest-framework


sacgup:
I used JSONOAuthLibCore to accept JSON data for getting tokens.
But after using it gives error for multipart form-data request calls to my api's which are made using django-rest-framework.  I need to use multipart-form data to upload files only and json for all other requests.

I used below class for authentication in settings.py

    'DEFAULT_AUTHENTICATION_CLASSES': (
        'oauth2_provider.ext.rest_framework.OAuth2Authentication',
    ),

Error given is "You cannot access body after reading from request's data stream".

On investigation i found django doesn't support to read from request body twice. Once it is read in JSONOAuthLibCore so when django reads data again in "django/http/request.py" it gives error at that time.

I used JSONOAuthLibCore so that my all api call to my view and oauth2 to get token are similar and accepts only json.

1.) Why oauth2 need to read body data only to authenticate request ? It should only read Authorization header and check the token to validate the user.  It should read body data only when granting tokens.

2.) Any workaround for this problem ? 

I can make my oauth2 call to send "x-www-form-urlencoded" data from client apps and json for rest of my views but i really don't want to do that.

PS: I am new to django-rest-framework, oauth2  and token authentication. So if i am missing something obvious please bear with me.

masci:
Hi @sacgup could you provide a minimum working example I can work with? I haven't found a simple workaround and need to investigate further.

sacgup:
Hi @masci 

Just override the default oauth2 backend class from OauthLibCore to JSONOauthLibCore in settings.py.

    OAUTH2_PROVIDER = {
        # this is the list of available scopes
        'SCOPES': {'read': 'Read scope', 'write': 'Write scope'},
        'OAUTH2_VALIDATOR_CLASS': 'external.validators.EmailOAuth2Validator', # Self over ride : To use email as username to get access tokens
        # 'OAUTH2_BACKEND_CLASS': 'oauth2_provider.oauth2_backends.JSONOAuthLibCore'
    }

and use the authentication class from oauth2_provider in django rest framework.
Make a request to any api of rest framework in multipart form-data or x-www-form-urlencoded content type.

You will be able to see the error.

What i have found out till now: 
JSONOauthLibCore reads the request body to get the json data and authenticate the request.
Then, django also reads the body later in it's http request class and gives error at that point of time because django allows to read request body only once. ( Don't know Why ? )

[...]

PR #311, opened 2015-10-20
Improve django-rest-framework support

dulaccc:
Hi guys,

I stubbled across some difficulties with DRF, and since you have an integration guide with it I think it could be helpful to bring some updates.

With multipart form requests I had an odd issue, because the request body was read by DOT before letting the form parser read it. It has been reported in #296, but no pull request associated, so here it is.

And I've added a little addition to the `OAuth2Authentication` class to reject non active users (thanks to @pySilver for the heads-up).

Cheers :beers:

### DIFF
--- a/oauth2_provider/ext/rest_framework/oauth2_backends.py
+++ b/oauth2_provider/ext/rest_framework/oauth2_backends.py
@@ -15,5 +15,12 @@ class OAuthLibCore(oauth2_backends.OAuthLibCore):
             for authentication and not pass the credentials in the request body
         """
         if isinstance(request, Request):
-            return request.data.items()
+            try:
+                # support DRF 2.x
+                if not hasattr(request, 'data'):
+                    return request.DATA.items()
+                return request.data.items()
+            except (ValueError, AttributeError):
+                # complex json request (list?) is not easily serializable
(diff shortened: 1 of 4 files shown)


oauth2_provider/ext/rest_framework/oauth2_backends.py
```diff
@@ -0,0 +1,19 @@
...
+        """
+        We can read only once the body in Django, 
+        so in case of DRF we avoid to read it before the framework does.
+        This use case often happen during multipart form requests.
+
+        NB: it forces you to use the `Authorization` request header 
+            for authentication and not pass the credentials in the request body
+        """
+        if isinstance(request, Request):
+            return request.data.items()
```
pySilver: This would fail once you will start using any permission classes like TokenReadWriteScope and so on. Problem is that this perm classes will access this method and will fail if request.data is not a dictionary :) imagine request with json list payload.

p.s. DRF 2.x should be supported, where .data is not available.

jleclanche:
I'm going to have to close this PR, the library changed quite a lot and so did DRF support. If you want to PR any improvements that you deem are still necessary, please do!

Issue #274, opened 2015-07-15
BUG: Application registration form not working for extended application models


mark-tan:
Submitting the registration form redirects to a 404 page when using a custom application model.

Upon initial inspection of the code, the problem seems to have been introduced in commit f5a1a493b939447076cb7af7553522b398b1d1eb while fixing #151. It looks like the form is using the default application model.

jannon:
I just ran into this, too after upgrading to 0.9.0.  I'm not sure what issue really caused #151, but I'm running Django 1.7.10 and when I change the `RegistrationForm` code to 
```python
...
from .models import get_application_model

Application = get_application_model()

...

class RegistrationForm(forms.ModelForm):
    class Meta:
        model = Application
        fields = ('name', 'client_id', 'client_secret', 'client_type', 'authorization_grant_type', 'redirect_uris')
```

everything works just fine.  So +1 to changing `RegistrationForm` back to properly using custom application models

detectedstealth:
Any status on this I have just ran into the same situation tonight. Will the pull request that @jannon made make it into master soonish or do I have to hack my local version? On a deadline need to use a custom model. 

Using Django 1.8.4

PR #312, opened 2015-10-22
RegistrationForm should use custom Application model

jannon:
This is a fix for #274.  All tests pass so unless the author of https://github.com/evonove/django-oauth-toolkit/commit/f5a1a493b939447076cb7af7553522b398b1d1eb can provide a failing test, I think the custom application model support should be restored.  This is really a bug that completely breaks custom application model use in registration

### DIFF
--- a/oauth2_provider/forms.py
+++ b/oauth2_provider/forms.py
@@ -1,6 +1,8 @@
 from django import forms
 
-from .models import Application
+from .models import get_application_model
+
+Application = get_application_model()
 
 
 class AllowForm(forms.Form):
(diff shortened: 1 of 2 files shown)


outsbart:
@jannon could you please provide a failing test reproducing #274?

jannon:
Ah, knew I forgot something. Sure thing.

jannon:
actually, we need only use the `TestApplication` model in the settings and the existing application registration test will catch the failure

synasius:
This can't be merged because the `get_application_model` at module level raises the exception described in #151. 
BTW, I pushed a proper solution which I'll merge asap. See #323

PR #321, opened 2015-11-16
Added a created and modified date to RefreshToken

bobkarreman:
To know when the connection with the application was made


jleclanche:
+1 on this, although the migrations will have to be recreated.

PR #330, opened 2015-12-08
added missing swappable migration option

bartdag:
This migration option prevents the creation of the Application model if an alternative model is already provided.

The swappable option is already set in models.py, it was only missing in the initial migration.


synasius:
merged with a8187642472cf4411e5157352abbf84a761b66e4

PR #341, opened 2016-01-05
Fix #340 by defaulting encoding var when request.encoding is None

### LINKED ISSUE #340: Error when requesting an oauth access token with the same request encoding as the django default
Details in this SO question: http://stackoverflow.com/questions/34403912/when-trying-to-request-for-oauth-access-token-i-get-typeerror-encode-argumen

Error is `TypeError: encode() argument 1 must be string, not None`

dougkeen:
`None` is a valid value for `request.encoding`, as documented: https://docs.djangoproject.com/en/1.9/ref/request-response/#django.http.HttpRequest.encoding)

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -7,6 +7,7 @@ import logging
 from datetime import timedelta
 
 from django.utils import timezone
+from django.conf import settings
 from django.contrib.auth import authenticate
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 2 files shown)


powdahound:
:+1:

This will resolve the following issue: http://stackoverflow.com/questions/34403912/when-trying-to-request-for-oauth-access-token-i-get-typeerror-encode-argumen

I'm curious how people have been using django-oauth-toolkit successfully with this bug in place. It seems like it'd get in everyone's way, but I must be missing something.

PR #344, opened 2016-02-06
add limit support of application module

eastpiger:
add limit support of application module
add check scope from settings and allowed scopes of client
# Change:
- Add scope and default_scope to application module
- update the form of application module as well as its validator function
- change validate_scopes function to check if the request scope is allowed
- change get_default_scopes function to provide default scope from application
- remove setting DEFAULT_SCOPES and add DEFAULT_APPLICATION_ALLOWED_SCOPES & DEFAULT_APPLICATION_DEFAULT_SCOPES
# Reference:
## Oauthlib > suggest client module:

Scopes:

Required. The list of scopes the client may request access to. If you allow multiple types of grants this will vary related to their different security properties. For example, the Implicit Grant might only allow read-only scopes but the Authorization Grant also allow writes.

```
# You could represent it either as a list of keys or by serializing
# the scopes into a string.
scopes = django.db.models.TextField()

# You might also want to mark a certain set of scopes as default
# scopes in case the client does not specify any in the authorization
default_scopes = django.db.models.TextField()
Redirect URIs:
```


eastpiger:
Also fixed this [issue#345](https://github.com/evonove/django-oauth-toolkit/issues/345)

According to Release Notes of pip, Python3.2 is not supported from pip 8.0.0 which is released at 8.0.0 (2016-01-19)

so 
    - env: TOX_ENV=py32-django17
    - env: TOX_ENV=py32-django18
is moved to allow_failures

jleclanche:
Use the scopes backend from 0.12.0.

PR #347, opened 2016-02-11
Made all models swappable

thedrow:
This is useful for having an access model token that encrypts the tokens in the database using pgcrypto fields or for replacing the user fields with denormalised user fields using [django-denorm](https://github.com/django-denorm/django-denorm).
There are a lot of other cases that require all models to be swappable. We can discuss them if you want.
Unfortunately I uncovered a regression in related fields of swappable models at least in Django 1.9 on Python 2.7.
We need to figure out why and fix it if possible.

### DIFF
--- a/oauth2_provider/admin.py
+++ b/oauth2_provider/admin.py
@@ -1,12 +1,16 @@
 from django.contrib import admin
 
-from .models import Grant, AccessToken, RefreshToken, get_application_model
+from .models import (get_grant_model, get_access_token_model,
+                     get_refersh_token_model, get_application_model)
 
 
 class RawIDAdmin(admin.ModelAdmin):
     raw_id_fields = ('user',)
 
 Application = get_application_model()
+Grant = get_grant_model()
+AccessToken = get_access_token_model()
+RefreshToken = get_refersh_token_model()
 
 admin.site.register(Application, RawIDAdmin)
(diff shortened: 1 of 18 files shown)


thedrow:
The regression is present in Django 1.9 on all Python versions.
See https://travis-ci.org/evonove/django-oauth-toolkit/jobs/108660974#L171

thedrow:
Any idea why the tests fail on other Django versions as well?

thedrow:
And now you can see the failure in 1.7 as well. https://travis-ci.org/evonove/django-oauth-toolkit/jobs/108665464#L171

oauth2_provider/admin.py
```diff
@@ -1,12 +1,16 @@
 from django.contrib import admin
 
-from .models import Grant, AccessToken, RefreshToken, get_application_model
+from .models import (get_grant_model, get_access_token_model,
+                     get_refersh_token_model, get_application_model)
```
agriffis: `s/get_refersh_token_model/get_refresh_token_model/g`

menecio:
Any chance to help with this one?

thedrow:
You can take it from here. I'm no longer interested in working on this.

menecio:
Implemented on https://github.com/evonove/django-oauth-toolkit/pull/467 and landed on master. 

@jleclanche I think this one can be closed now.

PR #352, opened 2016-03-13
Use TemplateView as base for BaseAuthorizationView

stefanw:
The view class calls on `render_to_response` which is only defined in Django's `TemplateView`, not in `View`.


agriffis:
LGTM. `AuthorizationView` which inherits from `BaseAuthorizationView` depends on `FormView` so it works as is, but that wouldn't automatically be true for anybody else inheriting from `BaseAuthorizationView`

jleclanche:
I don't fully understand why there is a separate `BaseAuthorizationView` class, to be honest. The `error_response` method won't work outside of it, so all it does is add an `oauth2_data` dict and inherit login required.

If it's meant to make it easier to override the actual base functionality of the AuthorizationView we should move a lot more stuff to it.

Issue #357, opened 2016-03-16
Support multiple-user clients


poswald:
The DOT code makes the assumption that Application <-- 1:1 --> User and forces it to be non-null.

Reading the OAuth2 specification, this is not the case in some of the grant types. In `client_credentials` grant type, there is not necessarily a user (in the oauth sense) associated with the client. Since Django always expects an `auth.User` of _some_ sort, creating a dedicated `auth.User` to represent the client is a reasonable workaround. In [issue #38][issue-38] I had offered a workaround where an auth.User that represents the client could be used if one wished so Django/DRF is happy.

Now, it is also common to use the `password` grant type with mobile clients. In this mode, the `client_id` is unique and there is no `client_secret` (because it is assumed that the client can't keep it secret.) There is not a very good way for the server to authenticate the client. In this case, the user has to verify that they are using an official client some other way (out of band). In this flow the user's username/password is sent over and therefore it is the case that the same Application/client_id would be used by more than one `auth.User`. Generally, DOT seems to do the right thing and authenticate the client when it can and the user separately and returns that user.

At the end of #38 it seemed like there [was some agreement][agreement] for relaxing the constraint and making user optional on `AbstractApplication`. If that still stands, I'd like to make that PR and submit it for consideration.

[issue-38]: https://github.com/evonove/django-oauth-toolkit/issues/38
[agreement]: https://github.com/evonove/django-oauth-toolkit/issues/38#issuecomment-165028954

synasius:
Fixed by #358

PR #358, opened 2016-03-17
Nullable Users

poswald:
This is an attempt at solving Issue #357 

I would still like to review if there are any code paths that assume `User` is always filled in.

Also, I am not sure if migrations need special treatment if they are based on a swappable model... any advice here by anyone is appreciated.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -16,3 +16,4 @@ Hiroki Kiyohara
 Diego Garcia
 Bas van Oostveen
 Bart Merenda
+Paul Oswald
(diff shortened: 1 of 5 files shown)


synasius:
PR looks good to me. I checked the code to see whether the Application.user field was not used in any of the oauth2 flows and I can confirm it's not. 

Before we merge please add a line to the changelog, both in in the `README.md` and in the `docs/changelog.rst`. I think you need to add the section for `development` changes to both the files.

Also add yourself to the Contributors list in the AUTHORS file.

synasius:
@poswald thanks! lot of people will love this patch!

Issue #386, opened 2016-05-26
Making templates easier to customize


jacktrades:
I just had the need to add some branding and navigation to the application registration pages.  I didn't notice a base template that was easy to override, so I ended up copy/pasting the base template into my project (with a few minor tweaks) and setting up an inheritance scheme similar to the Django admin.

    _base.html  >  base.html  >  application_*.html

This simple change gives users a hook to modify the base templates simply by adding a `oauth_provider/base.html` file to their templates directory, and overriding blocks there.

    {% extends 'oauth2_provider/_base.html' %}

    {% block content_header %}
        <h1>Hello World</h1>
    {% endblock content_header %}

I have a [branch here](https://github.com/jacktrades/django-oauth-toolkit/tree/overridable-base-template) that implements this, and will submit a pull request after submitting this issue.

Adding/reorganizing a few blocks in the base template would make the application templates even easier to customize.  If there's any interest, I'm happy to submit a pull request for those as well.

PR #387, opened 2016-05-26
Allows users to override oauth2_provider/base.html

jacktrades:
This uses a template inheritance scheme similar to the Django admin so
users can easily override parts of the base template, without having to
copy/paste the base template and manually sync changes with upstream.

For example, users can add a header by pasting the following into
`template_directory/oauth2_provider/base.html`:

```
{% extends 'oauth2_provider/_base.html' %}

{% block content_header %}
    <h1>Hello World</h1>
{% endblock content_header %}
```

Addresses issue #386.


jleclanche:
Closing, this is just too old to merge.

Issue #318, opened 2015-11-03
Possible security issue


gabn88:
I have an existing API (DRF) that uses, 1. JWT, 2. 'normal' tokens, 3. Sessions to gather the date from the API. I have permissions and filters in place, so that users can see,write, change, delete only the stuff that they should be allowed to do.

Now I have created an app and used Oauth to log the user in. I did not pose any allowances on my API, but still the user is permitted all the actions that he could with a JWT, 'normal' token or Session!! This means that any other app the subscribes himself via de o/applications can BY DEFAULT do all this stuff (of course when the user allows him). 

This sounds like a crazy security issue, since any existing user can create an app and show the login form. I had more imagined it to be like, by default no permissions, only when you allow a certain app permissions the app has permission.

~~Also I have done the tutorial using DRF (using cUrl). The application that the user granted access also had access to ALL api endpoints and ALL actions are permitted by default. I really find this strange security policy and do not think this is a sensible default. Also it seems that the application can defined whatever scopes he likes to grant the user permission, but it can also be that I don't understand this part well enough.~~

Also I would like to know if it is possible to change this setting system wide? So that an application has access to only API endpoints with explicitly defined scopes?

poswald:
Can you please give some more description of your setup and what you did (which grant type you're using, etc..)? Are you saying you expected a user who is authenticated to have different permissions when authenticated through the API, or if you're saying that you expected the application endpoints not to be available to them?

cleder:
Also you should report security issues via an email directly to the author, not in the issue tracker!

gabn88:
## What I wanted it to be
A Facebook like Oauth system, where an application has to be given permissions by some kind of system superuser. A users that use the application for example get a message "Do you want to grant application XYZ to have permission to:  a. write on your wall, b. see your personal information, c. create new subaccounts, etc..." and the application only gets access/permissions to do that via the endpoints that are especially tagged with a certain scope.

## How it worked for me when I installed it
For me it worked like an Oauth system, where people themselves can create applications (without su control, although this should be not to difficult to implement, but I think it would be nice to have in the core). And when an application is given access by a certain user, this application has all the rights the user has. 

It could be that I was doing something wrong, but I found this to be very suspicious behaviour with regards to security. So much that I ended up not implementing it (yet).

palazzem:
Hello everyone!
[...]

Issue #130, opened 2014-06-18
apps with different scopes


gregbanks:
hi,

i'd like to implement multiple apps with different scopes that will be hitting the same django-oauth-toolkit provider (ex. app1 should just be associated with the "read" scope while app2 should have both the "read" and "write" scope). it appears that all apps are authorized for whatever scopes are in `oauth2_provider.settings.oauth2_settings._SCOPES` via a call into `oauth2_provder.oauth2_validators.OAuth2Validator.get_default_scopes` from oauthlib during the authorization flow.

am i missing something? if i am not, what would be the suggested way to implement the aforementioned functionality? should i extend `oauth2_provider.models.Application` to have a scopes attribute and then override `oauth2_provder.oauth2_validators.OAuth2Validator.get_default_scopes` to query that attribute?

thanks!

cheers,
greg

jleclanche:
The new model-based scopes backends implement this!

SinaZK:
Is this implemented? Can't find it in docs.

PR #395, opened 2016-06-29
App specific scopes

### LINKED ISSUE #216: feature request: application specific scope restrictions
I've been working on this myself, as it's a need for a project I'm working on, but here is a formal issue for discussion of the ideal implementation (mine is a bit hacked together), etc.

**What is the request**:

django-oauth-toolkit offers the ability to create custom scopes, but does not have a way to restrict each application to a separate subset of scopes.

the need may arise (as it has in my case) to not give all applications access to all scopes - some scopes may be restricted to more trusted applications/individuals.  For example, you may want to grant all applications a "read_basic_details" scope, but only trusted applications should be able to use the "read_full_details" scope.

**Suggested course of action**:

My implementation adds an allowed_scopes field to the application model.  Just like in the OAuth spec, the scopes are listed as space-separated strings.  When an access token is requested, the user is only given the option to request the scopes present in allowed_scopes -- this allows any application to be restricted only to specific scopes.

My pull request, to my own fork, is here:

https://github.com/honestbleeps/django-oauth-toolkit/pull/1

**TODO**:
- My code is probably suboptimal, and should be reviewed and perhaps redone.
- It stands to reason that if the site owner wants to restrict scopes, then application owners should not be able to edit the allowed_scopes field. I've not implemented this, as I'm just using waffle tags to only allow site administrators access to create or view the listing of applications for now. Ideally, to keep django-oauth-toolkit as an all in one / turnkey solution, this problem might be best addressed within the package itself.

Thanks for your consideration and review!

JensTimmerman:
rebased code from honestbleeps on latest master.
fixes #216 #130 
possibly also helps to resolve issues like #318


oauth2_provider/migrations/0004_application_allowed_scopes.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10b1 on 2016-06-29 15:18
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
cleder: should there also be a data migration that add the default scopes to existing Applications?

cleder: Should there also be a data migration that sets the scopes for existing applications?

JensTimmerman: I don't believe so, if no custom scope is set, the default scopes will still be used.

JensTimmerman:
I'm not sure if it makes sense to try to include this in the AuthorizationView, it usually doesn't work in the initial_data, because when you don't know the client_id, you can't do anything yet, 
and it doesn't seem in it's place in the validation. Can I get some feedback on this?

honestbleeps:
Thanks so much for doing the work to rebase my PR!  I've been extremely busy and ended up getting tied up too much to do the work that I felt might be necessary before vetting it as a PR to the original master branch, which is why it was just a PR against my own.

It'd be great to see this make it in!

honestbleeps:
I should add, just in case it's not obvious from the code:

There's no mechanism built into this to handle granting an application access to the requested scopes.  In my actual use of this code, I utilized the django admin for manually granting scopes to apps. I'd trigger an email for any app that requested scopes we considered "elevated", and I'd manually copy them over in the django admin.

A cleaner and more universally appropriate UX for app creators to request specific scopes (and see which are considered "elevated") and for site admins to grant those scopes would be optimal. However, this PR is still sufficient for people to use as long as they understand that they need to roll their own solution for granting scopes (and tracking requests of said scopes!)

mkjpryor-stfc:
@JensTimmerman: I am also keen on this functionality.

It seems that the problem with the Travis build is the use of `application` rather than `self`...

docs/advanced_topics.rst
```diff
@@ -72,3 +72,13 @@ You might want to completely bypass the authorization form, for instance if your
 in-house product or if you already trust the application owner by other means. To this end, you have to
 set ``skip_authorization = True`` on the ``Application`` model, either programmaticaly or within the
 Django admin. Users will *not* be prompted for authorization, even on the first use of the application.
+
+
+App Specific Scopes
+===================
+You can restrict each individual application to a separate subset of scopes.
+Using the ``allowed_scopes = 'scope1 scope2'`` field on the ``Application`` model you can limit the scopes an application
+can get. This is usefull if you trust an application owner only for e.g. read access and don't want
```
clintonb: s/usefull/useful

JensTimmerman:
@mkjpryor-stfc I'll look into fixing some more issues I encountered while working on this. But I'm awaiting some feedback from the original developers on this.

mkjpryor-stfc:
@JensTimmerman: No worries. Let me know if there's anything I can do to help this along.

oauth2_provider/views/base.py
```diff
@@ -87,6 +88,14 @@ def get_initial(self):
         }
         return initial_data
 
+    def get_application(self, client_id):
+        """
+        get an application from a client_id, this caches the applications
+        """
+        app = self._application_cache.get(client_id, get_application_model().objects.get(client_id=client_id))
```
JensTimmerman: this doesn't actually cache it, it is still recreated every time

honestbleeps:
@JensTimmerman are you waiting on me for feedback? I'm the original author but I don't recall seeing any open questions... perhaps I missed it?

or do you mean the original authors of django-oauth-toolkit?

JensTimmerman:
I mean the original authors of django-oauth-toolkit.

oauth2_provider/models.py
```diff
@@ -123,6 +129,21 @@ def clean(self):
     def get_absolute_url(self):
         return reverse('oauth2_provider:detail', args=[str(self.id)])
 
+    def get_allowed_scopes_from_scopes(self, scopes):
+        """
+        Filters the given list of scopes so it only contains the allowed scopes
+        """
+        scopes = scopes.split(' ')
+        allowed_scopes = None
+        if application.allowed_scopes:
```
jeremyll: `s/application.allowed_scopes/self.allowed_scopes/`. Excited for this PR to get merged.

clintonb:
@JensTimmerman @synasius @masci what is needed to get this merged?

honestbleeps:
I would also really love to see this merged (selfishly, of course, it was my suggestion and original contribution, though @JensTimmerman is responsible for doing a ton of work to clean up my initial pass at it)...

If I can do anything to assist, let me know.

JensTimmerman:
Hi all, 
Whilst working on this PR I found a security issue in the django-oauth-toolkit, I contacted security@evonove.it to report it, but have still not heard back from them. I was awaiting their response before going public with this.

honestbleeps:
if the security issue has to do with open redirect, that'd lie in the oauthlib dependency, and their argument for not fixing it is that it actually follows OAuth 2.0 spec... which it does.

if it's not, please carry on!

oauth2_provider/models.py
```diff
@@ -6,7 +6,7 @@
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.db import models, transaction
-from django.utils import timezone
+from django.utils import timezone, six, six
```
clintonb: You're importing `six` twice.

oauth2_provider/models.py
```diff
@@ -123,6 +129,22 @@ def clean(self):
     def get_absolute_url(self):
         return reverse('oauth2_provider:detail', args=[str(self.id)])
 
+    def get_allowed_scopes_from_scopes(self, scopes):
+        """
+        Filters the given list of scopes so it only contains the allowed scopes
+        """
+        if isinstance(scopes, six.string_types):
```
clintonb: This can be simplified:

```python
# When would scopes NOT be a string?
scopes = scopes.split(' ')
scopes = set(scopes)

allowed_scopes = self.allowed_scopes or ''
allowed_scopes = allowed_scopes.split(' ')
allowed_scopes = set(allowed_scopes)

scopes = allowed_scopes & scopes
return ' '.join(scopes)
```

oauth2_provider/models.py
```diff
@@ -216,6 +238,8 @@ def allow_scopes(self, scopes):
         """
         if not scopes:
             return True
+        if isinstance(scopes, six.string_types):
```
clintonb: When is `scopes` not a string? Are there steps that can be taken to ensure that it is always a string?

oauth2_provider/tests/test_application_views.py
```diff
@@ -76,6 +77,12 @@ def setUp(self):
         self.app_bar_1 = self._create_application('app bar_user 1', self.bar_user)
         self.app_bar_2 = self._create_application('app bar_user 2', self.bar_user)
 
+    def test_get_application(self):
+        """Test the get_application method from the AuthorizationView"""
+        self.assertEqual(self.app_foo_1, AuthorizationView().get_application(self.app_foo_1.client_id))
+        # test twice to test the cache
```
clintonb: Unless you're checking DB query counts or calls to the cache, you aren't truly validating caching behavior.

oauth2_provider/tests/test_application_views.py
```diff
@@ -76,6 +77,12 @@ def setUp(self):
         self.app_bar_1 = self._create_application('app bar_user 1', self.bar_user)
         self.app_bar_2 = self._create_application('app bar_user 2', self.bar_user)
 
+    def test_get_application(self):
+        """Test the get_application method from the AuthorizationView"""
```
clintonb: It would be helpful if the docstring explained what about the method you are testing. For example, `Verify the method returns the application corresponding to the given client ID.`

oauth2_provider/tests/test_models.py
```diff
@@ -41,6 +41,77 @@ def test_allow_scopes(self):
         self.assertTrue(access_token.allow_scopes(['write', 'read', 'read']))
         self.assertTrue(access_token.allow_scopes([]))
         self.assertFalse(access_token.allow_scopes(['write', 'destroy']))
+        self.assertTrue(access_token.allow_scopes('read write'))
```
clintonb: This test is a great candidate for [parameterization](http://doc.pytest.org/en/latest/parametrize.html).

oauth2_provider/tests/test_models.py
```diff
@@ -41,6 +41,77 @@ def test_allow_scopes(self):
...
+        self.assertTrue(access_token.allow_scopes('write read'))
+        self.assertTrue(access_token.allow_scopes('write read read'))
+        self.assertTrue(access_token.allow_scopes(''))
+        self.assertFalse(access_token.allow_scopes('write destroy'))
+
+
+    def test_get_allowed_scopes_from_scopes(self):
```
clintonb: This test is also a great candidate for [parameterization](http://doc.pytest.org/en/latest/parametrize.html).

clintonb: This test also needs a docstring, and, based on the length (60 lines!), should be split into multiple test cases.

oauth2_provider/views/base.py
```diff
@@ -72,12 +72,13 @@ class AuthorizationView(BaseAuthorizationView, FormView):
     server_class = oauth2_settings.OAUTH2_SERVER_CLASS
     validator_class = oauth2_settings.OAUTH2_VALIDATOR_CLASS
     oauthlib_backend_class = oauth2_settings.OAUTH2_BACKEND_CLASS
+    _application_cache = {}
 
-    skip_authorization_completely = False
+    def getscopes(self):
```
clintonb: `get_scopes`

oauth2_provider/views/base.py
```diff
@@ -87,6 +88,14 @@ def get_initial(self):
         }
         return initial_data
 
+    def get_application(self, client_id):
+        """
+        get an application from a client_id, this caches the applications
+        """
+        if client_id not in self._application_cache:
```
clintonb: Suggested alternative:

```python
application = self._application_cache.get(client_id)

if not application:
  application = get_application_model().objects.get(client_id=client_id)
  self._application_cache[client_id] = application

return application
```

oauth2_provider/models.py
JensTimmerman: if it is not a string it has to be a list, so then there is no need to split it into a list.

clintonb: I get that. Looking more closely at the existing code and docstring, this method expects an iterable of strings. You are weakening that contract to allow for a single string. However, the only place where you are consuming this is in tests. This change seems unnecessary. To keep the contract clean, if you do use `allow_scopes`, I would ensure that any new usage sends an iterable of strings rather than a delimited string.

oauth2_provider/models.py
kevin-brown: Considering DOT already depends on `six` being available, it would make the most sense to just `import six` instead.

oauth2_provider/views/base.py
jleclanche: I don't think that's better. But that aside this cache can be pulled into its own PR.

clintonb:
@synasius @Girbons @JensTimmerman what is needed to get this merged?

JensTimmerman:
@clintonb I'm still awaiting a reply from security@evonove.it

synasius:
This is definitely a feature we want to merge but I do not like the current implementation. I've been working recently on a feature that adds support for scopes backends. You can see the code here https://github.com/evonove/django-oauth-toolkit/compare/scopes-backend and I wrote about it on a blog post: https://evonove.it/blog/django-oauth-toolkit-0110-released/

At the moment scopes can be changed through settings, but there is no way to implement a custom flow to obtain scopes (for instance I'd like to determine the available scopes based on the groups to which the user belongs).
So I introduced a scopes backend that expose an API to work with scopes (see https://github.com/evonove/django-oauth-toolkit/compare/scopes-backend#diff-8959a5c87f26bd1ebe41c94fa5d65c1cR7).
The default implementation uses the settings so it doesn't change the default DOT behavior (see https://github.com/evonove/django-oauth-toolkit/compare/scopes-backend#diff-8959a5c87f26bd1ebe41c94fa5d65c1cR40).

I'd love to base the App-specific scope feature on this new API.

What do you think about it?

clintonb:
@synasius this is great! It meets the immediate need, and allows for extensions that are not immediately known. When can we have it!? :)

honestbleeps:
I wish the distaste with my initial approach was voiced long ago, I would have gladly redone / reworked it in a better way...

in any case, I hope it eventually gets added.

jleclanche:
@honestbleeps I'm helping evonove maintain the library now, I'll be working with @synasius to land an improved version of this.

oauth2_provider/models.py
JensTimmerman: well a string is in fact an iterable, using it by accident might show up weird errors, this is a bit more developer friendly in that it will do the thing that you wanted to happen when you pass a string. But you are also right that it might encourage the use of a single string and confuse more people when they see inconsitent usage of 'scopes'

JensTimmerman:
@synasius Great, I like that approach even more! Hope to see more of it soon 👍

jaddison:
@synasius Great suggestion, improved flexibility. What needs to be done to move it forward?

clintonb:
@jleclanche how can we get this merged and released?

jleclanche:
I was waiting on @synasius to finish what he mentioned before touching this.

@synasius do you need help finishing it off/taking that part over?

jleclanche:
Ok, since @synasius isn't replying, I'd like to suggest taking over his design from the scopes-backend branch.

I really like the plan outlined above.  The backwards-compatible default scopes backend is also really cool.

I can't see at first glance what's missing from that commit, so I'll be testing it in a few days (I'm currently trying to release something else on the side so I'm unable to fully focus on this just yet). @clintonb, I'd love your help. If you can help testing the scopes-backend branch (which I have just rebased), verify what's left to implement, I can finish it off myself.

synasius:
@jleclanche :sweat_smile:  :sweat_smile:  :sweat_smile: 
The only thing I can say is "thank you!!".

synasius:
@jleclanche If it helps I can say that the code in `scopes-backend` branch it's used in a production application deployed months ago with no problems. Proper documentation is missing

jleclanche:
Excellent, good to know!

In that case I'll play around with it, write some documentation and land it hopefully this week (if you're reading this in the future and I haven't, nudge me).

Once it lands, I'm in favour of doing a 0.12 release. We can then look into builtin app-specific scopes functionality for a 1.0 release (Seeing that DOT is widely used, works well and we're already careful with backwards compatibility, I don't see much point in staying on 0.x much longer).

synasius:
> In that case I'll play around with it, write some documentation and land it hopefully this week (if you're reading this in the future and I haven't, nudge me).

:+1: 

> Once it lands, I'm in favour of doing a 0.12 release.

:+1: 

> We can then look into builtin app-specific scopes functionality for a 1.0 release (Seeing that DOT is widely used, works well and we're already careful with backwards compatibility, I don't see much point in staying on 0.x much longer).

Totally agree! We've been careful since 0.1 :smile:

JensTimmerman:
Great! Sounds really good to me.
So I guess I can just close this pr?

synasius:
Guess you can close it now and reference it later, after the `scopes-backend` branch has been merged

tximpa91:
Someone can upload an example of this beacause i am trying i cant doit . thanks

prafulbagai:
@tximpa91 - were you able to achieve it? @anyone - A simple example would be really helpful.

tximpa91:
Yes i managed to do it, working perfectly in production

prafulbagai:
@tximpa91 - Could please share a sample snippet of the same. I'm not able to understand the flow of it.

prafulbagai:
@tximpa91 - I've seen [this scope backed](https://github.com/uktrade/data-hub-leeloo/blob/develop/datahub/oauth/scopes.py). How do I connect it to oauth-toolkit?

PR #396, opened 2016-06-29
added an IsAuthenticatedOrTokenHasScope Permission

### LINKED ISSUE #265: Django REST Framework - Browseable API
According to [this question on stackoverflow](http://stackoverflow.com/questions/24387009/django-restframework-browsable-api-login-with-ouath) it should be possible to use the browseable API from the Django REST Framework package.

However, declaring the `SessionAuthentication` authentication class didn't work, I get an error that I'm lacking permission:

```
REST_FRAMEWORK = {
    'DEFAULT_AUTHENTICATION_CLASSES': ( 
        'oauth2_provider.ext.rest_framework.OAuth2Authentication', 
        'rest_framework.authentication.SessionAuthentication', 
    ), 
    'DEFAULT_PERMISSION_CLASSES': ('rest_framework.permissions.IsAuthenticated',),
    'PAGE_SIZE': 10,
    'TEST_REQUEST_DEFAULT_FORMAT': 'json',
}
```

These are my scopes:

```
OAUTH2_PROVIDER = {
    # this is the list of available scopes
    'SCOPES': {
        'read': 'Read scope',
        'write': 'Write scope',
        'foods': 'Access to Foods'
    }
}
```

and this is my view that cannot be accessed through the browseable API if the `TokenHasScope` permission class has been set:

```
class FoodGroupViewSet(ViewSetBase):
    permission_classes = [IsAuthenticated, TokenHasScope]
    required_scopes = ['foods']
    queryset = FoodGroup.objects.all()
    serializer_class = FoodGroupSerializer`
```

Is this a bug or am I doing anything wrong here?

JensTimmerman:
fixes #265 
based on code writen by  @werthen

we use this permission to allow people to browse and use the api, based on their model perimissions, and use the rest api based on the required scopes.
This means a user can browse everything he/she has permissions to browse, 
and an app with a token can get all the information the user has permissions to, as long as it's token has the correct scope (so the users granted the right permissions)
```
permission_classes = (IsAuthenticatedOrTokenHasScope, DjangoModelPermissions)
required_scopes = ['scope1']
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,3 +17,4 @@ Diego Garcia
 Bas van Oostveen
 Bart Merenda
 Paul Oswald
+Jens Timmerman
--- a/README.rst
+++ b/README.rst
@@ -100,6 +100,7 @@ Changelog
 Development
 ~~~~~~~~~~~
 
+* #396: added an IsAuthenticatedOrTokenHasScope Permission
 * #357: Support multiple-user clients by allowing User to be NULL for Applications
 
 0.10.0 [2015-12-14]
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,7 @@ Changelog
 Development
 ~~~~~~~~~~~
 
+* #396: added an IsAuthenticatedOrTokenHasScope Permission
 * #357: Support multiple-user clients by allowing User to be NULL for Applications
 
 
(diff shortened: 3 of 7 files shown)


synasius:
Hi @JensTimmerman 
PR is good but a test is missing.

Take a look at the [test_rest_framework.py](https://github.com/evonove/django-oauth-toolkit/blob/master/oauth2_provider/tests/test_rest_framework.py) module.

JensTimmerman:
as it turns out, an extra check was needed, setting a token will also always authenticate, so ignore users authenticated using the OAuth2 backend and check the scope for those requests.

JensTimmerman:
@synasius tests added

PR #399, opened 2016-07-10
Fix coverage since pytest

agriffis:
Coverage measurement has been broken since updating to pytest. See https://coveralls.io/repos/3350/builds?page=7 and https://travis-ci.org/evonove/django-oauth-toolkit/jobs/141547004, or the following screenshots

screenshot from 2016-07-10 12-57-43

screenshot from 2016-07-10 12-58-12

It's not completely clear WHY coverage broke, but it's simple to fix by updating to the latest pytest-cov and coverage.py

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,15 @@ python:
 sudo: false
 
 env:
-  - TOX_ENV=py27-django18
-  - TOX_ENV=py27-django19
-  - TOX_ENV=py32-django18
-  - TOX_ENV=py33-django18
-  - TOX_ENV=py34-django18
-  - TOX_ENV=py34-django19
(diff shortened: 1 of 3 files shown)


agriffis:
Also note that the coverage badge in the README is misleading. Since coverage measurement has been broken, the coveralls command hasn't been able to send updates to coveralls.io, so the badge has been frozen at 95%

This commit shows that coverage has actually improved since then to 97%

agriffis:
This PR should be all set now. Python 3.2 is excluded from the coverage measurement because of incompatibilities, but that should be fine. DOT should be able to fully drop Python 3.2 support at the end of 2016 to match the Django project.

See https://coveralls.io/builds/6943749 for the result.

PR #422, opened 2016-11-05
OpenID Connect Support

### LINKED ISSUE #110: OpenID Connect
Hello,

I was waiting for the approval of the OpenID Connect [spec](http://openid.net/connect/), and now it's approved, I wonder if you would like a pull request with that feature (the "OpenID connector" part on your roadmap refers to that, right?)

I'm already learning the [Core spec](http://openid.net/specs/openid-connect-core-1_0.html) to figure out how it can be implemented. It seems like a thin layer on top of OAuth2, for example the authentication part is done by a request to `/o/authorize` with "openid" as one of the scopes, and the token exchange now returns one more parameter, the ID Token.

There's two more (optional) specs, but my intent is implement the core on top of django-oauth-toolkit, since it seems enough to be an OpenID provider. I'm already reading django-oauth-toolkit's source code, but any tips are welcome.

So, what do you think?

clintonb:
This is a continuation of work started by @joelstevenson.

Fixes #110


clintonb:
We have decided to move away from OpenID Connect, and simply use OAuth 2.0 with JWT access tokens.

PR #436, opened 2016-12-10
Minor improvements

jleclanche:
This PR drops old Python 2.6 compatibility shims, adds better administration for the builtin models (including a crucial fix adding `access_token` to `raw_id_fields`) and splits the base and management views in `urls.py` so that either can be accessed and included separately - without breaking backwards compatibility.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,6 +97,11 @@ Roadmap / Todo list (help wanted)
 Changelog
 ---------
 
+0.12.0 [Unreleased]
+~~~~~~~~~~
+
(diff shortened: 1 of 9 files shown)


jleclanche:
Travis was failing due to a recent change in Django 1.11 master, which makes query dicts immutable. The latest commit fixes that by dropping a shim for backwards compatibility with pre-0.6.1.

jleclanche:
@synasius ping - i sent you an email as well. I'd really like to help the project move along, I don't want to be forking it :(

PR #437, opened 2016-12-15
Removed skip-authorization

Girbons:
This pr adds on `ApplicationUpdate` view the following method:

```python
def get_form_class(self):
```

Removed field `expires` from `RefreshTokenAdmin`

Removed unused import from tests


JensTimmerman:
Thank you for getting this merged in. Are you planning any new releases soon with this fix in it?

PR #442, opened 2017-01-17
Add on_delete to all ForeignKey and OneToOne fields

akheron:
This fixes most Django 2.0 warnings.


jleclanche:
Looks good thanks!

PR #446, opened 2017-01-23
Fix three typos

noahsark769:
Hey there! Just wanted to reach out and say this library is super awesome and useful. Thanks for creating it! When I was trying to figure out how AuthorizationView worked I noticed that there were some typos in the docstring, so I figured I would fix them.

Thanks again!


jleclanche:
Thanks!

Issue #444, opened 2017-01-20
Documentation on templates


NBajanca:
Hello,

I've recently opened and closed issue #439.
This happened because I couldn't find any information on Templates.

Even though this can be solved by examining the template folder in the project I think some documentation on this, like [Django has for authentication](https://docs.djangoproject.com/en/1.10/topics/auth/default/#all-authentication-views) wouldn't hurt.

Let me know if this makes sense and I can write them.

jleclanche:
Docs always welcome! Feel free to PR some.

PR #447, opened 2017-01-26
Docs on templates

NBajanca:
**Issue**: #444 

I added ``templates.rst`` to the ``docs``.
I gave complete docs on the templates because the Views' docs are automatic and I think this way it is easier to understand how everything was implemented (simple version).

I also enabled the [intersphinx extension](http://www.sphinx-doc.org/en/1.5.1/ext/intersphinx.html) so I could map python and Django classes.

There were also some warnings on compile ("title length must match underline") and I got rid of them.
Changes in other files were only for this reason and to add references.

Feel free to change anything.

### DIFF
--- a/docs/advanced_topics.rst
+++ b/docs/advanced_topics.rst
@@ -1,6 +1,7 @@
 Advanced topics
 +++++++++++++++
 
+.. _extend_app_model:
 
 Extending the Application model
 ===============================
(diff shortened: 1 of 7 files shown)


jleclanche:
Looks good. Travis issue is unrelated. Thanks!

Issue #69, opened 2013-10-10
Grant type: All-in-one gives a "client is not authorized to use response_type token"


wezzynl:
When setting-up an Application with the "All-in-one generic" authorization grant type, I cannot make an authorization request with response_type of 'token' (or 'code' for that matter...)

Maybe I'm doing something wrong but It looks like the function 'validate_response_type' in 'oauth2_provider.oauth2_validators.py' doesn't take GRANT_ALLINONE into account?

yarbelk:
+1 for this - I have a use case where I have a resource (tied to an ApplicationModel) that needs to be editable by a secure sever and public clients, with different scopes.  Permissions get really squirrly if I have to tie two Application instances with the two grant types to to the resource.  Some users can use one flow, some use another.

jarcoal:
It would be really nice to see this supported.  I think that at some point most apps are going to need an Application that can support multiple flows.

Has anyone found a good work-around or is creating/maintaining 2+ applications the best way to go?

jarussi:
Sorry for resurrecting a dead thread, but I also need this. Anyone found a solution / workaround for it?

clintonb:
Any updates, @masci? What was the reason the all-in-one functionality was removed? Would you entertain reviewing/merging a restoration of similar functionality?

clintonb:
@synasius @Girbons any interest in a PR restoring this functionality?

jleclanche:
@clintonb If you can lay out a design doc + plan I can review that.

clintonb:
@jleclanche no need. https://github.com/evonove/django-oauth-toolkit/pull/448 meets my needs, and should probably be suitable for others.

jleclanche:
Cool, closing then.

PR #448, opened 2017-01-30
Customizable application grant type

ddelemeny:
Hello,

Using DOT in my app, I wanted to customize my Application model to support multiple (not combined) grant flows. If I understand correctly, this is @skrivanos's request on #69 .

This PR implements a new api method on AbstractApplication  to check if the instance allows a grant type and just a few modifications on the validator so that it doesn't make assumptions on Application.authorization_grant_type and asks the application model instead.

This is transparent for the existing tests and does not decrease coverage as far as I can tell.
It doesn't seem to contradict the RFC as it doesn't forbid multiple grant flows to be allowed for one client : 
[RFC 6749 section 1.2](https://tools.ietf.org/html/rfc6749#section-1.2)
> The client receives an authorization grant, which is a
> credential representing the resource owner's authorization,
> expressed using one of four grant types defined in this
> specification or using an extension grant type.  The
> authorization grant type depends on the method used by the
> client to request authorization and the types supported by the
> authorization server.

However, it may break existing custom Application models by introducing a new requirement on its API, so the PR probably needs more love to avoid this breakage.

Can someone please review and comment ?

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -125,6 +125,9 @@ class AbstractApplication(models.Model):
     def __str__(self):
         return self.name or self.client_id
 
+    def allows_grant_type(self, *grant_types):
+        return self.authorization_grant_type in grant_types
+
 
 class Application(AbstractApplication):
     class Meta(AbstractApplication.Meta):
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -263,7 +263,7 @@ class OAuth2Validator(RequestValidator):
         Validate both grant_type is a valid string and grant_type is allowed for current workflow
         """
         assert(grant_type in GRANT_TYPE_MAPPING)  # mapping misconfiguration
-        return request.client.authorization_grant_type in GRANT_TYPE_MAPPING[grant_type]
+        return request.client.allows_grant_type(*GRANT_TYPE_MAPPING[grant_type])
 
     def validate_response_type(self, client_id, response_type, client, request, *args, **kwargs):
         """
@@ -271,9 +271,9 @@ class OAuth2Validator(RequestValidator):
         rfc:`8.4`, so validate the response_type only if it matches 'code' or 'token'
         """
         if response_type == 'code':
-            return client.authorization_grant_type == AbstractApplication.GRANT_AUTHORIZATION_CODE
+            return client.allows_grant_type(AbstractApplication.GRANT_AUTHORIZATION_CODE)
         elif response_type == 'token':
-            return client.authorization_grant_type == AbstractApplication.GRANT_IMPLICIT
+            return client.allows_grant_type(AbstractApplication.GRANT_IMPLICIT)
         else:
             return False
 
--- a/oauth2_provider/tests/test_oauth2_backends.py
+++ b/oauth2_provider/tests/test_oauth2_backends.py
@@ -13,7 +13,7 @@ class TestOAuthLibCoreBackend(TestCase):
         self.factory = RequestFactory()
         self.oauthlib_core = OAuthLibCore()
 
-    def test_swappable_serer_class(self):
+    def test_swappable_server_class(self):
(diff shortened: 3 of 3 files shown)


jleclanche:
I'm good with this. @clintonb what do you think of the API?

oauth2_provider/models.py
```diff
@@ -125,6 +125,9 @@ def get_absolute_url(self):
     def __str__(self):
         return self.name or self.client_id
 
+    def allows_grant_type(self, *grant_types):
+        return self.authorization_grant_type in grant_types
```
clintonb: Add a docstring explaining the purpose of this method, and its intended usage.

mattau600:
We are also looking for a solution where we can combine different grant types into a single app, which can simplify the oauth client.  Our app has endpoints that requires authentication via username/password, but also read-only endpoints that simply require token granted via client credentials.   As @ddelemeny mentioned, this conforms with the specification, and its been done in other oauth implementation.   Having the client store 2 sets of client id/secret due to deficiencies in the implementation is unfortunate.

When will this request be merged?  I am hoping it won't die like #69 did.  Thanks so much in advance!

ddelemeny:
Hi, thanks everyone for your interest in this issue. I'll complete the PR
as soon as I have some spare time to do it, but right now I'm under a heavy
workload.

@clintonb : yes, your understanding is correct. I can provide some minimal
usage example for the documentation if necessary. What do you think about
the breaking API requirement I mentioned ? Should it be addressed
gracefully ?

Le 7 févr. 2017 12:17, "Matt Au" <notifications@github.com> a écrit :

We are also looking for a solution where we can combine different grant
types into a single app, which can simplify the oauth client. Our app has
endpoints that requires authentication via username/password, but also
read-only endpoints that simply require token granted via client
credentials. As @ddelemeny <https://github.com/ddelemeny> mentioned, this
conforms with the specification, and its been done in other oauth
implementation. Having the client store 2 sets of client id/secret due to
deficiencies in the implementation is unfortunate.

When will this request be merged? I am hoping it won't die like #69
<https://github.com/evonove/django-oauth-toolkit/issues/69> did. Thanks so
much in advance!

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/evonove/django-oauth-toolkit/pull/448#issuecomment-277969645>,
or mute the thread

clintonb:
@ddelemeny I don't see where/how you break the API. You've added a custom method with a sane default that is compatible with existing behavior.

jleclanche:
k. I'm good merging this. @clintonb if I land it do you want to PR some documentation?

ddelemeny:
Custom applications will require the allows_grant method to be implemented.
Any implementation which doesn't inherit from AbstractApplication will
break because of this new requirement.

Le 7 févr. 2017 15:09, "Clinton Blackburn" <notifications@github.com> a
écrit :

clintonb:
@ddelemeny the documentation makes no claims of support for models that do not extend `AbstractApplication`: https://django-oauth-toolkit.readthedocs.io/en/latest/advanced_topics.html. It is unfortunate that this change may break those users' projects; however, that is the risk you run when you do things that are not explicitly supported.

@jleclanche yes, I can submit a PR by next week.

mattau600:
Thanks guys for getting this merged so quickly.  I don't mean to be pushy, but when can we expect that 0.12.0 to be released so we can try it out?

jleclanche:
Id like to land the scopes-backend branch first (which should be trivial). But in any event I can't do releases myself, I don't have access.

jleclanche:
`scopes-backend` has landed in 7071c0e6861c41d26ab63bd4a4a5a575414732aa!

mattau600:
Who has permission to do a release?  Is there enough stuff in 0.12 to do a release, so we can start using it?

jleclanche:
That would be @synasius. I'm OK with a 0.12 release yeah, I think we're in a good state.

mattau600:
Thanks @jleclanche @synasius ! This feature reduces complexity for our clients integration, so we certainly would appreciate it!

PR #456, opened 2017-02-27
Ignoring requests containing list instead of dict

sassanh:
I had problems sending list as json object to DRF, oauth2_provider were raising this error:

```
File ".../lib/python3.6/site-packages/oauth2_provider/oauth2_backends.py", line 183, in extract_body
    body = json.loads(request.body.decode('utf-8')).items()
AttributeError: 'list' object has no attribute 'items'
```

It's solved by this commit.

### DIFF
--- a/oauth2_provider/oauth2_backends.py
+++ b/oauth2_provider/oauth2_backends.py
@@ -180,9 +180,11 @@ class JSONOAuthLibCore(OAuthLibCore):
         :return: provided POST parameters "urlencodable"
         """
         try:
-            body = json.loads(request.body.decode("utf-8")).items()
-        except AttributeError:
-            body = ""
+            data = json.loads(request.body.decode('utf-8'))
+            if isinstance(data, dict):
+                body = data.items()
+            else:
+                body = ""
         except ValueError:
             body = ""
 


jleclanche:
That doesn't look like the right fix. Can you add a test?

sassanh:
What about this one?

sassanh:
I think `oauth2_provider` shouldn't try to parse all requests even those not sent to one of its endpoints. It should only parse requests sent to its endpoints.

jleclanche:
That is most definitely not the right fix. You're passing around a list of tuples in one instance, an empty string in another.

What's the issue anyway?

jleclanche:
Please give me a test case so I can reproduce it.

oauth2_provider/oauth2_backends.py
```diff
@@ -180,7 +180,11 @@ def extract_body(self, request):
...
         """
         try:
-            body = json.loads(request.body.decode('utf-8')).items()
+            data = json.loads(request.body.decode('utf-8'))
+            if isinstance(data, dict):
+                body = data.items()
+            else:
+                body = ""
         except ValueError:
             body = ""
```
sassanh: @jleclanche It was already there I didn't change it.

sassanh:
@jleclanche I don't get what you mean. I just changed this line:

    body = json.loads(request.body.decode('utf-8')).items()

to this

    data = json.loads(request.body.decode('utf-8'))
    if isinstance(data, dict):
        body = data.items()
    else:
        body = ""

That's the only changed I made. I'm not passing around a litle of tuples and the empty string is what already were there. See my inline comment.

sassanh:
I'll prepare the test.

sassanh:
The issue is I'm using `django-oauth-toolkit` with `django-rest-framework`, I wanna send a bulk create request to `django-rest-framework` something like:

    curl -X POST 'http://localhost:8000/api/user/' -d '[{"firstname":"John","lastname":"Oliver"},{"firstname":"X","lastname":"Y"}]' -H 'content-type: application/json'

When I send this request it never reaches to `django-rest-framework` because `django-oauth-toolkit` raises an exception in the lines I edited above. Because `django-oauth-toolkit` is trying to parse that request while it's not even sent to `/o/token/`. `django-oauth-toolit` parses every request no matter what's the endpoint. That's the issue.

symptog:
@sassanh Every Request is handled by the Authentication Backend and therefor every request is handled by `OAuthLibCore.verify_request()`

:+1: for the Pull-Request.

sassanh:
@symptog I don't understand why it needs to parse body of all requests, I'd be thankful if you could explain it. As far as I know it just needs to check request headers to authorize user, am I missing something?

symptog:
The OAuth2 RFC[rfc6749](https://tools.ietf.org/html/rfc6749#page-48) doesn't describe how the token SHOULD be passed to the server. In the example they are using the Bearer and MAC method. But the token can also be passed as GET parameter or encoded in the body e.g as a key in a JSON object.
The (oauthlib)[https://oauthlib.readthedocs.io/en/latest/oauth2/endpoints/resource.html] tries to handle these cases.

oauth2_provider/oauth2_backends.py
```diff
@@ -180,7 +180,11 @@ def extract_body(self, request):
         :return: provided POST parameters "urlencodable"
         """
         try:
-            body = json.loads(request.body.decode('utf-8')).items()
+            data = json.loads(request.body.decode('utf-8'))
+            if isinstance(data, dict):
+                body = data.items()
+            else:
+                body = ""
```
menecio: Why do you set an empty string if not getting a dictionary?

sassanh: This is the original code:

https://github.com/evonove/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_backends.py#L182

It's setting empty string when it can't parse a JSON value. So empty string seems to be safe. When body is valid JSON but not a dictionary the old code would raise an exception (not intentionally) as it couldn't run `.items()` on a Python list. So if you use oauth-toolkit you can't send lists as body of your request even when it's absolutely irrelevant to oatuh-toolkit. The above code behaves with lists just similar to invalid JSONs as oauth-toolkit can't consume lists. But it shouldn't raise an exception. It should just ignore it.

symptog: this body will be passed to the oauthlib for validation. this validator tries look up an access token key in the body. If we pass a list to the oauthlib we currently receive an error, because it is unable to parse the list. As a list possibly contains no access token, we can skip the body and set it to an empty string.

menecio: You could also solve this by catching `AttributeError` on the except block and won't need to add any extra code

```
try:
    body = json.loads(request.body.decode("utf-8")).items()
except AttributeError, ValueError:
    body = ""
```

it's simpler, don't you think?

sassanh: Look at my first commit: https://github.com/evonove/django-oauth-toolkit/pull/456/commits/8646803b4876aac87cdd86b0e7581aee1fe129ec

menecio: Oh I see, well you could merge them as I proposed. Wonder why @jleclanche didn't like this approach ?

sassanh: @menecio `except AttributeError, ValueError:` is a syntax error.

PR #459, opened 2017-03-09
Make Resource Owner model configurable

ddelemeny:
Hello there,

Following [my (unanswered) thread on google groups](https://groups.google.com/forum/#!topic/django-oauth-toolkit/CAbH1sn7bKI) regarding the possibility to have another model than Django's AUTH_USER_MODEL as the Resource Owner for tokens (I've used django-organizations OrganizationUser as an example), I've been experimenting a little bit with the idea.

This PR attempts to implement a configurable setting allowing developers to change the model considered as Resource Owner for Grants, AccessToken and RefreshToken.

Usage:
- set OAUTH2_PROVIDER_RESOURCE_OWNER_MODEL setting as the import path of the desired Resource Owner Model. ex  ```OAUTH2_PROVIDER_RESOURCE_OWNER_MODEL = 'myapp.MyResourceOwner'```
- Override OAuthLibCore with a custom ```_extract_resource_owner``` method, to get a resource owner instance from the request

It's been a bit hard for me to follow the path of a request in the multiple reciprocal calls between DOT and OAuthLib, so I don't really know how far I am from a complete implementation. This patch works for my use case and passes existing tests.

What do you think ? Is it a sane way to achieve this kind of flexibility ? Is it desirable in DOT ?
Please review and comment.


jleclanche:
I'm torn. On the one hand this a pretty common pattern, but on the other hand we've talked about making all the token models etc abstract so that you can swap them out for your own version.

I'm +0 on this. We could have both.

ddelemeny:
@jleclanche I understand, I had a closer look at #252 and #347 and it would indeed make it easier to swap in this pattern. 
Having my proposal implemented in DOT or left in a cookbook for the expected abstract models is fine by me, as long as it it made possible.

What would be needed to make either solution go forward ?

jleclanche:
I think they're not mutually exclusive. We *definitely* need swappable models but #347 is just not mergeable. The PR here is more reasonable. I'm happy to review it but I need to find the time, and I don't want to be the only one making this decision. @clintonb can you chime in?

oauth2_provider/oauth2_backends.py
```diff
@@ -106,14 +115,14 @@ def create_authorization_response(self, request, scopes, credentials, allow):
         :param scopes: A list of provided scopes
         :param credentials: Authorization credentials dictionary containing
                            `client_id`, `state`, `redirect_uri`, `response_type`
-        :param allow: True if the user authorize the client, otherwise False
+        :param allow: True if the resource owner authorize the client, otherwise False
```
clintonb: s/authorize/authorized

oauth2_provider/tests/test_models.py
```diff
@@ -7,16 +7,20 @@
...
+from ..models import get_application_model, get_resource_owner_model
+from ..models import Grant, AccessToken, RefreshToken
 
 
 Application = get_application_model()
+ResourceOwnerModel = get_resource_owner_model()
 UserModel = get_user_model()
 
 
+
```
clintonb: Nit: This extra newline is not necessary.

oauth2_provider/tests/test_models.py
```diff
@@ -7,16 +7,20 @@
 from django.test.utils import override_settings
 from django.utils import timezone
 
-from ..models import get_application_model, Grant, AccessToken, RefreshToken
+from ..models import get_application_model, get_resource_owner_model
+from ..models import Grant, AccessToken, RefreshToken
```
clintonb: These two lines can be combined.

oauth2_provider/tests/test_models.py
```diff
@@ -177,3 +182,12 @@ class TestRefreshTokenModel(TestCase):
...
         self.assertEqual("%s" % refresh_token, refresh_token.token)
+
+
+@override_settings(OAUTH2_PROVIDER={'RESOURCE_OWNER_MODEL':'tests.TestResourceOwner'})
+class TestCustomResourceOwnerModel(TestCase):
+    def setUp(self):
+        self.user = UserModel.objects.create_user("test_user", "test@user.com", "123456")
+
+    def test_model(self):
+        self.user.resource_owners.all()
```
clintonb: Add a comment explaining what you're testing here.

menecio:
@jleclanche maybe you will be interested on reviewing this one? https://github.com/evonove/django-oauth-toolkit/pull/467

menecio:
As the swappable models feature is already pushed to master, could we work on this? I can help you if needed.

jleclanche:
I'll always take docs PRs. I still think they're not mutually exclusive, so if you guys still consider this wanted, it'll have to be rebased.

ddelemeny:
This kind of stalled on my end. The migration I pushed doesn't work when
the FK changes type with a db that has proper constraints (it fails
pointing to an uuid pk'ed model in postgres for example, the type cannot be
upcasted) , and the migration process isn't really complete.

Le 28 avr. 2017 22:11, "Jerome Leclanche" <notifications@github.com> a
écrit :

jleclanche:
Closing since this has stalled for too long and isn't ready to land.

PR #475, opened 2017-05-14
add missing migrations

symptog:
some migrations are missing
further work should be based on a clean codebase


jleclanche:
This doesn't look correct, shouldnt it have the swappable dependency?

symptog:
Yep, I think you're right. I can't find any information on migrating abstract models.
"if in doubt, leave it out" :D

PR #476, opened 2017-05-15
Disallow empty redirect URIs

rigaspapas:
The empty string does no longer pass the URI validator. This does not
make existing code that relies on `blank=True' fail, it just fixes the
case where the user provides an empty string.

Signed-off-by: Rigas Papathanasopoulos <rigas@arrikto.com>


jleclanche:
Thank you!

PR #477, opened 2017-05-18
Implement rfc7662

### LINKED ISSUE #342: RFC 7662 OAuth 2.0 Token Introspection
Is anyone working to add this to django-oauth-toolkit?

https://tools.ietf.org/html/rfc7662

symptog:
Implement RFC 7662 for Django OAuth Toolkit
closes #342


jleclanche:
Could you squash what needs to be squashed please? One commit per logical feature, no "fix previous commit". That'll make it easier to review and merge.

There's also some commits changing things irrelevant to the PR; it's fine to include cleanups, but please make them separate commits as well :)

JoshBroomberg:
@jleclanche is there a plan to merge this? My organization wants to implement OAuth2 with separated resource/auth servers and this seems to be required to do that. Is there another way?

jleclanche:
@JoshBroomberg I don't have time to work on it myself, but if you want to take the PR in its current state and clean it up so I can review it and land it.

JoshBroomberg:
@symptog do you plan on working on this? We can assign some dev time to clean up and merge if you're happy with that. @jleclanche are there any other problems beyond those you list above?

jleclanche:
Sorry, actually @symptog did squash, I just never got notified about it. I'll do a review.

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+            content = resp.json()
+        except requests.exceptions.RequestException as e:
+            log.error("Introspection: failed to lookup token: %s" % e.message)
+        except ValueError as e:
+            log.error("Introspection: failed to parse response: %s" % e.message)
+        else:
+            if 'active' in content and content['active'] is True:
+                if 'username' in content:
+                    user, _created = UserModel.objects.get_or_create(username=content['username'], defaults={
+                        'password': UserModel.objects.make_random_password(16)
```
jleclanche: What's happening here? If you want to create a user that cannot be logged into, don't create a password at all.

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+        try:
+            bearer = 'Bearer {}'.format(oauth2_settings.RESOURCE_SERVER_AUTH_TOKEN)
+            resp = requests.post(url=oauth2_settings.RESOURCE_SERVER_INTROSPECTION_URL,
+                                 data={'token': token},
+                                 headers={'Authorization': bearer})
+            content = resp.json()
+        except requests.exceptions.RequestException as e:
+            log.error("Introspection: failed to lookup token: %s" % e.message)
+        except ValueError as e:
+            log.error("Introspection: failed to parse response: %s" % e.message)
```
jleclanche: Use `log.exception()` without e.message, that will let you keep the whole traceback.

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
     def get_default_redirect_uri(self, client_id, request, *args, **kwargs):
         return request.client.default_redirect_uri
 
+    def _get_token_from_authentication_server(self, token):
```
jleclanche: Instead of this, have the token and the url be passed as function arguments instead of fetching them from the settings (as you already do in the caller).

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+        try:
```
jleclanche: This whole try block should be split up. Here it is cleaned up:

```py
bearer = "Bearer {}".format(token)
try:
    response = requests.post(introspection_url, data={"token": token}, headers={"Authorization": bearer})
except requests.exceptions.RequestException:
    return

try:
    content = response.json()
except ValueError:
    log.exception("Introspection: Failed to parse response as json")
    return

if "active" in content and content["active"]: ...
```

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+            resp = requests.post(url=oauth2_settings.RESOURCE_SERVER_INTROSPECTION_URL,
+                                 data={'token': token},
+                                 headers={'Authorization': bearer})
+            content = resp.json()
+        except requests.exceptions.RequestException as e:
+            log.error("Introspection: failed to lookup token: %s" % e.message)
+        except ValueError as e:
+            log.error("Introspection: failed to parse response: %s" % e.message)
+        else:
+            if 'active' in content and content['active'] is True:
```
jleclanche: don't `if x is True`, just `if x`.

oauth2_provider/oauth2_validators.py
```diff
@@ -237,6 +239,57 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+                if 'username' in content:
+                    user, _created = UserModel.objects.get_or_create(username=content['username'], defaults={
```
jleclanche: You should use `UserModel.USERNAME_FIELD` instead of hardcoding a `username` lookup on the user model.

tests/test_introspection_auth.py
```diff
@@ -0,0 +1,169 @@
...
+
+@override_settings(ROOT_URLCONF=__name__)
+class TestTokenIntrospectionAuth(TestCase):
+    """
+    Tests for Authorization through token introspection
+    """
+    def setUp(self):
+        self.validator = OAuth2Validator()
+        self.request = mock.MagicMock(wraps=Request)
+        self.resource_server_user = UserModel.objects.create_user("resource_server", "test@user.com", "123456")
```
jleclanche: Use `test@example.com` for no risk of real emails leaking.

jleclanche:
Well it looks good. But I can't judge the functionality seeing as I'm not familiar with the RFC.

I reviewed what I could. Most issues are minor, the main problem I see is with the user passwords being created for no apparent reason.

@JoshBroomberg i would very much appreciate you testing this PR and reporting back confirming whether it works as you expect. It would go a long way towards merging it.

oauth2_provider/oauth2_validators.py
symptog: :+1:

oauth2_provider/oauth2_validators.py
symptog: :+1:

oauth2_provider/oauth2_validators.py
symptog: Part of splitting up the `try`-block

oauth2_provider/oauth2_validators.py
symptog: Won't fix. This value must be a Boolean.
[RFC7662 2.2](https://tools.ietf.org/html/rfc7662#section-2.2)
> active
       REQUIRED.  Boolean indicator of whether or not the presented token
       is currently active. [...]

oauth2_provider/oauth2_validators.py
symptog: this will throw a `SyntaxError`
```
>>> User.objects.get_or_create(User.USERNAME_FIELD='foobar')
  File "<console>", line 1
SyntaxError: keyword can't be an expression
```
How to fix this?

oauth2_provider/oauth2_validators.py
symptog: :+1:

Didn't thought it is possible to create users without password.

tests/test_introspection_auth.py
symptog: :+1: 

you should fix the other tests too ;)

oauth2_provider/oauth2_validators.py
symptog:
```

jleclanche:
@symptog Why is `AccessToken.application` being made nullable?

symptog:
@jleclanche the Resource Server doesn't now about the applications on the Authentication Server, but creating a Access Token required an Application. Therefore I made `AccessToken.application` nullable so the RS can create Tokens without application.

jleclanche:
@symptog Is there a way we can avoid that? I admittedly don't understand the problem space well enough to take a hard-line decision, but this sounds like a workaround rather than a solution.

jleclanche:
I cleaned up the PR a bit. I want to land this, but the AccessToken change is giving me pause.

symptog:
@jleclanche The RS and AS are different servers running DOT, but not sharing the same database.

The User/App authenticates itself on the AS, but the resources are requested on the RS.
As the RS and AS don't share their DB, the RS don't know any Token or Application (or I would need to configure Application on both servers). But the RS need to verify the token the request is called with. For that you could use JWT, OpenID Connect ... or the RFC 7662.
In RFC 7662 the RS sends the token to the AS and the AS respond if the token is valid. The RS can now decide whether to allow or deny the request.
There is no need for the RS to know about the application. I'm using the existing DOT infrastructure to create Tokens (without Applications) and verifying the request.

jleclanche:
Alright. Landing, thanks a bunch!

symptog:
Thank you :D

JoshBroomberg:
@symptog thanks for this. We will report back on our implementation of this over the next few weeks.

manelclos:
Hi! finally got to use this today. It works great, thanks a lot to everyone, the functionality is awesome!

I have some comments:

- The documentation for Resource Server is NOT online: http://django-oauth-toolkit.readthedocs.io/

- Documentation advises to add the "introspection" scope to SCOPES, but that means any application can get such permission and in fact it is assigned by default if you don't define DEFAULT_SCOPES and the application does not specify scope parameter in the authorize view. I removed it from SCOPES and everything seems to work as expected, anyway the token for the Resource Server is created manually and I assigned the introspection scope manually too. Maybe I'm missing something here?

- Sometimes you use introspect, others introspection, specially the scope is named "introspection" (noun) where I was expecting "introspect" (verb), like in "write". This is just curiosity on my part :)

Thanks again!

symptog:
I'm sorry for confusing these terms.  Maybe you can change it as desired and create a PR.

The `DEFAULT_SCOPES ` and `SCOPES` should be defined. I did not change the defaults, it's only an example in the docs. And the token for the RS needs to be created manually. So, you're not missing anything.

PR #484, opened 2017-06-06
make it i18n and add zh_hans locale file. 

gikoluo:
Add name for model fields, so the admin page shows the locale name.


jleclanche:
Closing as there's too many conflicts now so it's clear this won't get merged.

PR #489, opened 2017-06-16
Add signals when a token is generated

samupl:
This PR adds a very simple signal that fires up each time a token has been generated for an application.

Now I don't know if this is the right place to have such signal, but it looks like the right one (oauthlib isn't django-specific, so it didn't seem a good place to put it there in `create_token_response`).

Please let me know if it's a feature that you'd like to see in the upstream - if so, I will write additional unit tests.

### DIFF
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -192,7 +192,7 @@ class TokenView(OAuthLibMixin, View):
                     token=access_token)
                 app_authorized.send(
                     sender=self, request=request,
-                    application=token.application)
+                    token=token)
         response = HttpResponse(content=body, status=status)
 
         for k, v in headers.items():
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -14,7 +14,7 @@ from .mixins import OAuthLibMixin
 from ..exceptions import OAuthToolkitError
 from ..forms import AllowForm
 from ..http import HttpResponseUriRedirect
-from ..models import get_access_token_model, get_application_model, AccessToken
+from ..models import get_access_token_model, get_application_model
 from ..scopes import get_scopes_backend
 from ..settings import oauth2_settings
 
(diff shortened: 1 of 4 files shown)


samupl:
I wonder, however, if simply binding to `oauth2_validators.OAuth2Validator._create_access_token` or even `oauth2_validators.OAuth2Validator.save_bearer_token` and just trigger the signal when the token is created, what do you think?

oauth2_provider/views/base.py
```diff
@@ -183,6 +185,13 @@ class TokenView(OAuthLibMixin, View):
     @method_decorator(sensitive_post_parameters("password"))
     def post(self, request, *args, **kwargs):
         url, headers, body, status = self.create_token_response(request)
+        if status == 200:
+            access_token = json.loads(body).get('access_token')
+            if access_token is not None:
+                token = AccessToken.objects.get(token=access_token)
```
menecio: `AccessToken` is a swappable model, you need to instantiate is using `get_access_token_model`

samupl: Applied, thanks

jleclanche:
This looks great, could you add documentation?

samupl:
@jleclanche Here you go ;)

samupl:
I did a small change to the code - instead of passing the application down to the signal, I'm now passing the token, as it contains more useful information

samupl:
So, anything else is needed in this PR? Can it be safely merged?

jleclanche:
Landed in 18bbdc177f5f68f7db173bd3b883787418816a81 - thank you for your contribution!

adam-tokarski:
When this change will be deployed to pip?

jleclanche:
I don't think any time soon seeing as it's the only change since 1.0. Tip: you can do `pip install https://github.com/evonove/django-oauth-toolkit/archive/master.zip` to install the latest master. Replace `master` with a commit hash if you want to bind to a commit:

```
pip install https://github.com/evonove/django-oauth-toolkit/archive/18bbdc1.zip
```

adam-tokarski:
Yes, I was installing it from the latest master directly. Thanks for response!

PR #501, opened 2017-07-20
Parse URI only once in AbstractApplication.redirect_uri_allowed.

fladi:
Just a minor inconvenience but it seems unnecessary to me to parse the same
parameter value multiple times in case there are more than one permitted redirect
URIs.


jleclanche:
Thanks! Sorry for the delay.

PR #505, opened 2017-08-16
Includes required scopes in 403 response

### LINKED ISSUE #504: Include required scopes in error response
To make it easier for a client to recover from a lack of scopes
it would be nice to get the required scopes back in the error response

Something like:

```http
HTTP/1.0 403 Forbidden
Content-Type: application/json

{
  "required_scopes": ["foo:write"],
  "detail": "You do not have permission to perform this action."
}
```

simonschmidt:
Requires django rest framework >=3.5.0

Makes it easier for a client to know which extra scopes to request

Resolves #504


jleclanche:
Thank you!

PR #512, opened 2017-10-03
Add client_id as a natural key to the default Application model

jdp:
Uses client_id as a natural key when serializing the default
Application model, so that when things like fixtures are loaded,
references are maintained even when the environment that the
fixture is being loaded to doesn't have the same primary keys as
the environment that the fixture dumped from.


jleclanche:
Thank you! Sorry for the delay merging.

PR #515, opened 2017-10-12
Pass request to Django's auth.authenticate function

knaperek:
Since Django 1.11 the request parameter was added and it is now passed to authentication backends.

https://docs.djangoproject.com/en/1.11/topics/auth/default/#django.contrib.auth.authenticate

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -514,7 +514,7 @@ class OAuth2Validator(RequestValidator):
         """
         Check username and password correspond to a valid and active User
         """
-        u = authenticate(username=username, password=password)
+        u = authenticate(request=request, username=username, password=password)
         if u is not None and u.is_active:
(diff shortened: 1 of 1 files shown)


jleclanche:
This will break compat with 1.10, I'm hesitant to merge before 2.0 is out.

knaperek:
I don't think adding another parameter to authenticate() would break 1.10. Authentication backends should already accept any additional parameters (that may be used by other auth backends).

However, what is much worse, I just realized this is all wrong and unfortunately I don't see any obvious way to support it. The problem is that `request` is not Django request, but it is an instance of `oauthlib.common.Request`. I've been playing around a bit but don't see any way to get actual Django request in this context which makes me very sad :-(

jleclanche:
You're right. It looks like the handoff point is in OAuthLibCore.verify_request; this is where the django request object gets turned into an oauthlib request object.

Issue #240, opened 2015-04-29
OAUTH2_PROVIDER_APPLICATION_MODEL default not loaded into settings


yarbelk:
I have a migration that depends on the application model being there - but since I'm not overriding the applicaiton model, I don't see the OAUTH2_PROVIDER_APPLICATION_MODEL in the settings.

Using django-configurations as well.

jkkjonah:
Hi yarbelk, did you ever find a solution to this issue? We're experiencing the same thing after migrating from django 1.5 to 1.7

jdp:
Seeing this in Django 1.9 when running `makemigrations`. Not trying to use a custom model or anything, but one of my models has a foreign key to `oauth2_provider.Application`.

> AttributeError: 'Settings' object has no attribute 'OAUTH2_PROVIDER_APPLICATION_MODEL'

Adding the setting fixes it, but it seems heavy-handed to fix it like that.

b35li:
@jdp How do you fix the settings, I get the error right now , but no idea how to fix it.

jdp:
@b35li copy the setting into your settings.py, like this:

``` python
OAUTH2_PROVIDER_APPLICATION_MODEL = 'oauth2_provider.Application'
```

jleclanche:
Closing as these versions of Django are no longer supported.

dv321:
@jleclanche  this is still happening with django 1.10.7, can you point me in the direction of what would need to be done to get a pr in for this?

apiraino:
Hi, as @jleclanche said this undocumented behaviour looks like is still there.

I'm running Django 1.11.5 and I think it's a problem of circular imports that I still didn't figure out. The issue is triggered by a deadly combo of these params I'm using in my tests:
`-n`: I've installed [Django Test Without Migrations](https://github.com/henriquebastos/django-test-without-migrations)
`-v 2`: with verbose level set to 1 or 0, the issue does not show up.

Example of complete cmd line I'm using:
`$ DJANGO_SETTINGS_MODULE=myprj.settings.test python manage.py test --noinput -n myprj.module.tests.test_class -v 2`

My workaround is to set in my projects settings the two following constants:
```
OAUTH2_PROVIDER_ACCESS_TOKEN_MODEL = 'oauth2_provider.AccessToken'                                                                         
OAUTH2_PROVIDER_APPLICATION_MODEL = 'oauth2_provider.Application'                                                                          
```

lcary:
[...]

PR #517, opened 2017-10-12
Limit max len of dbindex fields to 191 for mysql utf8mb4 compatibility

JonasSteur:
https://github.com/evonove/django-oauth-toolkit/issues/516

https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-conversion.html

When using mysql with the utf8mb4 character set db index fields can only have a max length of 191 instead of 255.
This pr makes changes in models.py to ensure compatibility. I squashed and adapted migrations while leaving the old ones in place to ensure migrations still keep working in both directions for pre existing installs. New installs will only use the squashed migration.

I did have to add some checks on following settings attributes to be able to run the migrations though:
OAUTH2_PROVIDER_ACCESS_TOKEN_MODEL
OAUTH2_PROVIDER_APPLICATION_MODEL
This probably solves issue #240 as well


orf:
Should we really limit the length of the fields for all database backends due to a rather ridiculous MySQL specific issue?

JonasSteur:
It's ridiculous but it's either this or fix mysql. For now we had to fork this repo to be able to use it at our company with out current db config.

jleclanche:
Yeah I'm not particularly thrilled about merging that myself to be honest. Especially since this is highly backwards incompatible.

jleclanche:
Actually I glanced over the fact that this is why you squashed the migrations, sorry.

Hmm.. I'm really not sure. Given that all of the models in questions are abstract, have you considered replacing the model fields in question with your own with the updated max length? I know it's a bit silly to ask that of users but maintaining multiple migration paths is a massive pain, and it *is* just mysql...

orf:
You could make the length configurable via a setting, perhaps?

jleclanche:
I'd avoid that. Settings are highly mutable and this would be misleading without doing your own migration.

JonasSteur:
To look at it another way : Will there ever be a case we will need more that 191 characters for these tokens though, no matter what database is in use? And if that need really would arise why not let those users replace the fields ?

This library is just one of the many found with this migration problem after some of our servers got switched to utf8mb4. I'm quite sure we are not the only ones who will run into this. That's why I did this PR instead of just replacing those fields. Usually there's no good reason at all having those fields at max size anyways.

jleclanche:
I agree with you, but I'm reticent on the grounds that complex django migration paths are very hard to maintain in libraries (I'm dealing with having to wipe migrations in dj-stripe currently and it's not simple to push out), and this affects one particular configuration of mysql, a db engine which has a lot of issues like these.

Given that in all the time d-o-t has been out, we haven't actually had a report about it my guess is the amount of users affected is very low. And given that there is a solution to it that doesn't require forking (= subclasses), I *think* it's okay to reject this. Happy to take more counter-arguments, I could still be convinced.

jleclanche:
(Also happy to document this behaviour and the solution if anyone wants to file a doc PR)

JonasSteur:
Well, the way I did it everything should be backwards compatible. The only 'downside' would be that from now on everyone would get a bit smaller fields. There only would be a real problem if they actually used tokens bigger than 191 characters as those would get truncated. I seriously wonder if that case actually exists considering how ridiculously large those tokens would have to be...

We have some public libraries of our own that we modified like this without any problems. Any new migrations after this point wouldn't be any more complex, they would just continue from the squashed one.

jleclanche:
If we actually want to merge this I think I would rather make it a backwards-incompatible change, single-migration path and advise users who have tokens longer than 191 characters to implement the subclassing method.

JonasSteur:
That's possible indeed. Although I don't think doing it this way would be that problematic. It's the recommended way to handle these situations.

I wouldn't mind doing a PR with a single migration path though, it's a matter of deleting all the old ones and generate a fresh one. Would be most simple and clear for sure.

jleclanche:
No need to delete the old migrations, I just meant a PR that changes the field size from 255 to 191. That migration will loudly fail if you have longer tokens and it can't proceed, thus if you hit that you can subclass your stuff.

Now that I say it out loud that actually doesn't sound too bad to do actually.

JonasSteur:
The problem is that if you do it like this you won't be able to run the old migrations as they will loudly fail as soon as they try to set an indexed field to 255 length. That's how I stumbled upon this issue after all. So the only 2 options are doing it like this or delete and generate new migrations. Other way won't work unfortunately.

jleclanche:
Crap, you're right. Urgh.

orf:
In Mysql is there a way to specify the character collation on a single
column? Is there a need for utfmb8 on this column?

JonasSteur:
It's indeed possible to specify character collation on a single table or column in mysql. To make it work for a user running a mysql utf8mb4 db it would still mean changing the currently existing migrations: Creating the table -> altering it so it's not utf8mb4 by default anymore -> and only then adding the 255 char indexes. Not an improvement over the other 2 solutions imo...

Issue #525, opened 2017-10-31
Add Documentation on how to customize tokens and grant models.


Agilulfo:
In the Advanced section of the documentation there are no instructions on how to associate custom informations to tokens.

For what I understood this can be achieved customizing at least grant and token models.
Customizing these models should be simple as customizing the Application model..
However I'm having several issues with generating and applying migrations.

For instance If I subclass AbstractGrant and the I try to generate migrations without this into settings
`OAUTH2_PROVIDER_APPLICATION_MODEL = 'oauth2_provider.Application'`
then I get the following error:
```
AttributeError: 'Settings' object has no attribute 'OAUTH2_PROVIDER_APPLICATION_MODEL'
Makefile:115: recipe for target 'makemigrations' failed
make: *** [makemigrations] Error 1
```

If I define the OAUTH2_PROVIDER_APPLICATION_MODEL a migration will be generated.
Once I apply the new migration and I add 
`OAUTH2_PROVIDER_GRANT_MODEL = 'oauth.MyGrant'`
as soon as I regenerate the migrations I get 
```
django.db.migrations.exceptions.CircularDependencyError: oauth2_provider.0001_initial, oauth.0001_initial
Makefile:115: recipe for target 'makemigrations' failed
make: *** [makemigrations] Error 1
```
Having basic instruction on how to proceed would be really appreciated.

MattBlack85:
After further investigation it seems wrong to me that the initial migration has dependencies to:
 - oauth2_settings.APPLICATION_MODEL
 - oauth2_settings.ACCESS_TOKEN_MODEL
 - oauth2_settings.REFRESH_TOKEN_MODEL
 - oauth2_settings.GRANT_MODEL

This way whatever subclass of one of this models lead to a circular dependency since for example the `Grant` model has a FK to the `Application` model  so `Application` must be created first and the `Grant`, but since `Grant` depends on `Application` and `Application` depends on grant we are not able to use a custom model

MattBlack85:
@jleclanche can we have an input on this please? We need to implement this thing very soon and we don't know if we have to fork the repo and go with the workaround or dig more into what we're doing wrong 🙏

jleclanche:
@MattBlack85 Have you tried adding the dependencies to the migrations? If it works feel free to PR

MattBlack85:
@jleclanche those dependencies are already in the migrations, the only way it works for us is removing them

jleclanche:
[...]

PR #531, opened 2017-12-04
Remove dependencies to models that create circular errors

MattBlack85:
This should fix circular migrations error like those in #525 
The changes in this PR are supported by 2 things:
 - tried to remove manually the dependencies that were creating circular migratiosn dependency error which worked
 - tried to rebuild django-oauth-toolkit migrations from scratch obtaining this result
migrations

Now I'm not an expert of swappable models and I don't get them totally but this solutions seems to me the right one not only because django creates the migration with only one dependency to the `user` model but because If anybody tries to subclass one of the models, I don't see how the migrations for the subclassed model can work since:
 - sublcassed model migrations depend on `django-oauth-toolkit 0001`
 - `django-oauth-toolkit` migration 0001 depends on `subclassed 0001`

Happy to take this discussion further @jleclanche 
/CC @menecio @synasius pinging you too since I noticed you added those additional dependencies

### DIFF
--- a/oauth2_provider/migrations/0001_initial.py
+++ b/oauth2_provider/migrations/0001_initial.py
@@ -11,11 +11,7 @@ from django.conf import settings
 class Migration(migrations.Migration):
 
     dependencies = [
-        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
-        migrations.swappable_dependency(oauth2_settings.APPLICATION_MODEL),
-        migrations.swappable_dependency(oauth2_settings.ACCESS_TOKEN_MODEL),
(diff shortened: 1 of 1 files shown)


Praseetha-KR:
This solution fixed the circular dependency caused by extending the AccessToken model. Thanks @MattBlack85. Is there any ETA on this PR merge?

jleclanche:
I have no idea if this is correct, but then again I have no idea if the status quo is correct either and at least I'm getting reports that this fixes some issues... so merging.

menecio:
Sorry to join late to this thread, I saw the reports about circular dependencies but we use DOT and we never had this issue, so I can't say I have knowledge on the issue.

Issue #497, opened 2017-07-09
Suggestion for better handling of refresh tokens


jturmel:
So one problem any application implementing OAuth2 can have is issuing a request to refresh the access and refresh token but not actually receiving them and then forcing the user to re-authenticate. The most common case of this for us is native mobile applications that do a lot of background work and run into issues of network connectivity and or background process time limits (ie iOS gives you 30 seconds) and are unable to complete the full request/response life cycle before one of those cuts it short. The application then doesn't have the current refresh token from the backend and the only recourse is to have the user re-authenticate, which isn't ideal.

Currently the library offers the ability to completely disable rotation, so combined with either a much longer lifetime, or just disabling it altogether, you can work around the issue, but that too isn't ideal from a security standpoint.

I've looked into the issue some and ran into a company that was open in how they dealt with it and you can read about that here: https://community.fitbit.com/t5/Web-API-Development/Refresh-token-amp-network-timeout/td-p/1102761 -- I don't think their solution was necessarily the most ideal, but it's definitely better than nothing.

My proposal would be to have the following:
- Instead of the refresh token being revoked immediately, associate it with the new access token directly (ie `previous_refresh_token`)
- When access token is "used" for the first time, check to see if a `previous_refresh_token` is associated with it, if so then revoke it.
- As a fallback to the case when the new access token is not used for an extended period of time, configurable as a setting (ie 1d default), have `cleartokens` revoke it if it's past that configured time period. To be able to determine if this period has passed, an explicit `expires` field will need to be added on the RefreshToken model.

jturmel:
Another side benefit of this is that `cleartokens` doesn't currently clear out older access tokens that are no longer in use because if the associated refresh tokens are set to never expire it can't, this would allow that to happen and keep the tables cleaner.

PR #543, opened 2018-01-15
Address timeout on refresh with 'cached' responses.

phillbaker:
Thanks for all the work on this project! There wasn't a ton of discussion on #497, but I wanted to put together a proposed approach for feedback.

This handles the case where a refresh token is used to generate a new access token but the new access/refresh token is never received, allow a refresh token to be re-used within a grace period, returning the same access token (appearing as if the request/response was 'cached').

This now marks the RefreshToken revoked with a timestamp, instead of deleting it.

If this approach fits with the design direction of the project, we'd probably also want to extract the hard coded 2 minutes into a setting similar to `ACCESS_TOKEN_EXPIRE_SECONDS`.

Close https://github.com/evonove/django-oauth-toolkit/issues/497


jleclanche:
cc @jturmel

Sorry there was little discussion on this and for the time addressing this.
I'm a little iffy merging nonstandard behaviour like this into DOT. At first glance, the problem is legitimate and fitbit's proposed solution does sound good, but I wonder if there isn't some security issue lying in wait there.

My instinct tells me making this possible in DOT is a good idea, but making it default behaviour is going too far.
The code looks good, if you can gate it behind a setting and add documentation on the setting and the behaviour, I'll merge it.

phillbaker:
Thanks for taking a look. That all makes sense to me!

I'll add the setting and documentation over the next couple of days and
ping folks when it's ready for another review.

phillbaker:
`ACCESS_TOKEN_EXPIRE_SECONDS` is already a setting, but I added a `REFRESH_TOKEN_GRACE_PERIOD_SECONDS` with a default value of zero and documented it.

Happy to add more documentation elsewhere if it'd help. If you could take another look, would be much appreciated!

tests/test_oauth2_validators.py
```diff
@@ -236,7 +236,7 @@ def test_save_bearer_token__with_new_token_equal_to_existing_token__revokes_old_
 
         self.validator.save_bearer_token(token, self.request)
 
-        self.assertEqual(1, RefreshToken.objects.count())
+        self.assertEqual(2, RefreshToken.objects.count())
```
jleclanche: It's unclear to me why this value changed?

jleclanche:
Ok, looks good, one question in tests and needs a conflict fix before I can land, but otherwise gtm.

tests/test_oauth2_validators.py
phillbaker: Since the refresh token is no longer deleted on refresh (just marked revoked), the count increases here. We could also write this: `self.assertEqual(1, RefreshToken.objects.filter(revoked__isnull=True).count())`

jleclanche: Yeha, could you add that in the test? Then I'll land 👌

phillbaker: Updated. Thanks!

Issue #514, opened 2017-10-12
System ends up in deadlock state when trying to revoke refresh token 


Ayub-Khan:
Hi,
Version : django-oauth-toolkit==0.12.0
We have been using django-oauth-toolkit for quite a while and recently we have started facing those issue that calls to save_bearer_token ends up in a deadlock state when trying to revoke the refresh_token_instance.
We are still trying to figure out if this issue is related to our system or not. Just want to know if this is a known and occur because of the old version and was fixed in a newer version. 
Traceback
`File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/newrelic/hooks/framework_django.py", line 527, in wrapper
    return wrapped(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/edx-platform/openedx/core/djangoapps/oauth_dispatch/views.py", line 94, in dispatch
    response = super(AccessTokenView, self).dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/edx-platform/openedx/core/djangoapps/oauth_dispatch/views.py", line 53, in dispatch
    return view(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
[...]

PR #544, opened 2018-01-17
Delete Refresh token before Access Token to avoid cascade deadlock

NiekHoekstra:
Fix for difficult to reproduce issue of #514. 
It is a concurrency issue which I fail to reproduce in a docker/controlled environment.
This change was done for Python 3.4
In absence of reproduction capabilities we deployed this change to our production servers.
The issue hasn't come up since.

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -335,8 +335,10 @@ class AbstractRefreshToken(models.Model):
         Delete this refresh token along with related access token
         """
         access_token_model = get_access_token_model()
-        access_token_model.objects.get(id=self.access_token.id).revoke()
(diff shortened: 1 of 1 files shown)


jleclanche:
Thank you!

bramski:
I can't find this bugfix in the changelog to figure out what version to upgrade to as I'm experiencing this issue.  What release is this fix in?

bramski:
Hello?

tremby:
@bramski looks like the first tag with this patch is 1.1.0: https://github.com/jazzband/django-oauth-toolkit/compare/1.0.0...1.1.0

PR #551, opened 2018-02-03
Expire AccessTokens even when REFRESH_TOKEN_EXPIRE_SECONDS is not set

ashb:
If refresh tokens are not being used or if REFRESH_TOKEN_EXPIRE_SECONDS is not set in the settings then access tokens are never cleaned up by the `cleartokens` command.

This used to be the behaviour, but it looks like there was an un-intended indentation change in https://github.com/evonove/django-oauth-toolkit/commit/be8bcccc0adce86862c9d35ae50afc42a3a28d95


jleclanche:
Thanks!

PR #557, opened 2018-02-16
Introspect with basic auth

n2ygk:
Many OAuth 2.0 servers use HTTP Basic Authentication for the RFC 7662 introspection endpoint rather than a Bearer token. This PR adds optional support for using Basic credentials.


oauth2_provider/oauth2_validators.py
```diff
@@ -240,13 +240,25 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+                                              introspection_credentials):
+        # some rfc7662 implementations use a Bearer token while others use Basic Auth
+        headers = None
+        if introspection_token:
+            headers = {"Authorization": "Bearer {}".format(introspection_token)}
+        elif introspection_credentials and len(introspection_credentials) == 2:
+            basic = base64.b64encode(introspection_credentials[0].encode("utf-8")
+                                     + b":"
+                                     + introspection_credentials[1].encode("utf-8"))
+            headers = {"Authorization": "Basic {}".format(str(basic, "utf-8"))}
```
jleclanche: In general you want `s.decode("utf-8")` rather than `str(s, "utf-8")`

oauth2_provider/oauth2_validators.py
```diff
@@ -240,13 +240,25 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
     def get_default_redirect_uri(self, client_id, request, *args, **kwargs):
         return request.client.default_redirect_uri
 
-    def _get_token_from_authentication_server(self, token, introspection_url, introspection_token):
-        bearer = "Bearer {}".format(introspection_token)
+    def _get_token_from_authentication_server(self,
+                                              token,
+                                              introspection_url,
+                                              introspection_token,
+                                              introspection_credentials):
```
jleclanche: minor: please use indented style instead of hanging brackets:

```py
def _get_token_from_authentication_server(
    self, token, introspection_url, introspection_token, introspection_credentials
):
    ...
```

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = django-oauth-toolkit
-version = 1.0.0
+version = 1.1.0
```
jleclanche: Please do not change the version, that is something that happens separately.

oauth2_provider/oauth2_validators.py
```diff
@@ -240,13 +240,25 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+        # some rfc7662 implementations use a Bearer token while others use Basic Auth
+        headers = None
+        if introspection_token:
+            headers = {"Authorization": "Bearer {}".format(introspection_token)}
+        elif introspection_credentials and len(introspection_credentials) == 2:
```
jleclanche: What's with the `len(introspection_credentials) == 2`?

oauth2_provider/oauth2_validators.py
```diff
@@ -240,13 +240,25 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+            basic = base64.b64encode(introspection_credentials[0].encode("utf-8")
+                                     + b":"
+                                     + introspection_credentials[1].encode("utf-8"))
```
jleclanche: This is a bit confusing, could you rewrite this like that for clarity:

```py
username = introspection_credentials[0].encode("utf-8")
password = introspection_credentials[1].encode("utf-8"))
authorization = "Basic {}:{}".format(username, password)
```

oauth2_provider/oauth2_validators.py
n2ygk: introspection_credentials is expected to be a tuple or list of (client_id, client_secret). Just making sure the user provided only 2 items. I guess I should do a type check too since they could provide a string of length 2.... Or expect a dict keyed by `client_id` and `client_secret`.  What do you think?

jleclanche: Not sure it's worth checking the format validity. It'll break if users pass the wrong format and that's Python for you. In 3.6 we'll add typings.

n2ygk:
@jleclanche LMK if my changes are OK.

oauth2_provider/oauth2_validators.py
```diff
@@ -240,20 +240,34 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
...
+        """
         headers = None
         if introspection_token:
             headers = {"Authorization": "Bearer {}".format(introspection_token)}
-        elif introspection_credentials and len(introspection_credentials) == 2:
-            basic = base64.b64encode(introspection_credentials[0].encode("utf-8")
-                                     + b":"
-                                     + introspection_credentials[1].encode("utf-8"))
-            headers = {"Authorization": "Basic {}".format(str(basic, "utf-8"))}
+        elif (introspection_credentials and
```
jleclanche: Just drop the safety checks as I mentioned before. `elif introspection_credentials:` is fine.

jleclanche:
The one change and gtg.

PR #563, opened 2018-03-02
add TokenHasMethodScopeAlternative

n2ygk:
These add a couple more permission classes for scope matching.  `TokenHasMethodScope` allows specifying required scope(s) on a per-method basis (in the case where READ_SCOPE and WRITE_SCOPE are not granular enough) and `TokenHasMethodPathScope` adds regex path matching and alternative lists of required scopes for a given regex match which is aligned with OpenAPI Spec (OAS) [security requirement object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#securityRequirementObject) list of alternative matching:

> "When a list of Security Requirement Objects is defined on the Open API object or Operation Object, only one of Security Requirement Objects in the list needs to be satisfied to authorize the request."


n2ygk:
This is embarrassing. 

FYI, when I run tox locally it's throwing this flake8 error:
```
ERROR:   flake8: Error creating virtualenv. Note that some special characters (e.g. ':' and unicode symbols) in paths are not supported by virtualenv. Error details: InvocationError('/Users/alan/src/django-oauth-toolkit/env/bin/python3 -m virtualenv --python /Users/alan/src/django-oauth-toolkit/env/bin/python3 flake8 (see /Users/alan/src/django-oauth-toolkit/.tox/flake8/log/flake8-0.log)', 100)
```
but when I run flake8 manually it seems to work OK. No excuse for forgetting to run flake8 manually.

jleclanche:
Mmh, looks good. Second opinions?

n2ygk:
@jleclanche I'm thinking about adding another permission class named something like `IsRegisteredClient` which calls `Oauth2Validator.validate_client_id`.

In the case of using the built-in oauth2 provider this validation is implicit. For an external oauth2 provider, this would confirm that the client app is "trusted" by this resource server which involves registering the client application.

My example use case is server-to-server trust using Client Credentials grant. This is analogous to the non-OAuth way of using Basic Auth in which each server has a registry of clients it trusts. It probably applies to Authorization Code grants as well.... Just because a client is registered with AS doesn't mean it should be permitted to talk to every RS that can introspect from the same AS.

Does this make sense?

n2ygk:
@jleclanche @symptog pinging you for thoughts on [proposed](https://github.com/evonove/django-oauth-toolkit/pull/563#issuecomment-370859623) IsRegisteredClient.

jleclanche:
I've never encountered this use case myself so it's hard for me to make a judgement on it. I'd really like to be able to get opinions from other folks but I'm not sure anyone's listening.

n2ygk:
@jleclanche on further consideration, I think the best way to do this is to use a resource server-specific scope that only "approved" clients can use. Similar to how "introspection" scope is currently used in oauth2_provider. So ignore my ramblings;-)

n2ygk:
Regarding `TokenHasMethodPathScope` do you think that makes sense or would it make more sense to build this into a url router, using `TokenHasMethodScope`? A URL router that magically adds the RelationshipView would be a cool thing....

symptog:
> @symptog I'm thinking of changing oauth2_validators._get_token_from_authentication_server to change application=None to do a lookup of the application by client_id. Is that reasonable?

My thoughts were that the AS knows about the applications, clients and scopes and the RS only validates the token towards the AS and so the RS don't need to know anything else except if the token is valid for the requested resource.

Towards your commit:
The `TokenHasMethodPathScope` doesn't open to me. As you usually map URLs to view functions/classes, adding required scopes to those functions/methods is more intuitive and more clear.

n2ygk:
@jleclanche @symptog I've refactored to what I hope makes more sense.

n2ygk:
@jleclanche just a reminder that this looks like it's ready to merge. @symptog's thumbs-up sounds like a second opinion;-)

jleclanche:
Sorry for the delay. This will land after 1.1.0. Could you rebase against master please?

n2ygk:
@jleclanche Sorry for the multiple attempts, but ready for merge now, including some test coverage improvements to pre-existing code, leaving the campground cleaner than I found it;-)

n2ygk:
@jleclanche @symptog OK to merge?

jleclanche:
Can you fixup&rebase it please? I've been pretty busy but I'll try to do a final review asap.

n2ygk:
Already did rebase it (merged master into my branch as there were too many
conflicts to do a clean rebase).

jleclanche:
That's messing with things. I can't merge something with conflicts either way; if it's a problem I recommend cherry-picking and re-applying the conflicting parts manually.

n2ygk:
I wasn’t aware there were still conflicts. Maybe some more recent commits?
I’ll take another look but thought it was clean and mergeable.

n2ygk:
@jleclanche I'm looking at a screen that shows three green check-marks:
[x] Review requested
[x] All checks have passed
[x] This branch has no conflicts with the base branch
and, I assume since I've recently joined Jazzband, "Merge pull request" is available for me.

Where are you seeing conflicts?

jleclanche:
Sorry again for the delays. I've merged this now. I'm not super happy with the naming but we can change that before 1.2.0 release if needs be.

n2ygk:
@jleclanche Thanks! Agree about the name. It's too long. Maybe drop "Alternative" from the end? Use OAS nomenclature and call it `TokenMeetsSecurityRequirements` or `TokenMeetsOpenApiRequirements`? Ugh.

jleclanche:
`TokenMatchesOASRequirements` or something similar sounds good.

PR #564, opened 2018-03-02
check response status_code from authentication server

manelclos:


oauth2_provider/oauth2_validators.py
```diff
@@ -277,6 +277,11 @@ def _get_token_from_authentication_server(
             log.exception("Introspection: Failed POST to %r in token lookup", introspection_url)
             return None
 
+        if response.status_code == 403:
```
jleclanche: You should check against the expected status code rather.

oauth2_provider/oauth2_validators.py
```diff
@@ -277,6 +277,11 @@ def _get_token_from_authentication_server(
...
+            log.exception("Introspection: Failed to authenticate to %r",
```
jleclanche: Use log.error(). You're not in an exception.

oauth2_provider/oauth2_validators.py
```diff
@@ -277,6 +277,11 @@ def _get_token_from_authentication_server(
...
+                          introspection_url)
```
jleclanche: Please don't indent like that :) And a single-line is enough here.

oauth2_provider/oauth2_validators.py
manelclos: ok, but then we can't say that we failed to authenticate or permissions are wrong (401, 403), "if not status_code == 200", is that ok?

oauth2_provider/oauth2_validators.py
manelclos: sure

oauth2_provider/oauth2_validators.py
jleclanche: If you want to handle specific status codes that is fine. You should have an else clause in there though if you do.

jleclanche: BTW a couple of tests would be nice for this.

manelclos:
@jleclanche, new commit ready for review. I'll squash when the review is finished.

PR #570, opened 2018-03-10
Uses user_passes_test and is_superuser (instead of settings.DEBUG) to…

chandler37:
… protect internal endpoints.

https://github.com/evonove/django-oauth-toolkit/issues/568

### DIFF
--- a/docs/tutorial/tutorial_02.rst
+++ b/docs/tutorial/tutorial_02.rst
@@ -35,9 +35,12 @@ URL this view will respond to:
 .. code-block:: python
 
     from django.conf.urls import url, include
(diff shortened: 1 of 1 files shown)


jleclanche:
These are the docs, I'd rather not make them more complicated. the `if settings.DEBUG` block is easier to understand imo.

chandler37:
https://docs.djangoproject.com/en/2.0/howto/deployment/checklist/ says "You must never enable debug in production."

Are you open to changing the code to require an active superuser on all but the three public endpoints? That would make what you do in tutorial_01 (this PR is about tutorial_02) secure by default as well.

PR #573, opened 2018-03-15
Add settings to enable/disable management views

allisson:


jleclanche:
The URLs are already split up. A better way to do this that does not involve settings is to pull in `base_urlpatterns` standalone.

That said, I would rather drop `management_urlpatterns` altogether from `urlpatterns` and issue changelog notes on how to recover it (by puilling in `management_urlpatterns` from an include).

PR #575, opened 2018-03-20
Updates to get example code to work with latest Django and DRF

philratcliffe:
Modified the example code so it now works Django 2.0.3 and Django REST Framework 3.7.7


jleclanche:
Thanks!

Issue #453, opened 2017-02-17
403 Forbidden error code is returned instead of 401 Unauthorized in OAuthLibMixin


veranika-sab:
According to [RFC6750 3.1 Error codes](https://tools.ietf.org/html/rfc6750#section-3.1) there are different response codes for **invalid access** token and for valid access token, but **insufficient scope**:

>    invalid_token
         The access token provided is expired, revoked, malformed, or
         invalid for other reasons.  The resource SHOULD respond with
         the HTTP 401 (Unauthorized) status code.  The client MAY
         request a new access token and retry the protected resource
         request.

> insufficient_scope
         The request requires higher privileges than provided by the
         access token.  The resource server SHOULD respond with the HTTP
         403 (Forbidden) status code and MAY include the "scope"
         attribute with the scope necessary to access the protected
         resource.

As for now there is no possibility to differentiate whether it is an invalid access token or insufficient scope. It is validated in one method: [AccessToken.is_valid()](https://github.com/evonove/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L200-L206).

Are there any reasons of not-following the RFC?

jleclanche:
This is #367 isnt it?

I'll take a PR that fixes it :)

veranika-sab:
@jleclanche I'm not sure. As I understand #367 it is about differentiating the requests with invalid token and without token. The case I've described is about differentiating the requests with invalid token and invalid scope (while token is valid).

kevin-brown:
> Are there any reasons of not-following the RFC?

I'll note that it says _SHOULD_ instead of _MUST_, so technically it's following the RFC.

You may find that this is actually because of how DRF handles returning a 401 or a 403 for authentication and authorization errors. This is something outside of Django OAuth Toolkit's control, because it is based on how you defined your settings.

http://www.django-rest-framework.org/api-guide/authentication/#unauthorized-and-forbidden-responses

Pear0:
I just ran into this issue while using `IsAuthenticatedOrTokenHasScope` where all requests were returning 403 Forbidden. 

I found out through tracing that DRF only uses the first authentication class when determining whether to coerce 401 responses to 403 responses, and I had specified my `SessionAuthentication` class before my `OAuth2Authentication` class. 

Turns out I could have saved myself quite a bit of time by just reading the docs so for anyone having this issue, here's the relevant part from the link kevin-brown posted:

[...]

PR #584, opened 2018-04-12
Feature/return error code when token is not valid

christopherdcunha:
This PR tries to implement the recommendations in https://tools.ietf.org/html/rfc6750#section-3. 
Namely, these changes will add the `error` and `error_description` attributes to the `WWW-Authenticate` header.

---

I think this relates to the comments in https://github.com/jazzband/django-oauth-toolkit/issues/453

CC: @veranika-sab  @emoragaf


christopherdcunha:
Thanks for having a look so fast @jleclanche.

Do you think this will be able to make the next release (presumably 1.1.0)?

jleclanche:
Yeah sure.

PR #586, opened 2018-04-30
Return state with authorization denied error

whytheplatypus:
Conforms to [rfc6749 section 4.1.2.1](https://tools.ietf.org/html/rfc6749#section-4.1.2.1)
> REQUIRED if a "state" parameter was present in the client
>         authorization request.  The exact value received from the
>         client.


jleclanche:
Thanks for the excellent PR.

PR #597, opened 2018-05-12
Rename TokenHasMethodScopeAlternative to TokenMatchesOASRequirements

n2ygk:
See suggestion for a better name in #563


n2ygk:
@jleclanche OK if I go ahead and merge this name change?

Issue #583, opened 2018-04-12
Migrations Broken With django-pyodbc-driver


beh9540:
When running migrations on a new project using [django-pyodbc-driver](https://github.com/michiya/django-pyodbc-azure), migrations fail due to the changing of the primary keys from AutoField to BigAutoField in [this migration](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/migrations/0005_auto_20170514_1141.py) because of the limited capability of SQL Server to edit an IDENTITY property.

I'm able to work around this by manually changing the migrations (obviously not preferred) but I'm curious if you'd be open to collapsing this in the future? I'd be happy to work on putting a PR together.

synasius:
Totally agree with @jleclanche 

@beh9540 you might consider using `MIGRATION_MODULES` setting to override migrations for the `oauth2_provider` app. This way you can maintain your own migrations for an external package.

See https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-MIGRATION_MODULES

beh9540:
@synasius yeah that's what I'm doing now, but I figured it'd be worth bringing up for anyone else who might run into this in the future.

n2ygk:
@jleclanche @synasius We are also seeing this problem with pyodbc. 

What's the reluctance about doing a migration reset? It seems like the cleanest fix.  The current clutter of automigrations is messy and I don't think there's any value in maintaining history of old database structure; other than in git. 

I don't see how this would cause problems for anybody, would make the package cleaner, and would solve the pyodbc problem.

Anyone who has had the package for a while has already migrated their models and new users don't need that history.  Am I missing something?

```
oauth2_provider/migrations/0001_initial.py
oauth2_provider/migrations/0002_08_updates.py
oauth2_provider/migrations/0003_auto_20160316_1503.py
oauth2_provider/migrations/0004_auto_20160525_1623.py
oauth2_provider/migrations/0005_auto_20170514_1141.py
oauth2_provider/migrations/0006_auto_20171214_2232.py
oauth2_provider/migrations/__init__.py
```

n2ygk:
[...]

PR #615, opened 2018-06-28
WIP: Workaround for MS SQL Server/pyodbc

n2ygk:
Workaround for this error (see #583): 
`pyodbc NotImplementedError: the backend doesn't support altering from/to AutoField.`
which is caused by the inability to AlterField from `AutoField` to `BigAutoField`

This alters the migrations (0001 & 0005) only when the default database engine is pyodbc.

I've tested with MS SQL Server and it seems to do the right thing. This may break migrations for anyone using this toolkit with MS SQL Server prior to migration 0005 of 2017-05-14 11:41 (commit c1cd5d7b575f12313f664036e52383ba5cabf0ec)

@jleclanche Please take a look as I'm not really sure this is the best approach, but hopefully it's one that will let projects to continue to use MS SQL Server without having to maintain separate migrations.

Also, I'm not sure checking `settings.DATABASES['default']['ENGINE']` is the right way to identify the database that the migration is using.


jleclanche:
Shouldn't this be a bug filed against django or something?

Maybe we can reset the migrations at some point (like I did in dj-stripe) to prevent this from being an issue in the future, but for the time being, as all the migration does is alter the field size, maybe we can just recommend to --skip it for the very few mssql users out there.

Furthermore, the next version of Django will include migrations from Auto to BigAuto in the core itself so I suspect we will see a fix for this if there isn't one already.

I think I would rather reset the migrations than let this be an issue.

n2ygk:
It's apparently a [limitation of mssql](https://github.com/michiya/django-pyodbc-azure/blob/549026d0f6b031f69ca27f0a488cfe2266677f10/sql_server/pyodbc/schema.py#L129-L133) so there will be breaking migrations for all django projects in the world that use mssql (not so many;-).

I agree that resetting the migrations would be way cleaner but wouldn't that also be a breaking change for anyone with existing models who has not yet updated to a newer release of DOT? I'm OK with that -- just something to document in release notes.

I've opened https://github.com/michiya/django-pyodbc-azure/issues/172  and referenced https://github.com/django/django/pull/8924

jleclanche:
I've done migration reset releases twice in the past now. The pattern is:

- Release a major version with all the migrations
- Immediately after, release a new major version which resets all the migrations (and does nothing else).

n2ygk:
Closing this PR as it only works around a single edge case (deficiency in MS SQL) that can be handled outside this project.

PR #620, opened 2018-07-10
Fix 500 error for refresh with revoked access token.

### LINKED ISSUE #585: Refreshing a revoked access token throws an error 500
When you revoke an access token and try to refresh it afterwards, you end up with an error 500.

This seems to be related to the `get_original_scopes` method of `oauth2_provider/oauth2_validators.py` which tries to retrieve the access token related to the refresh token. However, since it was revoked, it no longer exists and Django throws a `DoesNotExists` error which isn't catch.

When we revoke an access token, maybe we should revoke refresh tokens too ? Or we could add a clause to the queryset on `validate_refresh_token` method to exclude refresh token without `access_token`.

robrap:
fixes #585

Note that there are no integration tests, so the unit tests don't
actually show the 500 error that would have been seen with a call
to oauth2_provider/oauth2_backends.py:create_token_response.

Also, there was no coverage for validate_refresh_token.  I added some coverage since I was updating it, but I did not add full coverage.

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -625,6 +625,14 @@ class OAuth2Validator(RequestValidator):
         if not rt:
             return False
 
+        try:
+            # ensure access token was not revoked and later calls to get_original_scopes
+            # will not raise AccessToken.DoesNotExist.
+            if not rt.access_token_id:
+                AccessToken.objects.get(source_refresh_token_id=rt.id)
+        except AccessToken.DoesNotExist:
+            return False
+
         request.user = rt.user
         request.refresh_token = rt.token
         # Temporary store RefreshToken instance to be reused by get_original_scopes and save_bearer_token.
--- a/tests/test_oauth2_validators.py
+++ b/tests/test_oauth2_validators.py
@@ -270,6 +270,43 @@ class TestOAuth2Validator(TransactionTestCase):
         self.assertEqual(0, RefreshToken.objects.count())
         self.assertEqual(1, AccessToken.objects.count())
 
+    def test_validate_refresh_token__with_valid_access_token__returns_true(self):
+        access_token = AccessToken.objects.create(
+            token="123",
+            user=self.user,
+            expires=timezone.now() + datetime.timedelta(seconds=60),
+            application=self.application
+        )
+        refresh_token = RefreshToken.objects.create(
+            access_token=access_token,
+            token="abc",
+            user=self.user,
+            application=self.application
+        )
+        is_refresh_token_valid = self.validator.validate_refresh_token(
+            refresh_token=refresh_token, client=self.request.client, request=self.request
+        )
+        self.assertTrue(is_refresh_token_valid)
+
+    def test_validate_refresh_token__with_revoked_access_token__returns_false(self):
+        access_token = AccessToken.objects.create(
+            token="123",
+            user=self.user,
+            expires=timezone.now() + datetime.timedelta(seconds=60),
+            application=self.application
+        )
+        refresh_token = RefreshToken.objects.create(
+            access_token=access_token,
+            token="abc",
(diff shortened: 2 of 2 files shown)


robrap:
UPDATE: This did in fact fix my issue, and now returns a 4xx instead of a 5xx error.

robrap:
Why did coveralls report?:
0 of 0 changed or added relevant lines in 0 files are covered.

jleclanche:
This doesn't look right... at first glance it should probably be happening in the same transaction as what's happening before.

robrap:
@jleclanche: Thanks for looking at this.  I'm not sure what "...the same transaction as what's happening before" refers to?

Here are a lot more details about the problem.

Our original code used django-auth-toolkit 0.12.0. In that version, if you do the following:
1. get a refresh token with an access token.
2. revoke the access token.
3. call `create_token_response` with this refresh token.  
We would get a 401.

When I upgraded to django-auth-toolkit 1.1.2, the behavior had changed, due to an issue I will describe in more detail.

For reference for my description below, here are the relevant lines in `oauthlib` in `validate_token_request`:
https://github.com/oauthlib/oauthlib/blob/v2.1.0/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py#L109-L116

**The problem:**
In `oauthlib: validate_token_request`, the refresh token is currently being returned as valid by `request_validator.validate_refresh_token`, but when it moves on to `request_validator.get_original_scopes`, it can't find the access token because it has been revoked and returns an error, that in turn becomes a 500 error.

What is probably warranted is an integration test that shows the whole flow between `create_token_response`, through `oauthlib`, calling back to the relevant `oauth2_provider` code.

Here are additional details of the error I am seeing...

**Error Message:**
DoesNotExist: AccessToken matching query does not exist.

**Relevant stacktrace:**
```
...
../../edx-venv/local/lib/python2.7/site-packages/oauth2_provider/views/base.py:206: in post
../../edx-venv/local/lib/python2.7/site-packages/oauth2_provider/views/mixins.py:125: in create_token_response
    return core.create_token_response(request)
../../edx-venv/local/lib/python2.7/site-packages/oauth2_provider/oauth2_backends.py:140: in create_token_response
    headers, extra_credentials)
../../edx-venv/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/endpoints/base.py:64: in wrapper
    return f(endpoint, uri, *args, **kwargs)
../../edx-venv/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/endpoints/token.py:117: in create_token_response
    request, self.default_token_type)
../../edx-venv/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py:59: in create_token_response
    self.validate_token_request(request)
../../edx-venv/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py:117: in validate_token_request
    request.refresh_token, request))
../../edx-venv/local/lib/python2.7/site-packages/oauth2_provider/oauth2_validators.py:608: in get_original_scopes
    return AccessToken.objects.get(source_refresh_token_id=rt.id).scope
```

robrap:
@jleclanche: I also did not find any indication that the refresh token should be invalid if the access token is revoked in the RFC.  So maybe the refresh token should continue to work as long as it is not revoked?  If so, maybe this was a bug in 0.12.0, and we want to correct the behavior?

In that case, would the solution be to have the refresh token include the original scopes in its model so it doesn't need to refer back to the (possibly revoked) access token?  Other?

robrap:
At this time, I think that both the old version (0.12.0) and the new version (1.1.2) have different versions of the same bug.  The old version returned a 401.  The new version returns a 500.  In both cases, this should probably be a 200.  I'm going to close this PR, which was just replicating an earlier bug, rather than fixing the issue.

Issue #621, opened 2018-07-11
ValueError: invalid literal for int() with base 10: 'abcdefg' (/o/applications/{pk}/)


ptrojanowski:
Is there a reason to have this regex in urls.py `[\w-]` instead of  `[0-9^/]` (for `pk`)?

Visit  for example: http://example.com/o/applications/1asdgsdg/, you will get this error:

> ValueError: invalid literal for int() with base 10: '1asdgsdg'

Take a look at this pull request:

MattBlack85:
closing with the same reason of the closed PR

PR #622, opened 2018-07-11
fixed regex in urls.py (pk)

ptrojanowski:
https://github.com/jazzband/django-oauth-toolkit/issues/621


jleclanche:
This is not the correct fix. Users can override the integer ID with a non-integer ID.

Issue #625, opened 2018-07-23
Cannot revoke OAuth2 RefreshToken without related Token


rooftopcellist:
##### SUMMARY
Django-OAuth-Toolkit supports deleting a token, but does not cascade on delete for it's RefreshToken anymore as of DOT 1.1.2.  This creates an issue when you try to revoke a RefreshToken that no longer has an associated AccessToken.  This will cause a traceback at [this line](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L380).  

> This could also potentially become an issue in reverse [here](https://github.com/jazzband/django-oauth-toolkit/blob/1.1.2/oauth2_provider/models.py#L365), but is unlikely to be hit unless a user is manually deleting RefreshTokens.  

Currently, users have no way to delete RefreshToken's as of DOT 1.1.2 given these circumstances, which is an issue because RefreshTokens can be used to issue a new AccessToken.  

##### ENVIRONMENT
* Django OAuth Toolkit version 1.1.2
* Python version 2.7.5

##### STEPS TO REPRODUCE

1. Create App
2. Create AccessToken
3. Revoke AccessToken using either method (`api/o/revoke_token`)
4. Observe that the RefreshToken remains and the AccessToken is deleted. 
^^ This all is as expected
5. Revoke RefreshToken (we currently can't do this)

```
awx.main.models.oauth.DoesNotExist: OAuth2AccessToken matching query does not exist.
```

Related Issue from AWX: https://github.com/ansible/awx/pull/2044

PR #630, opened 2018-08-06
fix a bug in refresh token revocation

ryanpetrello:
fixes: https://github.com/jazzband/django-oauth-toolkit/issues/625


jleclanche:
Landed in #653.

PR #637, opened 2018-08-27
Fix issue for: "DoesNotExist: AccessToken matching query does not exist."

zahlio:
We had an issue where the AccessToken would be updated in db, but due to our master-slave(s) setup this would result in a DoesNotExist error. This fix resolves the issue.


Geekfish:
We've noticed the same problem, it doesn't seem to happen all the time but it's not negligible.
Would it be possible to include this in the next release?

jleclanche:
I believe this is fixed in 5b51da74019046ef4c8c81c9975db029a2113d52

Issue #638, opened 2018-08-27
Multiple concurrency issues


gbataille:
Hi all,

Issue
====

So for 2 years that I have been using django-oauth-toolkit, I have been plagued with concurrency issues of the form
```
DoesNotExist: AccessToken matching query does not exist
```
originally from 
```
oauth2_provider/oauth2_validators.py in get_original_scopes at line 427
```
but after moving to version 1.1.2, it's
```
oauth2_provider/oauth2_validators.py in get_original_scopes at line 608
```

The thing is that (for me), 1.1.2 has also brought a new kind of issues
```
IntegrityError: duplicate key value violates unique constraint "oauth2_provider_accesstoken_source_refresh_token_id_key"
DETAIL:  Key (source_refresh_token_id)=(80187) already exists.
```
from
```
oauth2_provider/oauth2_validators.py in save_bearer_token at line 534
--> oauth2_provider/oauth2_validators.py in _create_access_token at line 566
```

I finally got to really look into them.

Problem
======

For me, those are concurrency issues that are due to the fact that we get the refresh_token from the db early in the process and that we try to use it later (to create a new access token for example).

I have drawn the process in a small diagram (source [here](https://gist.github.com/gbataille/d430a9da3ec351d9d4b4a0c95b02151a))

With the problem being in between the 2 red steps.
We must not have 2 processes going to this rightmost branch with the same refresh token.

![refresh_token_logic](https://user-images.githubusercontent.com/1113065/44662592-8e644480-aa0e-11e8-9090-e47d70853ada.png)

To Reproduce
==========

You can try and force some concurrency like so.
https://gist.github.com/gbataille/099a9eb4989c4277e9ce3312c8014391

Using it locally, I can make it fail in one of the 2 cases above quite often.

Fix
===

I think I have a fix. At least the logic seems sound, and I can't reproduce the issue anymore after applying it.

Submitting the PR

The key is to add the green step
![refresh_token_logic_fixed](https://user-images.githubusercontent.com/1113065/44663005-97a1e100-aa0f-11e8-8a8c-4b350a808173.png)

gbataille:
Ok, so I think the (second) problem is different now. Might not even be a concurrency issue.

I have `RefreshToken` in my db that have no `access_token` linked. That tends to mean (to my understading) that those `RefreshToken` have been used. Yet
* they are not marked as expired (no revoked date)
* there is no `AccessToken` with `source_refresh_token` pointing to the `RefreshToken`

[...]

PR #639, opened 2018-08-27
fix(#638): concurrency issue on new token from refresh token

gbataille:
Seems to fix the many issues I have in production

The green step does the trick
refresh_token_logic_fixed

Code to generate the diagram [here](https://gist.github.com/gbataille/28573bf1ac4aca95e53351aa0056ff0a)

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -511,6 +511,12 @@ class OAuth2Validator(RequestValidator):
             else:
                 # revoke existing tokens if possible to allow reuse of grant
                 if isinstance(refresh_token_instance, RefreshToken):
+                    # First, to ensure we don't have concurrency issues, we refresh the refresth token
+                    # from the db while acquiring a lock on it
+                    refresh_token_instance = RefreshToken.objects.select_for_update().get(
+                        id=refresh_token_instance.id
+                    )
+
                     previous_access_token = AccessToken.objects.filter(
                         source_refresh_token=refresh_token_instance
                     ).first()
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -511,11 +511,13 @@ class OAuth2Validator(RequestValidator):
             else:
                 # revoke existing tokens if possible to allow reuse of grant
                 if isinstance(refresh_token_instance, RefreshToken):
-                    # First, to ensure we don't have concurrency issues, we refresh the refresth token
+                    # First, to ensure we don't have concurrency issues, we refresh the refresh token
                     # from the db while acquiring a lock on it
+                    # We also put it in the "request cache"
                     refresh_token_instance = RefreshToken.objects.select_for_update().get(
                         id=refresh_token_instance.id
                     )
+                    request.refresh_token_instance = refresh_token_instance
 
                     previous_access_token = AccessToken.objects.filter(
                         source_refresh_token=refresh_token_instance


MattBlack85:
can you write a test (with threads probably to simulate concurrency) that explains exactly where is the issue? I don't get very well why locking the row for a read guards against concurrency problems.

gbataille:
writing test for concurrency is flaky at best.

What don't you understand?
after retrieving the refresh token, we are deriving things from it and/or
modifying it.
In a concurrent context, several process take decision/modify the
underlying database object without refreshing the RefreshToken instance in
memory. The lock is to prevent 2 concurrent process acting on the same
RefreshToken to enter together this part of the code

jleclanche:
@gbataille Why the token lock instead of a db transaction?

gbataille:
there is already a db transaction.
But a db transaction does not do anything if you don't take a lock. the
refreshtoken has already been put in memory. anything you do in django does
not care about the fact that the db is in a transaction. you are still not
operating on the "latest" data in a non-concurrent way

jleclanche:
Yeah after reading more of the context I see what you mean. This is gtm and I don't think writing threaded tests for this is a particularly good idea either tbh.

MattBlack85:
not objecting about tests, I made them when the case was really clear been successful with 0 problems.
select_for_update does not prevent other processes to access the resource, it doesn't allow other actors to write at the same time the same row afaik hence I don't get why you use the lock to get the actual value of the refresh token from the db

gbataille:
select_for_update does a db lock while getting the record. If someone else
has a db lock on this object, it'll block until the lock is released.
this works because you are in a db transaction already.

In my graph, that means that everything below the green box where I take
the lock is sure to execute in a single thread fashion, which is what we
need since we take decisions/mutate state based on the refresh token record

MattBlack85:
I used select_for_update often and I never noticed it blocking when fetching items from the DB, it does when you want to write the row which is locked tho. I also made a quick test where I wrote a small function acquiring a lock and sleeping. The 2nd actor is indeed able to fetch the same row  while the first hold the lock from the DB but it's not able to save it until the first actor relases the lock

gbataille:
From the doc:

select_for_update(*nowait=False*, *skip_locked=False*, *of=()*)

Returns a queryset that will lock rows until the end of the transaction,
generating a SELECT ... FOR UPDATE SQL statement on supported databases.


But then it depends on your db and your db config (typically isolation
level of reads).

I have solved a lot of concurrency issues in my production with this trick.
That's what it's meant for. I use to do a dummy update of the record to
make sure it was locked, but select_for_update does just that.

I can guarantee that this strategy works (in a regular postgres setup at
least) in other cases. For this particular issue I investigated, I just put
my fork of the lib in production today, so I'll have more details in a
couple of days (the concurrency issue is reported about 10 times a day in
my production before fix)

gbataille:
go damn man, I was just building up an example. I was about to publish some
lock traces :D you should have continued to disagree :D

Are you using sqlite locally? that might behave quite differently.

For "risk management" purpose, I did put this code in my production today,
which receives 100s of refresh token requests per days. So if you wait a
couple of days I can tell you
- whether the concurrency issue has disappeared (at least in my context)
- whether I saw some other undesirable effect come up

How does that sound?

MattBlack85:
:D I am fully convinced of how your change is going to work and I'm sure it can fix the problem you are having :dancer: 
I can pretend further I disagree tho :bug: 
:+1:

gbataille:
Sigh, so the situation seems slightly better in my production, but the issue did not disappear

image
The new code was put in prod where the cursor is. Histogram bars are issues per day. Seems to have halved (maybe) but definitely not stopped.

I'll try to spend a bit more time on it then

gbataille:
I think we can get to the `get_original_scopes` thing without being in a transaction, and having taken the refresh token lock. I'm not 100 % sure yet of how django-oauth-toolkit and oauthlib are working together

This is the call stack when it fails
```
File "oauth2_provider/views/base.py", line 206, in post
  File "oauth2_provider/views/mixins.py", line 125, in create_token_response
    return core.create_token_response(request)
  File "oauth2_provider/oauth2_backends.py", line 140, in create_token_response
    headers, extra_credentials)
  File "oauthlib/oauth2/rfc6749/endpoints/base.py", line 64, in wrapper
    return f(endpoint, uri, *args, **kwargs)
  File "oauthlib/oauth2/rfc6749/endpoints/token.py", line 117, in create_token_response
    request, self.default_token_type)
  File "oauthlib/oauth2/rfc6749/grant_types/refresh_token.py", line 59, in create_token_response
    self.validate_token_request(request)
  File "oauthlib/oauth2/rfc6749/grant_types/refresh_token.py", line 117, in validate_token_request
    request.refresh_token, request))
  File "oauth2_provider/oauth2_validators.py", line 616, in get_original_scopes
    return AccessToken.objects.get(source_refresh_token_id=rt.id).scope
```

Still digging

MattBlack85:
what error is raising?

gbataille:
Still the same `DoesNotExist: AccessToken matching query does not exist.` when trying to validate the refresh token (that has most likely been used a few milliseconds before)
I think this method in the stack is called
1. to validate the token
2. when creating the new token

And I guess I have fixed only 2 (if anything :) )

cleder:
Fixing 1 thing is better than none. 
Should the scope of this PR be reduced to fix just the one thing?

gbataille:
From my point of view yes. There is a different issue opened now for the
other thing

PR #643, opened 2018-09-09
Fix issue 636, pass request object to authenticate function.

W0126:
Pass request object to authenticate function,  request may used in authenticate function.


oauth2_provider/oauth2_validators.py
```diff
@@ -572,7 +572,7 @@ def validate_user(self, username, password, client, request, *args, **kwargs):
         """
         Check username and password correspond to a valid and active User
         """
-        u = authenticate(username=username, password=password)
+        u = authenticate(request, username=username, password=password)
         if u is not None and u.is_active:
```
cleder: `is_active` is already checked in django (>=1.11 I think) do we need to check this here again?

W0126: This fix make sure request as a parameter to authenticate function, so authentication module can get request information.

adamchainz:
This would be useful for me. I just discovered django-axes, *requires* `request` to be passed, and throws an exception if it's not: https://django-axes.readthedocs.io/en/latest/3_usage.html#authenticating-users

What's the blocker on this getting merged? I'm happy to take over.

W0126:
@adamchainz  we may need at least 1 reviewers with write access to approve this.

adamchainz:
@cleder ? Also I volunteer myself to be given write access if that helps :)

oauth2_provider/oauth2_validators.py
IvanAnishchuk: @cleder I agree regarding `is_active`, in fact Django also skips the check if user model doesn't implement `is_active`. Let's remove this part.

IvanAnishchuk: It could be a separate PR though. Approving as the fix is correct.

adamchainz:
Please revert this, it's passing a non django request to `authenticate()` !

Nocturn78:
> Please revert this, it's passing a non django request to `authenticate()` !

Exactly

n2ygk:
I will try to look at this over the weekend. Unfortunately the project
settings at the time allowed a merge without review although I’m not sure
another reviewer would have noticed as the tests passed. A PR with a
failing test that shows the error and fixes it by reverting would be
appreciated.

Updated contributing guidelines hopefully reinforce why good test coverage
is needed:
https://django-oauth-toolkit.readthedocs.io/en/latest/contributing.html

indigodomo:
When is this going to get updated in pypi? The broken version is in there now. I really don't want to have to deploy with my own copy...

n2ygk:
You can always use a git reference in your requirements.txt. For example:
```
git+https://github.com/jazzband/django-oauth-toolkit.git@9bd4c1bfcf26eecf8fde793baa48c386885c19be
```

adamchainz:
Or install from a commit has on github: https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/

PR #629, opened 2018-08-06
fix a bug in refresh token revocation

ryanpetrello:
fixes: https://github.com/jazzband/django-oauth-toolkit/issues/625

PR #653, opened 2018-10-12
Fix Refresh Token revocation when the access token does not exist

### LINKED ISSUE #625: Cannot revoke OAuth2 RefreshToken without related Token

jleclanche:
Fixes #625

----

Lands #629 on master for cherry-picking.

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -377,7 +377,10 @@ class AbstractRefreshToken(models.Model):
             if not self:
                 return
 
-            access_token_model.objects.get(id=self.access_token_id).revoke()
+            try:
+                access_token_model.objects.get(id=self.access_token_id).revoke()
+            except access_token_model.DoesNotExist:
+                pass
             self.access_token = None
             self.revoked = timezone.now()
             self.save()
--- a/tests/test_token_revocation.py
+++ b/tests/test_token_revocation.py
@@ -151,6 +151,31 @@ class TestRevocationView(BaseTest):
         self.assertIsNotNone(refresh_token.revoked)
         self.assertFalse(AccessToken.objects.filter(id=rtok.access_token.id).exists())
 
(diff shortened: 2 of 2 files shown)


jleclanche:
@jezdez What changed in this repository's config? I can't land on master, which means I can't land release update commits without PRing them which is pretty excessive. I'd like to land 1.2.1 straight from master tip, just as we did 1.2.0.

jleclanche:
1.2.1 is on hold, I'd like to land one additional fix in it.

PR #656, opened 2018-10-18
PKCE support

Maronato:
This PR adds support for the PKCE technique so that native apps can use the Authorization Code Grant securely, per [rfc 7636](https://tools.ietf.org/html/rfc7636).

Since oauthlib still [does not support](https://github.com/oauthlib/oauthlib/issues/601) PKCE, I had to intercept and alter some of the responses from oauthlib in order to maintain the API clean.

This means that I had to add the missing credentials and extra_credentials to oauthlib's Request object.

I don't know if this is the best way to implement this, so suggestions are appreciated.


jleclanche:
This looks pretty good.

The best way to implement this would be to add support in oauthlib to be honest, that'd be much preferred. But if there is consensus that regardless of oauthlib's implementation this would be how we do it in DOT (and that there's a fully backwards-compatible path should we go from one to the next), then this is okay to merge as is, albeit not ideal.

One issue: Migration should bring in non-nullable charfields and default existing ones to `""`.

Maronato:
Yeah, the best way would certainly be to use oauthlib's implementation. However, the way I implemented tries to keep most of the logic within OAuthLibCore and OAuth2Validator so, when oauthlib releases a new version with PKCE support, there shouldn't be much trouble upgrading the code.

Also, the latest commit should take care of the issue you mentioned. Let me know if anything else needs changing.

jleclanche:
Can you squash and keep to a single migration please?

Maronato:
Sure, sorry about that

gugahoa:
Any update?

JonathanHuot:
Hi @Maronato, we're going to release oauthlib-3 soon (https://github.com/oauthlib/oauthlib/issues/619) including PKCE support (https://github.com/oauthlib/oauthlib/pull/617). Do you have time to review it to see if it is compatible and easily mergeable with your PR/implementation ? 

Cheers

Maronato:
Hey @JonathanHuot !
Sorry for the delay! I will take a look tomorrow and create another PR using oauthlib/oauthlib#617.

JonathanHuot:
This PR could now be closed in favor of #678

PR #668, opened 2018-11-20
squash migrations 0001-0006

n2ygk:
@jleclanche I needed to do this anyway for overriding migrations due to django-pyodbc-azure/SQL Server deficiencies (#663, #583)  so thought I might as well submit it. Merge or Close at your pleasure.


n2ygk:
FWIW https://github.com/michiya/django-pyodbc-azure/pull/189 fixes the django-pyodbc-azure limitation that lead to this need to squash.

jnm:
@n2ygk, @auvipy, thank you for your work on this project. I do want to mention, for future reference and in light of issues like #809, that this method of squashing migrations does not follow the [long-standing](https://github.com/django/django/blame/f18da11b8a7a4f9b80798704085b37194a3627fd/docs/topics/migrations.txt#L669-L683) instructions in the [Django documentation](https://docs.djangoproject.com/en/3.2/topics/migrations/#migration-squashing):
> You should commit this migration but leave the old ones in place; the new migration will be used for new installs. Once you are sure all instances of the codebase have applied the migrations you squashed, you can delete them.

I'd humbly request that if future squashes are needed:
1. The creation of the new, squashed migration and the deletion of the old migrations do not happen in the same release;
1. The squashed migration name indicates that it is a squash (as shown in the Django documentation's example, `0001_squashed_0004_undo_somthing.py`) instead of reusing an existing migration name, which will already be logged as complete in existing installations' `django_migrations` table. That should make any problems due to missed intermediate migration steps or overlooked release notes more obvious.

Thanks for your consideration and thanks again for your work on this project.

Issue #506, opened 2017-09-03
Please update the docs


musemby:
The docs especially at https://django-oauth-toolkit.readthedocs.io/en/latest/rest-framework/getting_started.html are out of date. For instance the the import for `OAuth2Authentication`  is documented as `oauth2_provider.ext.rest_framework.OAuth2Authentication` yet as of v1.0.0 it's `oauth2_provider.contrib.rest_framework`

The docs are already updated in code but the update has not yet been deployed to read the docs. It can be pretty frustrating for first time users to figure this out. The copyright even reads `© Copyright 2013`.

fbidu:
I second that. 

A few days ago I was updating a toolkit's installation in an old project and ran into this problem. I considered fixing the docs but found out that it was already fixed on GitHub. It would be really nice to update the docs ASAP.

jieter:
@jleclanche seems easy to fix, maybe just a build error in readthedocs?

jleclanche:
I don't have access to rtd.

jieter:
OK, who does?

There seems to be a notion of version 1.0.0 at http://django-oauth-toolkit.readthedocs.io/en/1.0.0/, but it only shows an error message stating it doesn't exist.

jleclanche:
@synasius ping

musemby:
I have compiled the latest docs: http://django-oauth-toolkit-latest.readthedocs.io/en/latest/index.html for anyone that might find themselves in this situation. I will try keep it as up to date as possible. @jleclanche I propose we link to it on the README for now

alexm:
@musemby could you please make a pull request with the changes needed for the rtd build to work?

Thus, maybe it'll be easier for @synasius to review and fix this issue.

Thanks both in advance!

jleclanche:
Closing for #553

jleclanche:
Wrong issue, didn't mean to close. I'm working with @synasius to get the repository transferred to an org that can take better care of it, stay tuned :)

camilonova:
@jleclanche there was something not working here at jazzband ? Let me know how I can help.

jleclanche:
I don't know what the state of this is now that it's been transferred. I think it's fixed?

camilonova:
Looks like it was already fixed. I'm closing this.

n2ygk:
[...]

PR #671, opened 2018-12-05
Pin oauthlib to avoid breakage with 3.0.0 release.

JonathanHuot:
We're going to release OAuthLib 3.0.0 soon, which contains some breaking changes.

This pinpoint is to avoid problems before we are able to address the compatibility.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ include_package_data = True
 zip_safe = False
 install_requires =
 	django >= 2.0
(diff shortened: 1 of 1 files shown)


jleclanche:
Gtm

cleder:
What are the required changes to make dot oauthlib 3.0.0 compatible? 
Background we need the fix: `expires is no longer overriding expires_in #506`

PR #678, opened 2018-12-16
PKCE support with oauthlib 3

Maronato:
This PR modifies #656's implementation to use oauthlib's [built-in support for PKCE](https://github.com/oauthlib/oauthlib/pull/617).

Since this version of oauthlib is based on version 3, I had to modify some test cases to comply with the changes.

I also modified the tox and setup files to use the new version of oauthlib.
All modifications to the base test cases and tox/setup files are contained in a491bb1.


auvipy:
could anyone check why the build is failing?

Maronato:
Seems to be because of my attempt to support oauthlib 3 unofficially.
This dep is to blame:
https://github.com/jazzband/django-oauth-toolkit/blob/a491bb1994ae94e6d0ac9adca632185c805b6ef5/tox.ini#L23

Maronato:
I'll fix it once I get a hold of my pc

auvipy:
ping me after making travis green.

auvipy:
https://github.com/oauthlib/oauthlib/releases seems 3.0.1 is the latest

svanscho:
Any progress? Would love to use this in our new app!

sindrig:
Same here. We're anxiously waiting for this to be able to supply our vendors with the ability to use PKCE.

oauth2_provider/migrations/0007_auto_20181019_0349.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.1.1 on 2018-10-19 06:49
```
auvipy: can you remove this file and pull from upstream and regen? or could you check the migrations failures? If you don't have time and have any fix in mind let me know, will push on your branch.

Maronato:
@auvipy Sorry it took so long. Fixed the migration and pulled from master.

auvipy:
really great work! thanks for tackling this and reviewers for reviewing.

svanscho:
Thanks guys! Awesome work :-) 👍

tests/test_authorization_code.py
```diff
@@ -537,6 +539,40 @@ def get_auth(self):
...
+            "state": "random_state_string",
+            "scope": "read write",
+            "redirect_uri": "http://example.org",
+            "response_type": "code",
+            "allow": True,
+            "code_challenge": code_challenge,
+            "code_challenge_method": code_challenge_method,
+        }
+
+        response = self.client.post(reverse("oauth2_provider:authorize"), data=authcode_data)
```
Abhishek8394: I had a quick question about this. Shouldn't this be a GET request? I have been trying to get pkce to run against master but couldn't get  it to work. I would appreciate any pointers no how to get it working.

n2ygk:
Yes, GET and optionally POST.
https://tools.ietf.org/html/rfc6749#section-3.1

Abhishek8394:
Here are the two issues I ran into:

1. GET requests fail, both in tests and actual endpoint usage.

2. POST fails due to csrf tokens missing, and I think tests should have
checked for it. Should the tests be checking for this or is it entirely
intentional to skip that?

I think having POST requests will be more secure in that browsers won't
store our code challenges in history, however removing CSRF protection
might introduce bigger issues.

However the spec says GET should work too. So can I open an issue to have
PXCE working on GET requests; and work (get assigned) on it?

Maronato:
This is odd. This pull request does not change anything related to how the requests are handled. It only adds new fields to the existing ones. Would you mind sharing how you are making the requests?

Abhishek8394:
I simply enter the prepared url in browser. Although,  the app is
registered as public app and pkce setting is set to true.

I can provide a more thorough way to recreate it by end of today or
tomorrow. I'll also double check to make sure I'm not messing something on
my side.

Maronato:
I just tested it and you are correct. I was under the impression that oauthlib would capture PKCE credentials, but it doesn't. I've submitted a PR fixing the issue: #707

Abhishek8394:
Awesome! That was quick!

PR #687, opened 2019-01-10
Fix empty name of application

KirOlga:
An empty field in the application name leads to a bug in the admin panel. When you try to edit data relating to the application, instead of a clickable name, an "empty space" appears.

screenshot_20190110_200316


jleclanche:
This is not the correct fix. You should either override or change `__str__`.

hebertjulio:
This #767 pull request fix this issue.

PR #695, opened 2019-02-27
Add hook for creating custom refresh tokens

MattBlack85:
In our project we have a custom attribute on refresh token that we need to inject when creating those, for now we have the entire method save_beaer_token duplicated when we need only this bit. I thought that since there is already something similar for the AccessToken maybe the same path could be applied for RefreshToken

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -520,13 +520,7 @@ class OAuth2Validator(RequestValidator):
                         source_refresh_token=refresh_token_instance,
                     )
 
-                    refresh_token = RefreshToken(
-                        user=request.user,
-                        token=refresh_token_code,
-                        application=request.client,
-                        access_token=access_token
-                    )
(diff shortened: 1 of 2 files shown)


MattBlack85:
I am not changing anything here so I am not sure what you asking for @auvipy, the internal remains exactly the same but the plus is that the refresh tokens creation will be aligned with the access one

auvipy:
this is a common practice followed by popular projects. as it is said that, "code without tests is broken by design." never mind. You did great work. tests needed for formally validating any upcoming change :+1:

jleclanche:
@MattBlack85 I think @auvipy is asking for a test that makes use of create_refresh_token() so that it's not accidentally removed again. I would tend to agree.

MattBlack85:
@auvipy @jleclanche I added a test for my change that covers also `_create_access_token`, let me know if this is what you meant

auvipy:
thanks a lot!

PR #707, opened 2019-04-19
Fix PKCE credentials not being captured during authorize requests

Maronato:
PKCE credentials are not being captured during GET authorize requests. oauthlib should(?) return `code_challenge` and `code_challenge_method` from [grant_types.authorization_code.validate_authorization_request](https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py#L318).

@JonathanHuot, if oauthlib does not return them, clients using oauthlib must capture them manually. Is that intended?

### DIFF
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -32,6 +32,7 @@ class BaseAuthorizationView(LoginRequiredMixin, OAuthLibMixin, View):
     * Implicit grant
 
     """
+
     def dispatch(self, request, *args, **kwargs):
         self.oauth2_data = {}
         return super().dispatch(request, *args, **kwargs)
@@ -132,6 +133,8 @@ class AuthorizationView(BaseAuthorizationView, FormView):
     def get(self, request, *args, **kwargs):
         try:
             scopes, credentials = self.validate_authorization_request(request)
+            credentials["code_challenge"] = request.GET.get("code_challenge", None)
+            credentials["code_challenge_method"] = request.GET.get("code_challenge_method", None)
         except OAuthToolkitError as error:
             # Application is not available at this time.
             return self.error_response(error, application=None)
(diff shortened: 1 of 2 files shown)


auvipy:
somewhat weired

Maronato:
> somewhat weired

What you think is weird?

JonathanHuot:
Hi @Maronato, oauthlib should receive the full request (method, uri, headers, payload). So if oauthlib receives correctly `code_challenge `& `code_challenge_method`, then it should works. 

Do you know if DOT is filtering the list of fields before sending them to oauthlib ?

Also, do you have specific use-case/issue to how reproduce it ?

oauth2_provider/views/base.py
```diff
@@ -132,6 +133,8 @@ def form_valid(self, form):
     def get(self, request, *args, **kwargs):
         try:
             scopes, credentials = self.validate_authorization_request(request)
+            credentials["code_challenge"] = request.GET.get("code_challenge", None)
+            credentials["code_challenge_method"] = request.GET.get("code_challenge_method", None)
```
JonathanHuot: Have you tried to not specify these two lines ? These seems unused until overriden lines 155/156.

JonathanHuot:
Ahhh sorry I think I get it. Yes, you're right, it should be added into oauthlib's `validate_authorization_request`.

I can fix it in oauthlib.

JonathanHuot:
Hi @Maronato, any chances to test https://github.com/oauthlib/oauthlib/pull/671 before releasing it ? Thx

Maronato:
Just tested it and it seems to be working fine, @JonathanHuot! Let me know when oauthlib/oauthlib#671 is released so I can revert the workaround.

darvids0n:
FYI, the latest oauthlib includes these changes. https://github.com/oauthlib/oauthlib/tree/v3.1.0

Issue #633, opened 2018-08-15
Undocumented (error prone) django-rest-framework authentication side-effect


gbataille:
Hello,

In here https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/contrib/rest_framework/authentication.py, the `authenticate` method has a hidden that effect. On authentication error, it sets a `oauth2_error` attribute on the request and then `authenticate_header` assumes it's there.

This caused me an issue because I have subclasses `OAuth2Authentication` and to override `authenticate`. Of course I did not know of this "side effect" and a lot of tests started to fail on my project. Was quite hard to debug.

Granted it's odd that we override this class, we have legacy reasons to do so. I think the `authenticate_header` method should be more relaxed and look like
```
if hasattr(request, 'oauth2_error'):
    www_authenticate_attributes.update(request.oauth2_error)
```

What do you think?

PR #716, opened 2019-05-29
Use getattr for oauth2_error access (#633)

george-pearson:
If the request doesn't have a oauth2_error property the
authenticate_header method errors. This can happen when the
oauthlib_core.verify_request method raises exceptions in authenticate.
It is useful to be able to raise AuthenticationFailed exceptions from
within a custom validate_bearer_token method which causes this.


george-pearson:
Hi @auvipy, I've added a simple test that will highlight the problem if run on master.

n2ygk:
@auvipy please close #633 if you agree this merged PR fixes that issue.

PR #723, opened 2019-07-23
upgrades to admin: search_fields, list_filters and raw_id_field

PetrDlouhy:
Few upgrades to make admin better arranged.

### DIFF
--- a/oauth2_provider/admin.py
+++ b/oauth2_provider/admin.py
@@ -19,16 +19,22 @@ class ApplicationAdmin(admin.ModelAdmin):
 class GrantAdmin(admin.ModelAdmin):
     list_display = ("code", "application", "user", "expires")
     raw_id_fields = ("user", )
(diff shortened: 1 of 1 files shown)


oauth2_provider/admin.py
```diff
@@ -19,16 +19,22 @@ class ApplicationAdmin(admin.ModelAdmin):
 class GrantAdmin(admin.ModelAdmin):
     list_display = ("code", "application", "user", "expires")
     raw_id_fields = ("user", )
+    search_fields = ("code", "user__email")
```
IvanAnishchuk: Hmm... You sure this won't break things it user model doesn't have a email field?

jleclanche: +1. Rest looks good.

MattBlack85:
closing due lack of tests at this moment, feel free to re-open in case!

PR #725, opened 2019-07-29
HTTP Basic Auth support for introspection (Fix issue #709)

### LINKED ISSUE #709: HTTP Basic Auth support for introspection
**Expanding upon the issue #529** 

It seems that introspection endpoint is not allowing client authentication via HTTP Basic auth. As mentioned in the comments in the code at [introspect.py#L13](https://github.com/jazzband/django-oauth-toolkit/blob/2fdb0fea4efec9bc79708f0ac1b4b24884f26f97/oauth2_provider/views/introspect.py#L13) 
The RFC allows for HTTP basic auth: [https://tools.ietf.org/html/rfc7662#section-2.1](https://tools.ietf.org/html/rfc7662#section-2.1)

It seems that the authentication is bound to access tokens. This is a problem because the only straight forward way to get access tokens is using `client_credentials` grant type. But currently apps are allowed only one grant type. 

**Reproducing issue**
1. Create an `confidential` application with grant type `authorization_code`
2. Get an access token for some user with respect to the newly created app.
3. send introspection request

```python
import requests
import base64
CLIENT_ID='aa'
CLIENT_SECRET='aa'

def introspect(access_token):
	payload = {
		'token': access_token
	}
	auth = base64.b64encode((CLIENT_ID + ':' + SECRET).encode())
	resp = requests.post('http://localhost:8000/o/introspect/', payload, 
		headers={
			'Authorization': 'Basic ' + auth.decode('latin-1')
		})
	return resp

print(introspect('tokenhere').text)
```

There is a workaround proposed for going into django admin and creating an access token for the app but that wouldn't scale. 

An alternative would be to use the setting for resource auth token and exclusively allow it only for introspection routes. That might work too, but it would have you leak your introspection token across several resource services. 

I would like to work on this if this qualifies as a bug/issue.

Abhishek8394:
Fixes #709 

- Add a new mixin that allows authenticating with HTTP basic auth, credentials in body or access tokens
- Introduce and abstraction in views.generic to initialize the OauthLibMixin
- Change parent class of IntrospectTokenView from 'ScopedProtectedResourceView' to 'ClientProtectedScopedResourceView'

### Checklist
- [x] unit-test added
- [x] documentation updated
- [x] CHANGELOG.md updated (only for user relevant changes)
- [x] author name in AUTHORS


oauth2_provider/views/mixins.py
```diff
@@ -245,3 +254,29 @@ def get_scopes(self, *args, **kwargs):
...
+    Breaks off after first validation. 
+    """
+
+    def dispatch(self, request, *args, **kwargs):
+        # let preflight OPTIONS requests pass
+        if request.method.upper() == "OPTIONS":
+            return super().dispatch(request, *args, **kwargs)
+        # Validate either with HTTP basic or client creds in request body.
+        # TODO: Restrict to POST.
+        valid = self.authenticate_client(request)
```
Abhishek8394: I feel like this should also get `valid, r` where `r.user` is user associated with authenticated client. This would allow introspection to verify if given client is authorized to introspect given token. Can someone verify if this is the correct approach?

Abhishek8394:
I can work on  merging master in the branch if no one is already working on it

auvipy:
could you check the latest failures?

Abhishek8394:
Fixed the tests, issue was sending query params in a POST request. That is not allowed.
Also the travis test ran only on py34-django20? It should run on all versions till py37, correct?

Abhishek8394:
Also maybe update the docs to tell people that sending query params in any request except GET is a bad request.

oauth2_provider/views/mixins.py
```diff
@@ -245,3 +254,29 @@ def get_scopes(self, *args, **kwargs):
...
+            # Alternatively allow access tokens
+            # check if the request is valid and the protected resource may be accessed
+            valid, r = self.verify_request(request)
+            if valid:
+                request.resource_owner = r.user
+                return super().dispatch(request, *args, **kwargs)
+            else:
+                return HttpResponseForbidden()
+        else:
+            return super().dispatch(request, *args, **kwargs)
```
auvipy: no newline

Abhishek8394:
Fixed this. Can you merge this now?

> *@auvipy* commented on this pull request.
> ------------------------------
>
> In oauth2_provider/views/mixins.py
> <https://github.com/jazzband/django-oauth-toolkit/pull/725#discussion_r324292893>
> :
>
> > +        if request.method.upper() == "OPTIONS":
> +            return super().dispatch(request, *args, **kwargs)
> +        # Validate either with HTTP basic or client creds in request body.
> +        # TODO: Restrict to POST.
> +        valid = self.authenticate_client(request)
> +        if not valid:
> +            # Alternatively allow access tokens
> +            # check if the request is valid and the protected resource may be accessed
> +            valid, r = self.verify_request(request)
> +            if valid:
> +                request.resource_owner = r.user
> +                return super().dispatch(request, *args, **kwargs)
> +            else:
> +                return HttpResponseForbidden()
> +        else:
> +            return super().dispatch(request, *args, **kwargs)
>
> no newline
>
>
>

tests/test_token_revocation.py
```diff
@@ -52,13 +52,13 @@ def test_revoke_access_token(self):
...
-        query_string = urlencode({
+        payload = {
             "client_id": self.application.client_id,
             "client_secret": self.application.client_secret,
             "token": tok.token,
-        })
-        url = "{url}?{qs}".format(url=reverse("oauth2_provider:revoke-token"), qs=query_string)
-        response = self.client.post(url)
+        }
+        url = "{url}".format(url=reverse("oauth2_provider:revoke-token"))
```
IvanAnishchuk: This is rather long. Why not just `url = reverse("oauth2_provider:revoke-token"`?

Abhishek8394:
I got the test fixes from master. Is there anything more required to get this PR merged?

vijithv:
Hi, any updates on merging this PR?

vijithv:
Hey can we merge this? Tested the PR and it works fine. Just don't want to maintain a fork

Abhishek8394:
I'll get a fix out in a day or two. In the meanwhile do verify the checklist at the top of the page.

Abhishek8394:
There is not setting or configuration required, it just works and is backwards compatible so existing users won't fail. Detail explanation below:

So previously introspection only used to work with access tokens. Now it checks for client authentication first, if that fails then it falls back to checking the access token. Client authentication works as follows:
- Check for HTTP Basic Auth
- If prev fails, check for client_id and client_secret in request body (get/post params).
- If prev fails, fail client auth.

Abhishek8394:
Also documentation at [#setup-the-resource-server](https://django-oauth-toolkit.readthedocs.io/en/latest/resource_server.html#setup-the-resource-server) seems to document about using HTTP basic Auth for introspection. Let me know if I'm missing something on this.

PR #740, opened 2019-09-30
Authenticate is supposed to take the request in param

gbataille:
Sometimes the request is needed (for audit tracking of logging attempts for example)

https://github.com/django/django/blob/4116b369b191f1830bbe761fa7fb29bf1becc8ef/django/contrib/auth/backends.py#L16
https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#writing-an-authentication-backend


gbataille:
Tests are failing but they fail also on the commit before

PR #741, opened 2019-09-30
Fixed unpaired parentheses

MeiK2333:
Fixed unpaired parentheses, At the same time, Django now provides `path` by default. Should we change the `url` to `path`?


jleclanche:
> Should we change the url to path?

Sure.

Issue #732, opened 2019-08-17
Implementing custom token generator


KhashayarMirMS:
Hi,
I want to implement a custom token generator to use JWT in DOT.
There are several issues related to this problem which lead to this pull request: #467. But the pull request is about writing custom models for tokens which doesn't handle token generation.

The solution that I've currently found is to write a custom `Server Class` to pass a custom generator function to `BearerToken`. But I've seen that `Server Class` instances are initialized with a `token_generator` function passed to them, but I can't find out where the instances are initialized.

I would appreciate if you could help me with this matter.

JonathanHuot:
Hi, do you want to generate access_token as JWT ? If yes, please have a look at https://oauthlib.readthedocs.io/en/latest/oauth2/tokens/bearer.html

KhashayarMirMS:
Hi, yes I wanted to generate access token as JWT and I used the documentation you mentioned for the matter, my question can be related to this part of the documentation:
![image](https://user-images.githubusercontent.com/11523685/63967505-d2bb7200-cab2-11e9-89d9-17f89d38ba2e.png)
This initializes a new server object. But I couldn't find how to tell DOT to use this instance for it's OauthBackend instance.

IvanAnishchuk:
I think there are currently multiple places in the code where server instance is being initialized (three, if I remember correctly) and only one of them can be subclassed easily (`OAuthLibCore.__init__`). I think a good solution would be to add a settings option to allow passing custom kwargs to those server instances and maybe a couple specific settings for passing `token_generator` and `refresh_token_generator` there specifically (in a way that supports passing import strings in settings and perhaps making them classes that can accept parameters before being passed to server, client_id/client_secret generators are class-based and accept parameters, sounds like a usable pattern). This sounds like a small enough change that shouldn't break compatibility, let me see if I can fix a PR for this quickly....

Note however that you're likely to need to use a custom `Server` class anyway, sooner or later, if you want to customize anything regarding auth flow, You might as well start now, that's the simplest way to achieve what you need right now without updating the library.

PR #749, opened 2019-10-19
Add extension points for custom token generators (#732)

IvanAnishchuk:
Pass variables generated based on settings when initializing Server.


oauth2_provider/oauth2_validators.py
```diff
@@ -478,7 +500,11 @@ def save_bearer_token(self, token, request, *args, **kwargs):
         if "scope" not in token:
             raise FatalClientError("Failed to renew access token: missing scope")
 
-        expires = timezone.now() + timedelta(seconds=oauth2_settings.ACCESS_TOKEN_EXPIRE_SECONDS)
+        # expires_in is passed to Server on initialization
+        # custom server class can have logic to override this
+        expires = timezone.now() + timedelta(seconds=token.get(
+            'expires_in', oauth2_settings.ACCESS_TOKEN_EXPIRE_SECONDS,
```
jleclanche: Looks unrelated

IvanAnishchuk: Totally related, this changeset is about the way to configure Server instance and since we have the settings variable already it makes total sense to pass it there too now that a way to pass variables is added.

oauth2_provider/views/mixins.py
```diff
@@ -74,7 +74,12 @@ def get_server(cls):
         """
         server_class = cls.get_server_class()
         validator_class = cls.get_validator_class()
-        return server_class(validator_class())
+        server_kwargs = (
+            validator_class._get_server_kwargs()
```
MattBlack85: can we avoid using conditional expression here please?

IvanAnishchuk: Let me think about it for a moment. I want it to remain compatible with custom validator classes though, not of them necessarily subclass DOT's one.

We could probably just skip the logic I added in that method completely, and just pull the kwargs from settings as is here. But I want to avoid uncertainty as to how configuration values are set and used and custom token generators should be import strings and expires_in stuff already exists as separate variables. Yet ability to pass arbitrary kwargs seems important for custom servers too. Decision to add a validator method for this is based on the idea that we might not be able to figure out exact way to form kwargs for heavily customized setups but if integrators can subclass request validator they can have control over everything that happens there which wouldn't work as easily if we add code to settings module for example (most of the times when using custom server it makes sense to customize validator as well, and server can't be instantiated without a validator).

But I agree that this is kinda ugly and a bit more complex than probably needed. Let me see if I can improve it.

MattBlack85: I am fine with your addition, but i see here some duplicated code:
 this module L77
 oauth2_provider/oauth2_backends.py L206
 oauth2_provider/oauth2_backends.py L 20

What I think is maybe we can set the validator_class as class attribute or within the `__init__ ', that way we can create a new method `get_server_kwargs`

```Python
def get_server_kwargs(self):
    if hasattr(self.validator_class, '_get_server_kwargs'):
        return self.validator_class._get_server_kwargs()
    else:
        return {}
```

oauth2_provider/oauth2_validators.py
IvanAnishchuk: Also, it's required for a token generator (that can be unaware of DOT and its settings) to have access to this value -- it's normally passed as a request attribute (or its result when it's a callable).

oauth2_provider/views/mixins.py
IvanAnishchuk: Yeah, that would work in that case (I even thought originally to add it to the backend class and not server), but yet another place where a `Server` is currently initialized is `OAuthLibMixin` and I didn't like the idea of relying on `oauthlib_backend_class` being always correctly specified in any its subclasses and be available (it requires a server for instantiation) while request_validator can be assumed to be always present (you can't instantiate a server without it). Note how `get_oauthlib_core` is structured -- it creates a validator instance and a server instance before accessing OAUTH2_BACKEND_CLASS.

Also, two wrapper methods on two different classes just to merge a couple dictionaries essentially feels like an overkill. There ought to be a better way. How about some tweaking in settings? Does making `oauth2_settings.SERVER_KWARGS` a callable (that generates the dictionary based on the other settings) sound any better than this?

IvanAnishchuk: @MattBlack85 I decided we don't really need a validator method for this too badly. The settings dictionary is enough for most cases and when classes are heavily customized it doesn't matter anyway. So I added a property to the settings class instead. How do you like it now?

MattBlack85:
@IvanAnishchuk is a migration file really needed for tests only models?

tests/settings.py
```diff
@@ -74,6 +79,7 @@
     "django.contrib.sites",
     "django.contrib.staticfiles",
     "django.contrib.admin",
+    "django.contrib.messages",
```
MattBlack85: why is this needed here for tests?

IvanAnishchuk: Because `admin` is included.

tests/test_oauth2_validators.py
```diff
@@ -284,8 +284,8 @@ def test_save_bearer_token__with_new_token__calls_methods_to_create_access_and_r
 
         self.validator.save_bearer_token(token, self.request)
 
-        create_access_token_mock.assert_called_once()
-        create_refresh_token_mock.asert_called_once()
+        assert create_access_token_mock.call_count == 1
```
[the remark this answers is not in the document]
IvanAnishchuk: `asert_called_once` is typo, to begin with. Also https://github.com/PyCQA/pycodestyle/issues/624

IvanAnishchuk:
> @IvanAnishchuk is a migration file really needed for tests only models?

Yes, if we want tests to work on some of the versions and don't want to disable all migrations in tests. Otherwise everything just fails with setup error (caused by non-migrated models depending on migrated models as far as I could figure it out) but feel free to suggest alternative fixes if you find any, I don't like this either.

oauth2_provider/views/mixins.py
MattBlack85: :+1: :100:

MattBlack85:
> > @IvanAnishchuk is a migration file really needed for tests only models?
> 
> Yes, if we want tests to work on some of the versions and don't want to disable all migrations in tests. Otherwise everything just fails with setup error (caused by non-migrated models depending on migrated models as far as I could figure it out) but feel free to suggest alternative fixes if you find any, I don't like this either.

which versions specifically?

IvanAnishchuk:
[Django 3.0](https://travis-ci.org/jazzband/django-oauth-toolkit/builds/602472403) there's also a similar error with django 2.2 on python 3.4 but that seems to be a separate problem.

tox.ini
```diff
@@ -19,6 +24,8 @@ setenv =
 deps =
 	django20: Django>=2.0,<2.1
 	django21: Django>=2.1,<2.2
+	django22: Django>=2.2,<3
+	django30: Django>=3.0a,<3.1
```
auvipy: 3.0b1

IvanAnishchuk: It's `>=` so it'll be whatever the latest version is.

auvipy: then 3.0a1 as it's on pypi

IvanAnishchuk: It's not gonna change what's installed but sure, changed.

aleberguer:
Hi all, I'm trying out this PR to add JWT access' token.
I'm using `signed_token_generator` trough `EXTRA_SERVER_KWARGS` settings.
I'm wondering if there is a way to enrich the generated JWT with some user data such as username and scopes.
Currently, the scopes returned in the JWT are null.
Thank you!

MattBlack85:
@IvanAnishchuk I will find some time this week to test properly the tox matrix, the migration still doesn't look too right to me so I'd like to investigate this a bit, other than that :+1:

PR #751, opened 2019-10-20
Use Model.objects.create where applicable

jleclanche:
Reopen #747 which was closed by mistake after a bad rebase.


adamchainz:
Cheers

Issue #753, opened 2019-10-23
Skip auth form if valid refresh token exists?


madprime:
According to [documentation on skipping the authorization form](https://django-oauth-toolkit.readthedocs.io/en/latest/advanced_topics.html#skip-auth-form), `auto` can be used to skip the authorization form if an application has already been authorized with the same scope.

However as it's currently implemented, this only seems to work as long as an access token exists that is valid (default expiration: 10 hours) and has matching scope. But it _doesn't_ check for a valid refresh token.

If a valid refresh token exists (and its corresponding access token has matching scopes), it seems the app has equivalent evidence of valid authorization from that user? And so one could (should?) safely skip the authorization form? Unless I'm missing something, this would seem to match the anticipated behavior as described in documentation.

If that all sounds correct, I think a change to `oauth2_provider/views/base.py` can make this happen, and I'll open a related PR. :)

IvanAnishchuk:
Merged too quickly. Converting querysets to lists there is a bad idea.

madprime:
I did expect more discussion – especially about the intended **change in logic made on the code for auto-authorization**, i.e. whether it is correct to change it in this way.

IvanAnishchuk:
Yep. Let's continue the discussion then and figure out the correct way of doing it.

I think it is, generally speaking, a correct approach (it's common to skip authorization decision page if user has previously approved a request with the same scope for the same application, it also does make sense that refreshing a token or anything other than explicit user actions has side-effects on auto-authorization), but maybe we should look at the grants rather than/in addition to access/refresh tokens -- theoretically there can be access tokens issued for some other grant type that user didn't explicitly approve (not a common situation with DOT, applications support only a single grant type by default).

Existence of a Grant can confirm "user authorization decision" i.e. that user clicked the "authorize" button with the same scope and AccessToken or RefreshToken can confirm that user hasn't logged out (revoked access) since. I think we need to establish both facts to consider approval established "via other means" ([4.2.1](https://tools.ietf.org/html/rfc6749#section-4.2.1)).

[...]

PR #754, opened 2019-10-23
Auto-authorize if valid refresh tokens exist

madprime:
This implements a check on refresh tokens for `auto` skipping of the authorization form, as described in #753.

### DIFF
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -12,7 +12,7 @@ from django.views.generic import FormView, View
 from ..exceptions import OAuthToolkitError
 from ..forms import AllowForm
 from ..http import OAuth2ResponseRedirect
-from ..models import get_access_token_model, get_application_model
+from ..models import get_access_token_model, get_application_model, get_refresh_token_model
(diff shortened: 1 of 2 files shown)


madprime:
@auvipy thanks, added a test which would have previously failed, and passes now. see what you think?

n2ygk:
@madprime @auvipy I'm trying to document PRs merged between 1.2.0 and forthcoming 1.3.0 (#785) that did not make it into the CHANGELOG.md. Can you provide a one-line explanation of this feature. Is it a bug fix or added feature? Also, is this something to be documented in docs/? A documentation PR which would greatly appreciated if so.

I don't quite understand how this automatic use of Refresh Token is used. Is there a reference to the OAuth2 RFCs?

Thanks.

auvipy:
according to https://github.com/jazzband/django-oauth-toolkit/issues/753#issuecomment-548028519 and another WIP PR, I think we should revert this for now.

auvipy:
please check https://github.com/jazzband/django-oauth-toolkit/pull/755

Issue #235, opened 2015-04-23
support for oob/urn:ietf:wg:oauth:2.0:oob redirect uris


vegitron:
I have client applications that aren't able to run a web server, or get redirects to a custom protocol.  These are unix command line tools, not phone apps or websites.   I would need to have users copy/paste their access codes so the app can get the tokens.

I was able to hack this functionality into place, by allowing oob:// as a custom protocol.  Then in my service provider I subclassed the AuthorizationView, to turn any redirects to an oob:// url into a redirect to a custom view in my service provider that shows the access code:

class OOBAuthorizationView(AuthorizationView):                                   

```
def post(self, *args, **kwargs):                                            
    value = super(SSAuthorizationView, self).post(args, **kwargs)           

    location = value.get("Location", "")                                    
    if location and re.match("oob://", location):                           
        new_base = reverse("oauth_access_code")                             
        new_url = "%s?code=" % new_base                                     
        location = re.sub("^oob://.*\?code=", new_url, location)            
        value["Location"] = location                                        
    return value   
```

I would really like to be able to get rid of this hack.

masci:
Hi @vegitron, do you mind to elaborate a bit on what is the workflow you're currently using, I'm not sure I understood what you need. TIA

vegitron:
Sure.  Part of my application is a command line tool, to let people either use my app as part of a script with other tools, or do things like uploading large datasets that they wouldn't want to do through their browser.

Considering the various authorization grant types:

Client credentials isn't appropriate, since this is manipulating user data, and definitely isn't trusted.
Password doesn't work for me, because my users are authenticating through 3rd party resources.
Implicit doesn't work because there's no browser to see the hash.
Authorization code does what I want - it forces user authentication, and application approval.

So what I'm doing is something like this: 

```
./bin/my_app upload file.dat
You need to log in to continue.  After logging in, enter your access code below.
Log in at  https://myserver.com/o/?...

Access code:
```

The user logs in, approves the application access, and sees a page that gives them the code they need to enter.  They enter it into the app, the app uses that to request the oauth access token and refresh  token, and then does it's work.   That access code is the same one that would have been given in the redirect uri.

Does that make my goal more clear?  Thanks!

kevin-brown:
I'm going to :+1: this.

[Here's some information on how Google handles it.](https://developers.google.com/identity/protocols/OAuth2InstalledApp#choosingredirecturi)

tcyrus:
[...]

PR #774, opened 2019-12-30
Fix for issue #235 (urn:ietf:wg:oauth:2.0:oob support)

marnanel:
This adds support for the "urn:ietf:wg:oauth:2.0:oob" and "urn:ietf:wg:oauth:2.0:oob:auto" URNs, as discussed in issue #235 .

The dict returned by "urn:ietf:wg:oauth:2.0:oob:auto" needs work because it's currently missing several members which are inspected by the standard client. I'm submitting this PR now to check whether I'm barking up the wrong tree further up, so to speak.


marnanel:
Coverage is now increasing.

n2ygk:
In the process of trying to document what's in the next release (#785) I googled this feature and see discussion of it being a non-standard google-only [deprecated practice](https://mailarchive.ietf.org/arch/msg/oauth/5CkOmL7sP2f_gHcbwknSY0MD60w/) that is replaced by [RFC 8252](https://tools.ietf.org/html/rfc8252) which offers a few mechanisms to get the same result for a command-line app.

Consider deprecating this in a future release.  I'll put a warning in the Changelog.

PR #788, opened 2020-02-27
drop py 34

auvipy:


kevin-brown:
This should also be dropped from the `setup.cfg` and `tox.ini` files.

auvipy:
ok thanks I will

Issue #807, opened 2020-03-08
unsupported_grant_type on rest example


Blanen:
**Describe the bug**
Doing the rest example with:

```
asgiref==3.2.3
certifi==2019.11.28
chardet==3.0.4
Django==3.0.4
django-oauth-toolkit==1.3.0
djangorestframework==3.11.0
idna==2.9
oauthlib==3.1.0
psycopg2==2.8.4
pytz==2019.3
requests==2.23.0
sqlparse==0.3.1
urllib3==1.25.8
```

doesn't work and simply returns this error:
`{"error": "unsupported_grant_type"}`

and this in the log:

```
Bad Request: /o/token/
[08/Mar/2020 04:31:42] "POST /o/token/ HTTP/1.1" 400 35
```

**To Reproduce**
Do the example

**Expected behavior**
It should work.

**Version**

- [x] I have tested with the latest published release and it's still a problem.
- [x] I have tested with the master branch and it's still a problem. 

**Additional context**
Quite hard to believe this, honestly.

Blanen:
The normal tutorial also mentions that it needs a valid redirect url, but the REST tutorial says nothing about it.

baranberkay96:
`OAUTH2_PROVIDER = {
    # parses OAuth2 data from application/json requests
    'OAUTH2_BACKEND_CLASS': 'oauth2_provider.oauth2_backends.JSONOAuthLibCore',
}`

In your settings.py,  do you have any configuration like that?

dnjscksdn98:
I have the same error as you. Did you fix the error?

Blanen:
@baranberkay96 yep

Blanen:
@dnjscksdn98 nope

dnjscksdn98:
> @dnjscksdn98 nope

After I deleted `'oauth2_provider.oauth2_backends.JSONOAuthLibCore'` 
it worked. But I don't know why

n2ygk:
The default value in `oauth2_provider/settings.py` is `oauth2_provider.oauth2_backends.OAuthLibCore`.
This sounds like perhaps a documentation error or otherwise incorrect workaround made in #734.

@rudolfolah Can you take a look at this and see if you can find a fix to your error that doesn't break other instances and submit a PR? Meanwhile, I'll go ahead and revert this PR for 1.3.1.

[...]

PR #817, opened 2020-03-20
Revert "getting_started.rst: add JSONOAuthLibCore as part of tutorial"

n2ygk:
Reverts jazzband/django-oauth-toolkit#734 per #807

### DIFF
--- a/docs/rest-framework/getting_started.rst
+++ b/docs/rest-framework/getting_started.rst
@@ -100,8 +100,6 @@ Also add the following to your `settings.py` module:
 .. code-block:: python
 
     OAUTH2_PROVIDER = {
-        # parses OAuth2 data from application/json requests 
-        'OAUTH2_BACKEND_CLASS': 'oauth2_provider.oauth2_backends.JSONOAuthLibCore',
         # this is the list of available scopes
         'SCOPES': {'read': 'Read scope', 'write': 'Write scope', 'groups': 'Access to your groups'}
     }
@@ -113,10 +111,6 @@ Also add the following to your `settings.py` module:
             'rest_framework.permissions.IsAuthenticated',
         )
     }
-
-`OAUTH2_PROVIDER` setting parameter sets the backend class that is used to parse OAuth2 requests.
-The `JSONOAuthLibCore` class extends the default OAuthLibCore to parse correctly
-`application/json` requests.
 
 `OAUTH2_PROVIDER.SCOPES` setting parameter contains the scopes that the application will be aware of,
 so we can use them for permission check.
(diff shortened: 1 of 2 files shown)


n2ygk:
@auvipy see reason why I reverted this at https://github.com/jazzband/django-oauth-toolkit/issues/807#issuecomment-601691819

docs/rest-framework/getting_started.rst
```diff
@@ -114,10 +112,6 @@ Also add the following to your `settings.py` module:
         )
     }
 
-`OAUTH2_PROVIDER` setting parameter sets the backend class that is used to parse OAuth2 requests.
```
rudolfolah: @n2ygk I would leave this portion of the PR in to let other devs know that it *is* an option to use JSONOAuthLibCore. As I mentioned in my original PR, the Angular HTTP library wasn't set up for using the correct content-type as per the OAuth RFC you linked.

Now that I think about it, maybe we should leave this part in and add on to it:

>you can also try to configure your HTTP requests to use the RFC-mandated header for OAuth tokens: `Content-Type: application/x-www-form-urlencoded`

What are your thoughts?

n2ygk: Can you cite a reference to using JSON-encoded data for the OAuth2 token POST being even allowed? All references I find seem to require form-encoded. Maybe I'm not reading the RFCs carefully enough?

There are angular libraries that work just fine as an OAuth2 client using POST with form-encoding. See https://github.com/manfredsteyer/angular-oauth2-oidc.

I really don't understand why JSONOAuthLibCore was added 5 years ago in the first place and see no usages of it anywhere in the code other than the test case that was added with it. I think it may have been added erroneously.

A documentation PR to clarify the required use of form-encoding would be appreciated.

rudolfolah: >Can you cite a reference to using JSON-encoded data for the OAuth2 token POST being even allowed? All references I find seem to require form-encoded. Maybe I'm not reading the RFCs carefully enough?

I think I didn't read the RFC closely enough (or the documentation on OAuth libraries). I ran into this issue while developing a video course.

>I really don't understand why JSONOAuthLibCore was added 5 years ago in the first place and see no usages of it anywhere in the code other than the test case that was added with it. I think it may have been added erroneously.

I'm sure someone else was implementing an OAuth client for production use and skipped over the Content-Type issue too.

>A documentation PR to clarify the required use of form-encoding would be appreciated.

I think your comments are that clarification, thanks for explaining how it all works and referencing the RFC too

auvipy:
thanks for finding out this!

PR #829, opened 2020-04-09
subclass Abstract models

dulmandakh:
## Description of the Change

Update document to subclass Abstract models from oauth2_provider.

## Checklist

- [x ] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [ x] documentation updated


n2ygk:
@dulmandakh will you be able to do this? Also, please rebase.

> Thanks for this. Please add your name to AUTHORS and update CHANGELOG to say you've corrected the documentation.
> 
> Also, would it make sense to add something to the tutorials about how to get swappable models to work with this stuff beyond the current documentation which I believe only documents swappable Application model?

PR #831, opened 2020-04-10
split authorization code creation into OAuth2Validator._create_authorization_code 

dulmandakh:
## Description of the Change

I'm subclassing OAuth2Validator to customize authorization code generation, and splitting authorization code creation into _create_authorization_code like _create_access_token, _create_refresh_token would make it easy. Otherwise, I would copy paste same code from save_authorization_code to customize it.

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk:
@dulmandakh: This is a change to a private API, but maybe if the subclassing you are doing is generally useful, this could be made into a public API change -- which would require documentation, new tests, etc. Can you tell us more about your use case?

PR #848, opened 2020-06-16
Fix #847 Log an exception when response status is not OK

### LINKED ISSUE #847: Return appropriate error messages when response from authentication server is not successful

If the introspection_token is not correct in the resource server, then a JSONDecoderError is raised when there is a token to be verified via oauth2_provider. This kind of logger error and exception would make the debugging procedure quite complicated because the user wouldn't know the cause of why there is a JSON_DecoderError.

Related method: [oauth2_validators.py line 307 - 311.
](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L307)

Example traceback in resource server when the authentication server returns a 403 response to the resource server.
```
Introspection: Failed to parse response as json
Traceback (most recent call last):
  File "C:\Python37\lib\site-packages\oauth2_provider\oauth2_validators.py", line 308, in _get_token_from_authentication_server
    content = response.json()
  File "C:\Python37\lib\site-packages\requests\models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "C:\Python37\lib\json\__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "C:\Python37\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\Python37\lib\json\decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

If the response from authentication server is not successful(e.g. status code is not 200), then an exception should be logged with the response's status code and its cause.

E.g. If we add the following code block before [line 307](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L307):
```
        if response.status_code != 200:
            log.exception("Introspection: Failed to get a valid response "
                          "from authentication server. status code ={}, "
                          "reason {}".format(response.status_code, 
                                             response.reason))
```

Then, the traceback error message would be:
```
Introspection: Failed to get a valid response from authentication server. status code =403, reason Forbidden
Traceback (most recent call last):
  File "/home/jun/.local/share/virtualenvs/maio-engine-OAc_wQKR/lib/python3.8/site-packages/oauth2_provider/oauth2_validators.py", line 371, in validate_bearer_token
    access_token = AccessToken.objects.select_related("application", "user").get(token=token)
  File "/home/jun/.local/share/virtualenvs/maio-engine-OAc_wQKR/lib/python3.8/site-packages/django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
oauth2_provider.models.AccessToken.DoesNotExist: AccessToken matching query does not exist.
```

Ideally, all connections with authentication server using RESOURCE_SERVER_AUTH_TOKEN should return the error messages and its status code back to resource server if the response is not successful.

**Additional context**

Response from authentication server:
`WARNING 2020-06-16 11:53:11,274 basehttp 18318 140462266509056 "POST /oauth2/introspect/ HTTP/1.1" 403 0`

Traceback image without verification:
image

Chouvic:
By providing the status code and its reason to logger, the error message would be more explicit for debugger whenever there is an issue related to resource token or other errors.

Fixes #847 

## Description of the Change
Added a checker to log an exception when the response from authentication server is not successful. With such change, the real issue with the response would not be changed to a JSONDecoderError which would be a distraction for debugging.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,3 +29,4 @@ pySilver
 Rodney Richardson
 Silvano Cerza
 Stéphane Raimbault
+Jun Zhou
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### added
 * added `select_related` in intospect view for better query performance
 
+### Fixed
+* #847: Fix inappropriate message when response from authentication server is not OK.
 
 ## [1.3.2] 2020-03-24
 
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -1,5 +1,6 @@
 import base64
 import binascii
+import http.client
 import logging
 from collections import OrderedDict
 from datetime import datetime, timedelta
@@ -302,6 +303,14 @@ class OAuth2Validator(RequestValidator):
             )
         except requests.exceptions.RequestException:
             log.exception("Introspection: Failed POST to %r in token lookup", introspection_url)
+            return None
+
+        # Log an exception when response from auth server is not successful
+        if response.status_code != http.client.OK:
+            log.exception("Introspection: Failed to get a valid response "
+                          "from authentication server. status code ={}, "
+                          "reason {}".format(response.status_code,
+                                             response.reason))
             return None
 
         try:
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -308,9 +308,9 @@ class OAuth2Validator(RequestValidator):
         # Log an exception when response from auth server is not successful
         if response.status_code != http.client.OK:
             log.exception("Introspection: Failed to get a valid response "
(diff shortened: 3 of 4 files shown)


Chouvic:
> can you please add a test for the proposed change?
Sure thanks, do you know how to set up the testing environment for this? 
I've set up a virtual environment and installed Django, oauthlib. It asked for settings OAUTH2_PROVIDER. Do I need to make django-oauth-toolkit as a Django project first?

Chouvic:
A unit test is added to check the expected logger information.

auvipy:
please update the changelog and document if needed.

Chouvic:
> please update the changelog and document if needed.

A CHANGLOG is added. I have checked the documentation, currently, it seems there is no section to cover things related to APIs with the authentication server.

PR #852, opened 2020-07-13
Update install.rst

ericwood8:
Added file location of `INSTALLED_APPS`


n2ygk (changes requested):
Instead of referencing `settings.py` how about you link over to https://docs.djangoproject.com/en/dev/ref/settings/#installed-apps where it is documented? 

Note that I suggested the `dev` version since the only other versions of the documentation are pinned to a specific version number which will change over time.

WoLpH (approved):

n2ygk (approved):
Good enough.

n2ygk:
Going to have to close this since gh actions were added after this was submitted and they are blocking merging it.

Issue #871, opened 2020-09-17
Make Access Tokens Model Swappable


SpencerPinegar:
It is impossible to use a custom Model for OAUTH2_PROVIDER_ACCESS_TOKEN_MODEL (or any model other than Application) even though in the documentation there is a [setting to customize the model](https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.get_refresh_token_model) and a [custom model access method](https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.get_refresh_token_model); this is extremely misleading and has led to 2 different feature requests for DOT about this issue in the last year  (#487 #634), along with some stack-overflow issues; [example](https://stackoverflow.com/questions/51103615/custom-swappable-models-for-django-oauth-toolkit).

This problem is caused because of circular dependencies on foreign keys not being handled correctly in the migration. This makes defining and using custom models for Access/Refresh Tokens impossible out of the box - but the documentation makes it seem like an included feature.

I suggest the project uses the same API Django uses to make the User model swappable - [django-swappable-models](https://github.com/wq/django-swappable-models). This would remove the circular dependencies and should make customizing any Model (Application, Refresh Token, Access Token, Grant, etc) work regardless of how migrations are run out of the box just like Django's User Model - the only thing the programmer would have to do is specify the model in settings.
To do this:
1. update Meta attribute swappable for all Abstract Models to use swapper.swappable_setting API method 
2. update ForeignKey dependencies for all Abstract Models to use swapper.get_model_name API method
3. update get_***_model method to use swapper.load_model API method
4. update migration scripts swappable_dependency model to user swapper.dependency API method
5. update migration scripts swappable option to use swapper.swappable_setting API method

The current implementation only works for AbstractApplication Model, is dependent on the ordering of the migrations, and requires customization of the migration file after it is run to ensure it works; All of these are undesirable.
The work arounds in #487, #634 require custom migration rewrites, are still order dependent, and have more boilerplate code; this is also undesirable.

This feature will make the documentation reflect the project more accurately and simplify using and customizing this app in the future. I am going to make a PR for this issue, let me know about what you think!

SpencerPinegar:
Here is my PR - https://github.com/jazzband/django-oauth-toolkit/pull/872 - looking for some help to get this over the line and merged in so this migration issue can be resolved permanently.

n2ygk:
[...]

PR #872, opened 2020-09-17
update Meta attribute swappable for all Abstract Models to use swappe…

SpencerPinegar:
Finished with Enhancement/BugFix #871

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,5 @@
 Django>=3.0,<3.1
 oauthlib>=3.1.0
 m2r>=0.2.1
+swapper>-1.1.2.post1
 .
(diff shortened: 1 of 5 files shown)


SpencerPinegar:
Wondering how I can test this effectively because it is just a migration/model change that won't modify functionality except for allowing base classes.

n2ygk:
> Wondering how I can test this effectively because it is just a migration/model change that won't modify functionality except for allowing base classes.

Perhaps you can extend a model with some added fields and then check for the existence of them in your test case?

Also, please make sure this PR follows the [contributing guidelines](https://django-oauth-toolkit.readthedocs.io/en/latest/contributing.html) in terms of updating CHANGELOG to describe the change, make sure your name is in AUTHORS, you've got a unit test, etc.

PR #882, opened 2020-10-10
Updated supported Django versions

smithdc1:
## Checklist

- [X] unit-test added
- [X] documentation updated

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,11 +2,11 @@
 envlist =
 	py37-flake8,
 	py37-docs,
-	py38-django{30,22,21},
-	py37-django{30,22,21},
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -23,6 +23,7 @@ deps =
 	django21: Django>=2.1,<2.2
```
adamchainz, line 23: you can remove this line since it's unused
```suggestion
```

adamchainz (approved):

tox.ini
smithdc1: Thanks!

auvipy (approved):

PR #859, opened 2020-08-28
Openid Connect Core support - Round 2

thinkwelltwd:
I would like to use DOT, but need oidc support so I rebased [shaun's](https://github.com/shauns/django-oauth-toolkit/tree/oidc-in-action) PR onto [wiliamsouza's](https://github.com/wiliamsouza/django-oauth-toolkit/tree/openid-connect) PR and then rebased that onto master.

If it's not kosher to open another PR with other people's work, please let me know and I'll delete it. Otherwise, I would be very glad to get oidc support in this library.

PR #887, opened 2020-10-12
Use black for formatting the code

tevansuk:
## Use black to format the code

* Add black configuration
* Add flake8-black to check black formatting during CI
* Update isort configuration to be black compatible
* Add pre-commit to run black, isort, flake8, AST checks on commit
* Add .editorconfig to fix inconsistent indentation in tox.ini/setup.cfg

I know this can be a little subjective; some people aren't keen on `black` formatting, and it's tedious to review..

I really did this because I'm looking at the failed merge of OpenID Connect support, and a lot of the changes in #859 appear to not be related to the OIDC support, but seem to be black autoformatting, so this will be useful for me to pull out the actual diff from that PR, even if this isn't accepted.

One bone of contention might be on the imports. Previously, isort was set to a lower line length limit than flake8 was; black cannot understand this, so the line-length has to be the same in black/flake8/isort. This has altered some of the import lines a bit. 

## Checklist

- [x] unit-test added
- [x] documentation updated

## Current line length histogram, for information

```
~/bin/line-length-hist.awk $(find oauth2_provider tests -name '*.py' ! -path '*/migrations/*')

  0-4     ************************************************************************************ 1680
  5-9     *********************** 443
 10-14    ************************** 520
 15-19    ******** 145
 20-24    ***************** 337
 25-29    ******************** 391
 30-34    ****************************** 585
 35-39    ******************************* 616
 40-44    ***************************** 562
 45-49    **************************** 547
 50-54    ************************************ 706
 55-59    ********************* 410
 60-64    ******************** 387
 65-69    ********************* 403
 70-74    ************** 273
 75-79    *********** 219
 80-84    ******** 149
 85-89    ******* 138
 90-94    ******* 136
 95-99    **** 63
100-104   *** 56
105-109   * 20
110-114   *** 52
```

```awk
#!/usr/bin/awk -f

function ceil(valor) { return (valor == int(valor)) ? valor : int(valor)+1 }

{ ++buckets[int(length()/5)] }
END {
    for (i in buckets) {
        b = i * 5
        printf "%3d-%-3d   ", b, b+4
        print gensub(/ /, "*", "g", sprintf("%*s", ceil(buckets[i]/20), "")), buckets[i] 
    }
}
```


MattBlack85:
@tevansuk personally I am a big fan of black, I don't mean I like all the styles it applies but the fact that I don't have to worry about code formatting is a big thing to me. Will mark this as a potential 1.4.0 patch waiting for others to chime in a say what they think about it

smithdc1:
I am +1 for black, I also like the addition of a pre-commit hook. I find this very helpful. A couple of small comments. 

I see you've added flake8-black. What are the benefits of this over the inbuilt check functionality of black? 
What was the reason behind 110 line length, if went with default could this remove the need for the `pyproject.toml` file?

tevansuk:
> I am +1 for black, I also like the addition of a pre-commit hook. I find this very helpful. A couple of small comments.

Thanks for the review :)

> 
> I see you've added flake8-black. What are the benefits of this over the inbuilt check functionality of black?

`flake8-black` integrates `black` linting when `flake8` is run during CI, so if a developer has not installed the recommended `pre-commit` hooks, and has allowed some code that violates `black` to be committed, it would get flagged during the current CI checks. It allows us to integrate `black` checking without adding a new CI step.

> What was the reason behind 110 line length, if went with default could this remove the need for the `pyproject.toml` file?

This just matches the current maximum line length specified in `flake8`'s configuration. If we kept the default line length then we wouldn't need `pyproject.toml`, but the diff would be significantly larger - at least 175 new changes. Line length changes are always contentious - its a "bike shed problem", everyone can have an opinion on what colour it should be painted :)

On my own projects I always change the line length because the default number in `black` is (accidentally) the same as a racist hate symbol.

PS: I probably won't keep chasing `HEAD` on this, but ping me any time you'd like this PR updated :+1:

MattBlack85:
@tevansuk I think this is good to go, do you mind updating your branch? I am gonna merge it after that

tevansuk:
> @tevansuk I think this is good to go, do you mind updating your branch? I am gonna merge it after that

Sounds good :+1: - I'll start updating it now

MattBlack85 (approved):

MattBlack85 (approved):

PR #893, opened 2020-10-22
searching for active access tokens moved to separate method

zzzevaka:
Hello!

I want to automatically grant access if access tokens exist for this app and user, even they expired.

Now I have to redefine get method entirely. 

I moved searching active access token to separate method for having an opportunity to redefine it without touching other logic.

Thanks!

### DIFF
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -136,6 +136,18 @@ class AuthorizationView(BaseAuthorizationView, FormView):
         log.debug("Success url for the request: {0}".format(self.success_url))
         return self.redirect(self.success_url, application)
 
+    def get_active_access_token(self, application, user, scopes):
+        tokens = get_access_token_model().objects.filter(
+            user=user,
+            application=application,
+            expires__gt=timezone.now()
+        ).all()
+
+        # check past authorizations regarded the same scopes as the current one
+        for token in tokens:
+            if token.allow_scopes(scopes):
+                return token
+
     def get(self, request, *args, **kwargs):
         try:
             scopes, credentials = self.validate_authorization_request(request)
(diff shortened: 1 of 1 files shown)


n2ygk:
Wouldn't the correct approach to your requirement be to use Refresh Tokens?  Using an expired Access Token is a violation of the OAuth 2.0 protocol so I don't see the point of this change.

zzzevaka:
Thanks for the answer!

I'm not going to use expired token. I want to create new one. Refresh token can expire too. I want to automatically grant access if user did it once. No matter how long ago it was.

oauth2_provider/views/base.py
```diff
@@ -136,6 +136,18 @@ def form_valid(self, form):
         log.debug("Success url for the request: {0}".format(self.success_url))
         return self.redirect(self.success_url, application)
 
+    def get_active_access_token(self, application, user, scopes):
+        tokens = get_access_token_model().objects.filter(
+            user=user,
+            application=application,
+            expires__gt=timezone.now()
+        ).all()
+
```
MattBlack85, line 145: .all() is redundant here,since you're changing this,can you drop it?

MattBlack85:
@zzzevaka changes look good, but we need some tests so we can be covered if somebody is gonna pome with it in the future

oauth2_provider/views/base.py
zzzevaka: done

FluidSense:
@zzzevaka Just for clarification: Do I understand you correctly that you want see if the user has consented to sharing the data earlier, by comparing previous access tokens' scopes for this user against the currently requested access token's scopes? 
Which for example can be used to issue a new access token without asking for consent?

FluidSense:
@MattBlack85 I see. However, wouldn't a discussion of the actual use case for overrides be relevant, to ensure that the PR do not open for protocol & security violations?

MattBlack85:
@FluidSense honestly? I don't believe so. I found myself having to maintain my own fork for a couple of months because we had custom models that were in need for custom attribute updates and we had to declare entire DOT modules into our project to do so. Being extensible should be the number one priority for a library that declares to be extendible with custom models, and this change reflects this exact thing. Let's think at django for example, would you think opening a discussion for making the User model extendible and totally swappable would be needed? I still think no, the libraries provide the rails and can guide us among best practices, but if one wants to change the default behavior, at that point is up to that person to think and know if the custom change would be a good or bad thing

MattBlack85:
similar case https://github.com/jazz band/django-oauth-toolkit/pull/695

MattBlack85:
@zzzevaka do you plan to add some tests to your PR?

zzzevaka:
> @zzzevaka do you plan to add some tests to your PR?

Yes, I will

MattBlack85:
@zzzevaka we want to do a new release early next week, you think you can finish this PR so we can include it?

n2ygk:
@zzzevaka @MattBlack85 I guess this is gonna miss 1.3.4 release which is already delayed several months since tests have not been added.

MattBlack85:
agree @n2ygk

n2ygk:
Closing this due to lack of follow-up.

PR #896, opened 2020-11-11
pass PKCE fields to AuthorizationView form

skarzi:
Hi :wave: 

first of all thanks for developing and maintaining a great project!

## Description of the Change

Pass `code_challenge` and `code_challenge_method` from query string to `AuthorizationView` form in `get()`.
Without this, it was impossible to use authorization code grant flow with `GET`, because `code_challenge` and `code_challenge_method` data were never passed to form, so they weren't in `form.cleaned_data`, which causes creating `Grant` with always empty `code_challenge` and `code_challenge_method`.

This issue was quite hard bug to discover because there are already few tests for authorization code flow pkce, however, they weren't checking form rendering in `GET` request, but only `response.status_code`, I have added asserts for these 2 values, please look at the changes in `test_public_pkce_plain_authorize_get` and `test_public_pkce_S256_authorize_get` tests in `test_authorization_code.py`.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,3 +31,4 @@ Silvano Cerza
 Stéphane Raimbault
 Jun Zhou
 David Smith
+Łukasz Skarżyński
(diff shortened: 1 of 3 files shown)


auvipy (approved):

MattBlack85:
looks good @skarzi , checked the tests locally and they're ok, thanks!

MattBlack85 (approved):

PR #903, opened 2020-12-11
Disable 255 chars length limit for redirect uri (#902)

### LINKED ISSUE #902: A limit on redirect_uri only up to 255 chars
**Describe the bug**
Cannot handle `redirect_uri` longer then 255 chars. [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.1.1) recommends to design system to be capable to work with URI at least to 8000 chars long.

> It is RECOMMENDED that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

**To Reproduce**
Send a request to `AuthorizationView` with `redirect_uri` longer than 255 chars.

**Expected behavior**
To save `redirect_uri` into `Grant` model.

**Version**
1.3.3

shaddeus:
Fixes #902

## Description of the Change
Changed model field `AbstractGrant.redirect_uri` (from `CharField(max_length=255)`) to `TextField()`.

Added a pytest for `AbstractGrant` model. Unfortunately it works effective only with database engines that limits strings with `VARCHAR(255)` data type) - not SQLite. [See documentation](https://sqlite.org/datatype3.html#affinity_name_examples).

Creates `BaseTestModels` class with a testing user. Uses `TestModels`, `TestCustomModels`, `TestGrantModel`, `TestAccessTokenModel`, `TestRefreshTokenModel` and `TestClearExpired`.

## Checklist

- [x] unit-test added
- [ ] documentation updated


oauth2_provider/migrations/0003_auto_20201211_1314.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('oauth2_provider', '0002_auto_20190406_1805'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='grant',
```
MattBlack85, line 14: I need to read about swappable models and migrations, not sure how the magic works for migrations

shaddeus: That is a good point. I didn't realize it.

I found very similar example in Django code. Swappable `User` model:

In [contrib/auth/migrations/0001_initial.py](https://github.com/django/django/blob/master/django/contrib/auth/migrations/0001_initial.py) is initialized User model as swappable.

In [contrib/auth/migrations/0003_alter_user_email_max_length.py](https://github.com/django/django/blob/master/django/contrib/auth/migrations/0003_alter_user_email_max_length.py) is updated `max_length` for `email` field and looks like normal migration.

MattBlack85: cool!

MattBlack85 (approved):

MattBlack85:
@shaddeus we need an entry for the CHANGELOG

shaddeus:
@MattBlack85 Yes, I'll add an entry into CHANGELOG.
Really?  Should I write me to a list of AUTHORS?  It would be a honor.

n2ygk:
Credit/(blame:-) to all who contribute!
It's why it's in the PR template checklist.

MattBlack85:
thanks for your contribution @shaddeus !

Issue #898, opened 2020-11-18
Registration custom class for  in django admin


folt:
I am getting an issue with admin page extension. When I do 
`admin.site.register (AccessToken, AccessTokenAdmin)`
where `AccessTokenAdmin` is my class, then I get the error "The model AccessToken is already registered".

I propose a solution when you can specify a class file for the django administration panel:
`
ACCESS_TOKEN_ADMIN = getattr (settings, "OAUTH2_PROVIDER_ACCESS_TOKEN_ADMIN", "custom.AccessToken")
`

Currently, we have to solve this problem as follows:
```
if not admin.site.is_registered (AccessToken):
     admin.site.register (AccessToken, AccessTokenAdmin)
```
I'm willing to contribute if the community supports it.

dopry:
fixed in #904

PR #904, opened 2020-12-11
#898 Added the ability to customize classes for django admin

folt:
Fixes #

## Description of the Change
#898 Added the ability to customize classes for django admin

## Checklist

- [x] unit-test added
- [x] documentation updated


MattBlack85:
@folt the PR looks good to me but I can't see any test for a custom admin class, could you add one for that case also?

oauth2_provider/settings.py
```diff
@@ -128,13 +134,13 @@ class OAuth2ProviderSettings:
 
     def __init__(self, user_settings=None, defaults=None, import_strings=None, mandatory=None):
         self.user_settings = user_settings or {}
-        self.defaults = defaults or {}
-        self.import_strings = import_strings or ()
+        self.defaults = defaults or DEFAULTS
```
MattBlack85, line 137: why you decided to pass DEFAULTS if we're not passing anything? This looks legit to me, tho I am wondering why it was {} and not DEFAULTS before

folt: I noticed that the library settings are very similar to the django-rest-framework settings.
My changes are inspired from [there](https://github.com/encode/django-rest-framework/blob/master/rest_framework/settings.py#L202)

This will allow [tests](https://github.com/jazzband/django-oauth-toolkit/pull/904/files#diff-f20069a79c955c52519f37001105df187ba738b9a4dd49af010c723ff810200cR17) to cover configuration errors. Previously, there were no tests for settings at all.

folt:
> @folt the PR looks good to me but I can't see any test for a custom admin class, could you add one for that case also?

A good idea. I will add PR for this case

oauth2_provider/settings.py
MattBlack85: makes sense, thanks!

MattBlack85 (approved):

PR #906, opened 2020-12-16
Fix select for update

### LINKED ISSUE #866: LIMIT/OFFSET is not supported with select_for_update on this database backend.
We are using Oracle Database 12c Enterprise Edition Release 12.2.0.1.0, which from what I researched it seems the issue is coming from [this statement](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L406-L409) since 12c doesn't allow limits in ```select_for_update``` statements. We were able to fix it with:
```python
            try:
                self = refresh_token_model.select_for_update().get(
                    pk=self.pk, revoked__isnull=True
                )
            except refresh_token_model.DoesNotExist:
                return
```

It would be great if we could stop maintaining the fork and just use the release version.

dag18e:
Fixes #866 
## Description of the Change
Oracle Database 12c does not allow limits on ```select_for_update``` statements. This change removes the limit and instead just uses a ```get```.

## Checklist
- [ ] unit-test added
- [ ] documentation updated


MattBlack85:
@dag18e I understand your point but your change is gonna drop the lock on the database row and it's completely changing the intended behaviour

MattBlack85:
@dag18e :facepalm: I was reviewing this on my mobile and I didn't notice how it really was. Ignore my first comment.
I have another question tho, I suppose .filter() was used cause we may have more than one token, if we use .get() and there is more than one token the exception is not caught?
Maybe a simple slicing could be used instead?

dag18e:
@MattBlack85 Hey man, sorry about the delay here. Had some other stuff come up + holidays so I'm just getting back to this. On your note about filter vs get, I changed it a bit to use filter. Basically, I wrap it in a list and grab the first in the list.

This ought to retain the original functionality, but it also doesn't apply a limit to the request so it works with Oracle 12c.

I'll be adding info to docs and changelog once tests pass Travis.

dag18e:
@MattBlack85 Actually, do you think I need to add anything to the documentation? I don't really feel this change is worthy of a change there, but you'd know better than me.

auvipy (approved):

n2ygk (approved):
@MattBlack85 I think this is good with the change to using a transaction.

Thanks @dag18e

oauth2_provider/models.py
```diff
@@ -401,13 +401,15 @@ def revoke(self):
...
-                refresh_token_model.objects.filter(pk=self.pk, revoked__isnull=True)
-                .select_for_update()
-                .first()
-            )
-            if not self:
+            try:
+                token = refresh_token_model.objects.select_for_update().filter(
+                    pk=self.pk, revoked__isnull=True
+                )
+            except refresh_token_model.DoesNotExist:
```
MattBlack85, line 408: @dag18e @n2ygk I think we added some things here that are not really needed:
1) AFAIK .filter() can't raise DoesNotExist
2) L412 cast the Queryset to a list and then slicing, I do no think this is needed and this slows down things a bit

I feel these things should be addressed in a follow up PR

PR #908, opened 2020-12-21
Updated Fury badge to high quality SVG icon

WoLpH:
## Description of the Change

Updated the fury badge to use svg like all other badges


jezdez:
Woops, I just fixed that directly. Sorry!

Issue #873, opened 2020-10-01
'OIDC_RSA_PRIVATE_KEY' is mandatory


n2ygk:
**Describe the bug**

Existing code using the latest master commit (pr #859) throws an error when not using OIDC features.

**To Reproduce**

Install DOT master commit 295c065a3e5a14b777f1e56c8e36adedae0af461

**Expected behavior**

Code not using OIDC should continue to work.

**Version**

master 295c065a3e5a14b777f1e56c8e36adedae0af461

- [ ] I have tested with the latest published release and it's still a problem.
- [x] I have tested with the master branch and it's still a problem.

**Additional context**

```
  File "/Users/alan/src/django-training/env/lib/python3.7/site-packages/oauth2_provider/settings.py", line 217, in validate_setting
    raise AttributeError("OAuth2Provider setting: %r is mandatory" % (attr))
rest_framework.request.WrappedAttributeError: OAuth2Provider setting: 'OIDC_RSA_PRIVATE_KEY' is mandatory
```

FluidSense:
I'm a bit interested in this, as we are currently using [django-oidc-provider](https://github.com/juanifioren/django-oidc-provider), but it has gone stale. It would be nice to migrate over to django-oauth-toolkit, but we would then need OIDC-support in django-oauth-toolkit. 
Could a solution to this be implementing RSA private key as in django-oidc-provider, e.g., adding a command for generating it, storing it in a model and accessing it through there? Then it could be moved out of settings and removed from `MANDATORY_SETTINGS` which seems to be the source of this error message.

Perhaps @thinkwelltwd can contribute with some insight into the decisions on having it in settings?

n2ygk:
@FluidSense We'v have had to revert #859 until it doesn't break existing non-OIDC usage. Probably just a few configuration tweaks needed here and there. It also needs to be documented and clearly some more test cases are needed. The plan is to make a small 1.3.3 release without the OIDC feature and then hopefully get the contributors to finish up the PR such that it is mergeable into a new major release. It sill needs a fair amount of work to make that happen.

Help making this happen would be appreciated!

FluidSense:
I see, that sounds like a good decision seeing as there has been no recent releases, but some fixes. 

Perhaps a place to start to look into things would be the reverted PR then, to see if I have something to contribute there, and if I can understand enough of it to document on behalf of the original authors

n2ygk:
Reverted in #877

PR #915, opened 2021-01-12
OpenID Connect support

### LINKED ISSUE #110: OpenID Connect

tevansuk:
Fixes #110 

## Description of the Change

*DRAFT* 

Adds support for OIDC authorization code flow and hybrid flow. This is mostly not my work; I've just rebased #859 onto the latest master and added support for swappable `ModelAdmins` for the new models. It's quite a large change, so I've put it up for review prior to working on finishing off the actual PR (see checklist).

Thanks to Dave Burkholder, William Souza, Allisson Azevedo, fvlima and Shaun Stanworth for their initial work on rounds 1 and 2 of OIDC support :+1:

## Checklist

Significant changes that still need to be made:

- [x] unit-test added
- [x] don't break existing non OIDC setups (#873)
- [x] documentation on enabling/customising OIDC is missing
- [x] don't double produce id token - save it in the authorization endpoint and retrieve it in token endpoint
- [x] helper functions for checking when `at_hash` and `c_hash` are required
- [x] support `id_token_hint` (`RequestValidator.validate_user_hint`)
- [x] support authorization code nonce (`RequestValidator.get_authorization_code_nonce`)


tevansuk:
There's still more to come :) 

I'm working this week on adding more tests + documentation, and then on the TODO items on the checklist

oauth2_provider/views/oidc.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import absolute_import, unicode_literals
```
auvipy, line 1: i think this is not needed in python 3 only support

oauth2_provider/views/mixins.py
```diff
@@ -282,11 +289,38 @@ def dispatch(self, request, *args, **kwargs):
...
-                return super().dispatch(request, *args, **kwargs)
-            else:
-                return HttpResponseForbidden()
+            try:
+                valid, r = self.verify_request(request)
+                if valid:
+                    request.resource_owner = r.user
+                    return super().dispatch(request, *args, **kwargs)
+            except ValueError:
+                pass
```
tevansuk, line 298: This code is from the original PR; it looks wrong to me as it will squash and hide `ValueError` raised by the view we are mixing in to, which should be the responsibility of the view itself, if it wants to hide those errors.

Are we trying to squash a potential `ValueError` from `self.verify_request()`? This would make more sense, but then the `try / except` should be just over that line.

oauth2_provider/oauth2_validators.py
```diff
@@ -665,3 +711,181 @@ def validate_refresh_token(self, refresh_token, client, request, *args, **kwargs
...
         request.refresh_token_instance = rt
         return rt.application == client
+
+    @transaction.atomic
+    def _save_id_token(self, token, request, expires, *args, **kwargs):
+
+        scopes = request.scope or " ".join(request.scopes)
+
+        if request.grant_type == "client_credentials":
+            request.user = None
```
tevansuk, line 721: I believe OIDC `client_credentials` flow doesn't support ID tokens, so I'm not sure how you would get here?

oauth2_provider/urls.py
```diff
@@ -30,5 +30,15 @@
     ),
 ]
 
+oidc_urlpatterns = [
+    re_path(
+        r"^\.well-known/openid-configuration/$",
+        views.ConnectDiscoveryInfoView.as_view(),
+        name="oidc-connect-discovery-info",
+    ),
+    re_path(r"^jwks/$", views.JwksInfoView.as_view(), name="jwks-info"),
```
tevansuk, line 39: I can't find an RFC where the `jwks` path is defined, and its not defined at [IANA](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml#well-known-uris-1). 

All well behaved OIDC clients should discover it from `jwks_uri` from `.well-known/openid-configuration` of course, but it seems *convention* has it `.well-known/jwks.json` ([Auth0](https://auth0.com/docs/tokens/json-web-tokens/json-web-key-sets), [AWS](https://aws.amazon.com/premiumsupport/knowledge-center/decode-verify-cognito-json-token/), [Akamai](https://blogs.akamai.com/2019/10/verify-jwt-with-json-web-key-set-jwks-in-api-gateway.html) etc). Okta *don't*, so it's clearly not a huge deal.

Should I update to `.well-known/jwks.json`?

oauth2_provider/settings.py
```diff
@@ -46,20 +48,41 @@
...
+        "code",
+        "token",
+        "id_token",
+        "id_token token",
+        "code token",
+        "code id_token",
+        "code id_token token",
+    ],
+    "OIDC_SUBJECT_TYPES_SUPPORTED": ["public"],
+    "OIDC_ID_TOKEN_SIGNING_ALG_VALUES_SUPPORTED": ["RS256", "HS256"],
```
tevansuk, line 81: Do we really support `HS256`? We don't have anything that allows you to configure the secret for it, ID tokens are signed using RSA keys - to me that says we only support `RD256`.

oauth2_provider/views/oidc.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    View used to show oidc provider configuration information
+    """
+
+    def get(self, request, *args, **kwargs):
+        issuer_url = oauth2_settings.OIDC_ISS_ENDPOINT
+
+        if not issuer_url:
+            abs_url = request.build_absolute_uri(reverse("oauth2_provider:oidc-connect-discovery-info"))
+            issuer_url = abs_url[: -len("/.well-known/openid-configuration/")]
```
tevansuk, line 24: This is duplicated code from `OAuth2Validator.get_oidc_issuer_endpoint`. Refactor out to a util function.

oauth2_provider/oauth2_validators.py
```diff
@@ -461,6 +485,24 @@ def get_code_challenge_method(self, code, request):
...
+        scopes = []
+        fields = {
+            "code": code,
+        }
+
+        if client_id:
+            fields["application__client_id"] = client_id
+
+        if redirect_uri:
+            fields["redirect_uri"] = redirect_uri
```
tevansuk, line 498: I don't see why we are filtering by these values here. `oauthlib` says:

```
        With that the token endpoint knows when to include OpenIDConnect
        id_token in token response only based on authorization code scopes.

        Only code param should be sufficient to retrieve grant code from
        any storage you are using, `client_id` and `redirect_uri` can have a
        blank value `""` don't forget to check it before using those values
        in a select query if a database is used.
```

This function is only used to allow `oauthlib` dispatcher to work out whether to use an `openid` dispatcher or fallback to `oauth2` - code should be sufficient, the other fields are validated later.

oauth2_provider/oauth2_validators.py
```diff
@@ -461,6 +485,24 @@ def get_code_challenge_method(self, code, request):
...
+
+        grant = Grant.objects.filter(**fields).values()
+        if grant.exists():
+            grant_dict = dict(grant[0])
+            scopes = utils.scope_to_list(grant_dict["scope"])
```
tevansuk, line 503: No need for two DB queries here.

```python
scope = Grant.objects.filter(code=code).values_list("scope", flat=True).first()
if scope:
    return utils.scope_to_list(scope)
```

oauth2_provider/oauth2_validators.py
```diff
@@ -665,3 +701,174 @@ def validate_refresh_token(self, refresh_token, client, request, *args, **kwargs
...
+        """
+        if isinstance(audience, str):
+            audience = [audience]
+        return Application.objects.filter(client_id__in=audience).first()
+
+    def validate_user_match(self, id_token_hint, scopes, claims, request):
+        # TODO: Fix to validate when necessary acording
+        # https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/request_validator.py#L556
+        # http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest id_token_hint section
+        return True
```
tevansuk, line 841: This will have to be implemented later. There is an issue with oauthlib's implementation here. It says:

```
     def validate_user_match(self, id_token_hint, scopes, claims, request):
        """Ensure client supplied user id hint matches session user.

        If the sub claim or id_token_hint is supplied then the session
        user must match the given ID.
```

However, this method is called during the `validate_authorization_request` phase, which gives no opportunity to inject the current session user into the oauthlib `Request`. This means we have no way of comparing the hint to the current session user. :shrug: 

I have a branch with an implementation of this method, but as it stands it would also need some changes to oauthlib.

tevansuk:
This is now complete. There are a couple of outstanding issues (all mentioned in review comments above):

* `validate_user_match` seems impossible to implement with current oauthlib
* in hybrid flow, saving the id token in auth endpoint and returning the exact same id token from token endpoint - I don't think its necessary
* error handling in `OAuthLibMixin`
* special casing `client_credentials` flow in id token generation - you can't get an id token with client credentials, so :shrug:

Cheers

Tom

n2ygk:
> This is now complete. There are a couple of outstanding issues (all mentioned in review comments above):
Thanks, Tom.  This is an important contribution. I'll start my review soon.

n2ygk:
@MattBlack85 Do we make this 1.5.0 or 2.0 release? It's a pretty big deal but not sure if there are breaking changes.

MattBlack85:
@n2ygk you have the necessary knowledge here to evaluate it (have no idea overall about openid), I just read from you this is not adding any breaking changes, so to me it's a 1.5.0 :)

n2ygk:
> @n2ygk you have the necessary knowledge here to evaluate it (have no idea overall about openid), I just read from you this is not adding any breaking changes, so to me it's a 1.5.0 :)

I'll try to get this review done this week. It's pretty large... 51 files!

MattBlack85:
> > @n2ygk you have the necessary knowledge here to evaluate it (have no idea overall about openid), I just read from you this is not adding any breaking changes, so to me it's a 1.5.0 :)
> 
> I'll try to get this review done this week. It's pretty large... 51 files!

if you want, I can try to be a rubber duck :duck:

oauth2_provider/models.py
```diff
@@ -433,6 +485,105 @@ class Meta(AbstractRefreshToken.Meta):
...
+
+    id = models.BigAutoField(primary_key=True)
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        blank=True,
+        null=True,
+        related_name="%(app_label)s_%(class)s",
+    )
+    token = models.TextField(unique=True)
```
tevansuk, line 510: This is a bit of a pain - only noticed this as we started to move our own app in to QA with Google Cloud SQL - mysql doesn't support indexes on TextField, it only allows a partial index of the text. Its my fault, I only tested against postgres, sqlite and mariadb in development - mariadb supports indexing `TEXT` fields by hashing the text and indexing that instead.

Even if we decided to add a partial index for mysql, its a real pain to do so with django - you have to add it as a RunSQL migration and tell it that its a partial index. I'm investigating to see whether there is a way to do this operation just on mysql DBs. It would also mean that there is a possibility that the uniqueness constraint would be violated if just the first 255 characters are the same (the limit on a UTF-8 string index in mysql 5.7/5.8). Probably would have to be a non-unique index in that case.

If we can't, we'd have to ship instructions on how to swap the model so that mysql users can create a migration themselves that will work and give instructions for that.

A final, more drastic action would be to not store the ID tokens at all. Currently, we only retrieve ID tokens to validate that an ID token that we've received is valid. It is possible instead to validate them by verifying the JWT token signature, and checking `iss`, `nbf`, and `exp`. Of course, this makes it impossible to revoke an ID token when revoking an access token.

tevansuk: As an example of how horrible this is, this is how we worked around it for now on our QA site:

```python


# Extend the IDToken model so that we can manually tweak the migration

class IDToken(AbstractIDToken):
    """
    We have to override this model because mysql cannot handle indexes on a
    TextField, and the standard migrations will break when run on mysql.

    By overriding the field and marking it as swapped for this modedl, Django
    won't attempt to run the migration for the base model and we can use our
    hand crafted migration to create the table as best we can.
    """

    pass


# Create a custom DB migration operation to do things differently on mysql

class MysqlOnlyRunSQL(migrations.RunSQL):
    """
    Mysql cannot create indexes on a whole text field, and django will not auto
    restrain the index to what mysql does support.

    We can hack this in a way that adds what mysql can support with a RunSQL
    migration, but that will fail on non mysql databases. This class wraps RunSQL
    and if the migration is run on a non mysql database, it will instead run the
    operations specified in the state_operations.

    This allows django to think that the desired state has been achieved on both
    mysql and non-mysql databases.
    """

    @property
    def reversible(self):
        """
        Can only be reversed if both the RunSQL and regular operations can be reversed
        """
        return self.reverse_sql is not None and all(
            op.reversible for op in self.state_operations
        )

    def database_forwards(self, app_label, schema_editor, from_state, to_state):
        """
        If we are using a mysql schema editor, run the supplied SQL.
        Otherwise, apply the operations in state_operations
        """
        if "mysql" in schema_editor.__class__.__module__:
            super().database_forwards(app_label, schema_editor, from_state, to_state)
        else:
            for op in self.state_operations:
                op.database_forwards(app_label, schema_editor, from_state, to_state)

    def database_backwards(self, app_label, schema_editor, from_state, to_state):
        """
        If we are using a mysql schema editor, run the supplied SQL.
        Otherwise, apply the operations in state_operations
        """
        if "mysql" in schema_editor.__class__.__module__:
            super().database_backwards(app_label, schema_editor, from_state, to_state)
        else:
            for op in self.state_operations:
                op.database_backwards(app_label, schema_editor, from_state, to_state)


# Handcraft the initial migration

        migrations.CreateModel(
            name="IDToken",
            fields=[
                ("id", models.BigAutoField(primary_key=True, serialize=False)),
                ("token", models.TextField()),
                ("expires", models.DateTimeField()),
                ("scope", models.TextField(blank=True)),
                ("created", models.DateTimeField(auto_now_add=True)),
                ("updated", models.DateTimeField(auto_now=True)),
                (
                    "application",
                    models.ForeignKey(
                        blank=True,
                        null=True,
                        on_delete=django.db.models.deletion.CASCADE,
                        to=oauth2_settings.APPLICATION_MODEL,
                    ),
                ),
                (
                    "user",
                    models.ForeignKey(
                        blank=True,
                        null=True,
                        on_delete=django.db.models.deletion.CASCADE,
                        related_name="pots_oauth_idtoken",
                        to=settings.AUTH_USER_MODEL,
                    ),
                ),
            ],
            options={
                "abstract": False,
            },
        ),
        # Hack to add partial indexes only on mysql, and full indexes on other DBs
        MysqlOnlyRunSQL(
            "CREATE INDEX `token_idx` ON `pots_oauth_idtoken` (`token`(255))",
            reverse_sql="ALTER TABLE `pots_oauth_idtoken` DROP INDEX `token_idx`",
            state_operations=[
                migrations.AlterField(
                    model_name="IDToken",
                    name="token",
                    field=models.TextField(unique=True),
                ),
            ],
        ),
```

Big eurgh!

n2ygk: @tevansuk Have you considered splitting the JWT into its 3 component parts in the model to enable the uniqueness constraint? The SHA-256 is guaranteed to be 64 bytes, so easily uniquely indexed. You don't really need to uniquely constrain the _entire_ JWT as the SHA-256 signature is sufficient to establish uniqueness since it is a secure hash of the header and payload.

```python
    header = models.TextField()
    payload = models.TextField()
    signature = models.CharField(unique=True, max_length=64)
```

I haven't yet read the code to see how `AbstractIDToken` is used so I may be totally off-base here.

tswfi:
@tevansuk thank you for your hard work! 

I can confirm that I was able to use this branch just now in my project to auth with keycloak. Did not read through all the documentation but found the parts needed to set everything up.

:+1:

oauth2_provider/models.py
tevansuk: This seems like a sensible approach, I'll give this a try and see what will need to be changed. Thanks Alan :+1:

tevansuk: I think using the signature will also not work - a SHA-256 signature using a 2048 bit key will be 2048 bits/256 bytes long in binary, or 342 bytes long when base64 urlencoded.

I think the best way forward will be to:

* include a `jti` (`uuid.uuid4()`) in the id token
* store the `jti` and the token separately in the DB, indexing the `jti` field
* use the `jti` as the unique key for the token

When we receive a token and want to retrieve it from the DB:

* verify the token's signature
* extract the `jti`
* look up the `IDToken` in the database using the `jti`

Whilst we could possibly index the SHA-256 signature as binary for a 2048 bit key, we might hit DB concerns again for a larger key, or if we later wanted to use a different signature algorithm. Indexing the `jti` claim would work whatever hash algorithm was used.

n2ygk: Hmm, according to [this mysql 5.7 doc](https://dev.mysql.com/doc/refman/5.7/en/innodb-limits.html): 

    The index key prefix length limit is 767 bytes for InnoDB tables that use the REDUNDANT or COMPACT row format. For 
    example, you might hit this limit with a column prefix index of more than 255 characters on a TEXT or VARCHAR 
    column, assuming a utf8mb3 character set and the maximum of 3 bytes for each character.

So that limit is for 255 **multi-byte** characters. Wouldn't this be a maximum of 256 single-byte characters we are talking about?

tevansuk: Yes, 256 binary characters, which we wouldn't want to stuff in to a django `CharField`! 

Part of the problem is Django sets its own limits when dealing with mysql; it has no idea (and no way to specify) what the content you are storing in a CharField is, and so if you are using mysql [it sets the limit](https://docs.djangoproject.com/en/2.2/ref/databases/#notes-on-specific-fields) on the maximum length of a `CharField` to be 255, and doesn't support db indexes on `TextField`. Django only provides one field for storing binary data, `BinaryField`, which on mysql maps to `LONGBLOB`; this suffers from the same issue that we have with `TEXT`, mysql requires a maximum key length for an index, and Django provides no way for us to specify this.

`jti` is designed for specifically for identifying a JWT; we have a JWT that is signed by us, we've verified the signature, `exp` and `nbf`, we now just want to know if it is still valid in the database.

I'm not even sure we should be storing the contents of the ID token. We use the ID token model for:

* Checking to see if the token exists in our database and hasn't been revoked
* Determining the application it was issued to
* Determining the scopes it was issued for

None of this requires storing the actual token itself if we are indexing by the `jti`. If access to the claims is also required; well we've received the JWT and validated it, so the claims are available directly in the JWT.

This is also desirable from a PII point of view; if we don't store the contents of the ID token, then there is no PII at all in the `IDToken` table.

Natureshadow:
This feature would be much appreciated. We are trying to get OpenID Connect provider support in AlekSIS and are currently struggling to integrate django-oidc-provider with django-oauth-toolkit. Having django-oauth-toolkit be able to do the whole job in time for our release would simplify things a lot!

Natureshadow:
We got it working in produciton, with one issue:

In `settings.oidc_issuer`, when the request is an `oauthlib.common.Request`, the scheme is not determined correctly (there simply is no way for the `django_request` object that is constructed to determine whether the site uses HTTPS or not). Thus, the issuer URL in the discovery document ends up to be correct (with https scheme), whereis the JWT will contain the issueer URL with http scheme.

We worked around that by setting Django's `SECURE_PROXY_SSL_HEADER` setting to something that forces Django's `HttpRequest.is_secure()` to return True, but that's not a solution (for instance, it will not work if used without a proxy).

oauth2_provider/models.py
n2ygk: Looks like you resolved this with d6f6082. Sorry for my delay in getting back.

oauth2_provider/oauth2_backends.py
```diff
@@ -122,7 +122,7 @@ def create_authorization_response(self, request, scopes, credentials, allow):
             credentials["user"] = request.user
 
             headers, body, status = self.server.create_authorization_response(
-                uri=credentials["redirect_uri"], scopes=scopes, credentials=credentials
+                uri=request.get_raw_uri(), scopes=scopes, credentials=credentials
```
n2ygk, line 125: I don't understand why this changes from using the supplied `redirect_uri`. Does this allow one to send in a redirect to other than the approved place(s)?

Can you explain what's happening here?

n2ygk (changes requested):
This looks really great. Just a few questions.

oauth2_provider/oauth2_backends.py
tevansuk: (I don't think anything actually uses this uri.)

The URI is passed to `oauthlib.oauth2.rfc6749.endpoints.AuthorizationEndpoint`, where it is used to construct an `oauthlib.common.Request`. I believe it is supposed to be the URI of the incoming request, and so it was an error for it to be passing the redirect uri as it was previously.

Its not documented as such in `AuthorizationEndpoint`, but if you look at the examples in `oauthlib.oauth2.rfc6749.grant_types.AuthorizationCodeGrant`, its clear that `oauthlib` is expecting the uri passed in to be the uri requesting the authorization endpoint:

https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py#L210-L235

The redirect uri is also in `credentials`, which are all set as attributes on the `oauthlib.common.Request`. 

https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/oauth2/rfc6749/endpoints/authorization.py#L89-L95

Checking the redirect uri uses this `request.redirect_uri` attribute, not the request URI. 

https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/oauth2/rfc6749/grant_types/base.py#L228-L241

tevansuk:
> We got it working in staging, with one issue:
> 
> In `settings.oidc_issuer`, when the request is an `oauthlib.common.Request`, the scheme is not determined correctly (there simply is no way for the `django_request` object that is constructed to determine whether the site uses HTTPS or not). Thus, the issuer URL in the discovery document ends up to be correct (with https scheme), whereis the JWT will contain the issueer URL with http scheme.
> 
> We worked around that by setting Django's `SECURE_PROXY_SSL_HEADER` setting to something that forces Django's `HttpRequest.is_secure()` to return True, but that's not a solution (for instance, it will not work if used without a proxy).

Ah, that's a bit of a pain. You can also fix the issuer by setting it explicitly in settings, but it would be preferable to avoid adding a setting just to get SSL working correctly. We have some control over how `oauthlib.Request` objects get created, I'll see if I can smuggle through whether the request is SSL or not and use that to create the phony django request.

n2ygk:
@tevansuk Are you still making changes to this PR? I had begun reviewing it but will wait if you aren't ready.

tevansuk:
> @tevansuk Are you still making changes to this PR? I had begun reviewing it but will wait if you aren't ready.

@n2ygk I'm not planning any more changes at this point. The last two changes were from bug reports I've received:

* Missing `kid` in the ID token header
* Userinfo endpoint raises an unhandled exception if the id token used is expired or invalid.

I'll triage new bug fixes into a new PR to avoid changing this PR any more :+1:

n2ygk:
> I'll triage new bug fixes into a new PR to avoid changing this PR any more 👍

@tevansuk Sounds like a plan. I assume once we release this as a 1.5.0 release -- hopefully in the next day or two -- there may be a flurry of bug reports, etc. which we can quickly push out as small patch releases. This many commits in a single release is difficult to review.

Thanks.

oauth2_provider/oauth2_backends.py
```diff
@@ -120,9 +125,14 @@ def create_authorization_response(self, request, scopes, credentials, allow):
 
             # add current user to credentials. this will be used by OAUTH2_VALIDATOR_CLASS
             credentials["user"] = request.user
+            request_uri, http_method, _, request_headers = self._extract_params(request)
 
             headers, body, status = self.server.create_authorization_response(
-                uri=credentials["redirect_uri"], scopes=scopes, credentials=credentials
+                uri=request_uri,
```
n2ygk, line 131: Hmm, the [documentation](https://oauthlib.readthedocs.io/en/latest/oauth2/endpoints/authorization.html) for this seems to indicate the previous behavior is correct:

```Post Authorization Request
Generally, this is where you handle the submitted form. Rather than using validate_authorization_request we use create_authorization_response which in the case of Authorization Code Grant embed an authorization code in the client provided redirect uri.```

**Review by n2ygk — CHANGES_REQUESTED** [review]
I'm still not convinced about that `redirect_uri` change you've made. Please convince me otherwise!

File: oauth2_provider/oauth2_backends.py [hunk 13]
```diff
@@ -74,6 +75,10 @@ def extract_headers(self, request):
             del headers["wsgi.errors"]
         if "HTTP_AUTHORIZATION" in headers:
             headers["Authorization"] = headers["HTTP_AUTHORIZATION"]
+        if request.is_secure():
+            headers["X_DJANGO_OAUTH_TOOLKIT_SECURE"] = "1"
+        elif "X_DJANGO_OAUTH_TOOLKIT_SECURE" in headers:
+            del headers["X_DJANGO_OAUTH_TOOLKIT_SECURE"]
```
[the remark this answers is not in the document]
tevansuk: When the OIDC issuer URL is not specified in settings, it is generated on the fly.  It is generated by OAuth2Validator.get_issuer_url(), where we don't have a django request, only an oauthlib request, and the oauthlib request lacks the context of whether the original django request was made over SSL or not.

This leads to it generating an incorrect URL for SSL sites (something like `http://mysite.com:443/o/`). To combat this, we set an additional header in to note whether django considers the request is being made over HTTPS or not, and pull this out when constructing the new URL.

oauth2_provider/oauth2_backends.py
tevansuk: Does it? The code sample below that line explains it more clearly:

```python
# Validate request
uri = 'https://example.com/post_authorize?client_id=foo
headers, body, http_method = {}, '', 'GET'

# Fetch the credentials saved in the pre authorization phase
from your_datastore import fetch_credentials
credentials = fetch_credentials()

# Fetch authorized scopes from the request
from your_framework import request
scopes = request.POST.get('scopes')

from oauthlib.oauth2 import FatalClientError, OAuth2Error
from your_framework import http_response
http_response(body, status=status, headers=headers)
try:
    headers, body, status = server.create_authorization_response(
        uri, http_method, body, headers, scopes, credentials)
```

The `uri` passed to `create_authorization_response` in the example is clearly not the `redirect_uri`, its the URI of the current request. , `oauthlib` doesn't use the `uri` as the `redirect_uri`, it either extracts it from the uri or from the passed in `credentials`. We pass it in the `credentials`, which sets `request.redirect_uri` as an attribute on the `oauthlib.Request`

https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/oauth2/rfc6749/endpoints/authorization.py#L89-L95

Checking the redirect uri uses this request.redirect_uri attribute, not the request URI.

https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/oauth2/rfc6749/grant_types/base.py#L228-L241

Here are examples of how other libraries using oauthlib are handling this, from [flask-oauthlib](https://github.com/lepture/flask-oauthlib/blob/d8369c8ae06150d37da2eba951ae9646235f940b/flask_oauthlib/provider/oauth2.py#L493-L496):

```python
        uri, http_method, body, headers = extract_params()
        try:
            ret = server.create_authorization_response(
                uri, http_method, body, headers, scopes, credentials)
```

and from [bottle-oauthlib](https://github.com/Refinitiv/bottle-oauthlib/blob/d696c8388a8c499cb307f865f9264db47873cf6d/bottle_oauthlib/oauth2.py#L249-L255):

```python
                uri, http_method, body, headers = extract_params(bottle.request)
                scope = bottle.request.params.get('scope', '').split(' ')

                try:
                    resp_headers, resp_body, resp_status = self._oauthlib.create_authorization_response(
                        uri, http_method=http_method, body=body, headers=headers, scopes=scope
                    )
``` 

and from [oauthlib's test cases](https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/tests/oauth2/rfc6749/endpoints/test_client_authentication.py#L84-L90):

```python
        # implicit grant
        auth_uri = 'http://example.com/path?client_id=abc&response_type=token'
        self.assertRaises(ValueError, self.mobile.create_authorization_response,
                auth_uri, scopes=['random'])


        self.validator.validate_client_id.side_effect = self.set_client_id
        h, _, s = self.mobile.create_authorization_response(auth_uri, scopes=['random'])
```

None of them are passing the `redirect_uri` as the `uri` argument of `create_authorization_response`, and I don't see any bit of code or documentation that says we should.

n2ygk (approved):

Natureshadow:
Great work, this is a really good addition!

Making a release very soon would be of much help to the AlekSIS project.

dopry:
@tevansuk hey I'm a new DOT maintainers, I noticed pytest got added to the project in this PR.  I'm trying to get some historic context regarding the mix of test fixture libraries in the project. Do you remember why you chose to add pytest over just working with django.test and decorators like `@override_settings`?

PR #916, opened 2021-01-28
Add live docs rebuilding using sphinx-autobuild

tevansuk:
Add a new tox target, `livedocs`, that uses `sphinx-autobuild` to monitor the docs and auto rebuild the HTML docs when they change.

It also starts a webserver on 127.0.0.1:8000 using `livereload` to serve the docs, which will automatically reload the pages in your browser as they are changed.

Add docs on updating the docs.

Add `sphinx-rtd-theme` so that the docs generated in development match the style that they are displayed in on readthedocs.

This shouldn't affect test run time; the `livedocs` tox target is only run when explicitly asked for, and packages are only installed when using `livedocs`

## Checklist

- [x] unit-test added
- [x] documentation updated


n2ygk (approved):
Hey, this is cool!  Thanks.

PR #945, opened 2021-03-17
update namespace error in example

aryan340:
Because the app_name attribute is not set in the included module (oauth2_provider), we cannot directly include endpoints with the namespace oauth2_provider.
We need to pass in a tuple of the endpoints and app-name instead. Document the same in the tutorial example.

## Description of the Change
Update tutorial documentation regarding namespace inclusion

## Checklist

- [ ] unit-test added
- [x] documentation updated


auvipy (approved):
looks good

CosmicReindeer:
Commit?

n2ygk:
Looks like @auvipy merged this early so it's part of release 1.5.0.

PR #953, opened 2021-04-01
Allow loopback redirect URIs using ports as described in RFC8252

pauldekkers:
## Description of the Change
Allow loopback redirect URIs with ports using http scheme and localhost addresses (127.0.0.1 or ::1),
as described in RFC8252 (section 7.3).

## Checklist
- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -131,6 +131,11 @@ class AbstractApplication(models.Model):
             parsed_allowed_uri = urlparse(allowed_uri)
 
             if (
+                parsed_allowed_uri.scheme == parsed_uri.scheme == "http"
+                and parsed_uri.hostname in ["127.0.0.1", "::1"]
+                and parsed_allowed_uri.hostname == parsed_uri.hostname
+                and parsed_allowed_uri.path == parsed_uri.path
+            ) or (
                 parsed_allowed_uri.scheme == parsed_uri.scheme
                 and parsed_allowed_uri.netloc == parsed_uri.netloc
                 and parsed_allowed_uri.path == parsed_uri.path
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -125,29 +125,7 @@ class AbstractApplication(models.Model):
 
         :param uri: Url to check
         """
-        parsed_uri = urlparse(uri)
-        uqs_set = set(parse_qsl(parsed_uri.query))
-        for allowed_uri in self.redirect_uris.split():
-            parsed_allowed_uri = urlparse(allowed_uri)
-
-            if (
-                parsed_allowed_uri.scheme == parsed_uri.scheme == "http"
-                and parsed_uri.hostname in ["127.0.0.1", "::1"]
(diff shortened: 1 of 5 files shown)


auvipy (dismissed):
thanks for your PR. can you check the PR checklist and update as appripriate? changelog, entry to the author, and possibly unit test?

pauldekkers:
I've modified the patch slightly to only accept random redirect_uri ports when there is no explicit port configured in the Application. I've added a line in the documentation, updated the CHANGELOG and AUTHORS.
I'm struggling a bit with the unit test as there is no existing test and these tests are a bit more complex than the suggested change. Would you accept without?

rtpg:
While reading over this back and forth I took the opportunity to just add in some tests and factor out this logic to make it easier to unit test. Thanks for doing all the legwork on this first though!

I think that this loopback special casing in redirect code (pretty sensitive area, securitywise) warranted a nice chunky comment, but my tests left me pretty happy that this is the "right" way of going forward.

@auvipy since I made a change I feel a bit uncomfortable approving the PR, please look over what I pushed in.

oauth2_provider/models.py
```diff
@@ -674,3 +658,50 @@ def clear_expired():
...
+            and parsed_allowed_uri.port is None
+        )
+        if (
+            allowed_uri_is_loopback
+            and parsed_allowed_uri.scheme == parsed_uri.scheme
+            and parsed_allowed_uri.hostname == parsed_uri.hostname
+            and parsed_allowed_uri.path == parsed_uri.path
+        ) or (
+            parsed_allowed_uri.scheme == parsed_uri.scheme
+            and parsed_allowed_uri.netloc == parsed_uri.netloc
```
rtpg, line 699: at first it looks like both of these branches could be merged! And for `scheme` and `path` it can, but `netloc` is actually inclusive of the port (it's roughly login info + hostname + port) , and for loopback IPs we _don't_ want to be requiring the port to be the same (the whole point is wanting to support ephemeral IPs)

pauldekkers:
@rtpg Thanks for your modifications and adding those tests, I can now tick the unit-test checkbox too ;-)

FWIW, I've tested your changes locally and of course it works just as well.

pauldekkers:
Hi; it looks like this PR is stuck on "changes requested" while I think they were actually applied @auvipy ? I kind of think @rtpg  made this PR even better 👍 but I'm not sure where we stand.

rtpg (approved):
alright a bit convoluted but given that @pauldekkers and I have been going back and forth on this I think we can consider this to be reviewed sufficiently

n2ygk:
@rtpg @pauldekkers - Hey welcome to the project. Thanks for your PR. @MattBlack85 and I are trying to be good [project leads](https://jazzband.co/projects/django-oauth-toolkit) and stay on top of changes and releases and will try to slot this in to a future release after we review it. 

Notably @rtpg I do not see your name in AUTHORS. Please submit a PR to correct that.

kargaj:
Hi, when you are planning the next release and adding this feature?

PR #957, opened 2021-04-06
Fix double oauth2_provider mountpoint in oidc view

### LINKED ISSUE #955: Is documentation correct about OIDC_ISS_ENDPOINT?
The [OIDC_ISS_ENDPOINT](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#oidc-iss-endpoint) documentation states that discovery is at `OIDC_ISS_ENDPOINT + /.well-known/openid-configuration/`. That would indicate that one should include the mount point of the `oauth2_provider.urls` right?

But `ConnectDiscoveryInfoView` uses `reverse` to get the path of the  `oauth2_provider:authorize`, `oauth2_provider:token`, `oauth2_provider:user-info`, `oauth2_provider:jwks-info` which results in the doubling of the mount point info.

So a if `oauth2_provider.urls` is mounted at `/some-initial-path/o` all the endpoints, except `issuer`, included in the response has doubled the mount point information. So if the `OIDC_ISS_ENDPOINT` is `http://localhost:8001/some-initial-path/o` the issuer will be `http://localhost:8001/some-initial-path/o` but `authorization_endpoint` will be `http://localhost:8001/some-initial-path/o/some-initial-path/o/authorize/`. Same pattern for `token_endpoint`, `userinfo_endpoint`, and `jwks_uri` 

Looking at the tests, there seems to be a little ambivalence about the topic. See below
https://github.com/jazzband/django-oauth-toolkit/blob/9d2aac2480b2a1875eb52612661992f73606bade/tests/test_oidc_views.py#L15

`test_get_connect_discovery_info` expects a url without a path and `test_get_connect_discovery_info_without_issuer_url` expects a url with a path `/o` (the default `oauth2` path?).

Anyways - I'm confused. Can anyone clarify if `OIDC_ISS_ENDPOINT` should just be the root url of the Django app or if it should include the mount point of the `oauth2_provider.urls`?

EDIT:
It looks as if the OIDC specification mentions [this](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). The correct pattern should follow the `django_oauth_toolkit` documentation. So `OIDC_ISS_ENDPOINT ` + `/.well-known/openid-configuration` should resolve.
If this is true, then the test `test_get_connect_discovery_info` should expect `http://localhost/o` instead of `http://localhost` as `issuer` - I think.

EDIT2:
If `OIDC_ISS_ENDPOINT ` is defined, couldn't it be located somewhere else (another domain) than where `ConnectDiscoveryInfoView` is located? If yes, isn't it then a mistake to base the location of `authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`, and `jwks_uri` on the use of `reverse` for the url patterns on the same host where `ConnectDiscoveryInfoView` is located.
Why not just hardcode the endpoints to `OIDC_ISS_ENDPOINT` + `{/authorize/, /token/, /userinfo/, o/.well-known/jwks.json}`?
Or urlparse `OIDC_ISS_ENDPOINT` and use the scheme + netloc + reverse of all the endpoints to fill the output of `ConnectDiscoveryInfoView`.

dollarklavs:
## Fixes the doubling of mountpoint path in the OIDC endpoints values for `.well-known/openid-configuration/`

According to the `django-oauth-toolkit` documentation for [OIDC_ISS_ENDPOINT](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#oidc-iss-endpoint) this settings variable should enable discovery at `OIDC_ISS_ENDPOINT` + `/.well-known/openid-configuration/`. This behaviour is backed by the OIDC [specs](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig).

But if you use the variable as described you'll end up with the correct URL for the `issuer` value but incorrect URL's for the values of `authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`, and `jwks_uri` in the json response when calling `/.well-known/openid-configuration/`. 

So if the `OIDC_ISS_ENDPOINT` is `http://localhost:8001/some-initial-path/o` the `issuer` will be `http://localhost:8001/some-initial-path/o` but `authorization_endpoint` will be `http://localhost:8001/some-initial-path/o/some-initial-path/o/authorize/`. Same pattern for `token_endpoint`, `userinfo_endpoint`, and `jwks_uri`

Fixes #955 

## Description of the Change
Instead of concatting `OIDC_ISS_ENDPOINT` with the output of`reverse` for the different OIDC views, first use `urlparse` to strip `OIDC_ISS_ENDPOINT` to only scheme + netloc (eg. `http` + `://` + `localhost:8000`).
Also updates tests to expect an `OIDC_ISS_ENDPOINT` that ends in `/o`.
## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,6 +32,7 @@ Hiroki Kiyohara
 Jens Timmerman
 Jerome Leclanche
 Jim Graham
+Jonas Nygaard Pedersen
 Jonathan Steffan
 Jun Zhou
 Kristian Rune Larsen
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #524 Restrict usage of timezone aware expire dates to Django projects with USE_TZ set to True.
-* #955 Avoid doubling of `oauth2_provider` urls mountpath in json response for OIDC view `ConnectDiscoveryInfoView`
+* #955 Avoid doubling of `oauth2_provider` urls mountpath in json response for OIDC view `ConnectDiscoveryInfoView`.
+  Breaks existing OIDC discovery output
 
 ## [1.5.0] 2021-03-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * #712, #636, #808. Calls to `django.contrib.auth.authenticate()` now pass a `request`
   to provide compatibility with backends that need one.
-  
+
 ### Fixed
 * #524 Restrict usage of timezone aware expire dates to Django projects with USE_TZ set to True.
+* #955 Avoid doubling of `oauth2_provider` urls mountpath in json response for OIDC view `ConnectDiscoveryInfoView`
 
 ## [1.5.0] 2021-03-18
 
--- a/oauth2_provider/views/oidc.py
+++ b/oauth2_provider/views/oidc.py
@@ -1,4 +1,5 @@
 import json
+from urllib.parse import urlparse
 
 from django.http import HttpResponse, JsonResponse
 from django.urls import reverse
@@ -32,12 +33,15 @@ class ConnectDiscoveryInfoView(OIDCOnlyMixin, View):
             )
             jwks_uri = request.build_absolute_uri(reverse("oauth2_provider:jwks-info"))
         else:
-            authorization_endpoint = "{}{}".format(issuer_url, reverse("oauth2_provider:authorize"))
-            token_endpoint = "{}{}".format(issuer_url, reverse("oauth2_provider:token"))
+            parsed_url = urlparse(oauth2_settings.OIDC_ISS_ENDPOINT)
+            host = parsed_url.scheme + "://" + parsed_url.netloc
+            authorization_endpoint = "{}{}".format(host, reverse("oauth2_provider:authorize"))
+            token_endpoint = "{}{}".format(host, reverse("oauth2_provider:token"))
             userinfo_endpoint = oauth2_settings.OIDC_USERINFO_ENDPOINT or "{}{}".format(
-                issuer_url, reverse("oauth2_provider:user-info")
+                host, reverse("oauth2_provider:user-info")
             )
-            jwks_uri = "{}{}".format(issuer_url, reverse("oauth2_provider:jwks-info"))
+            jwks_uri = "{}{}".format(host, reverse("oauth2_provider:jwks-info"))
+
         signing_algorithms = [Application.HS256_ALGORITHM]
         if oauth2_settings.OIDC_RSA_PRIVATE_KEY:
             signing_algorithms = [Application.RS256_ALGORITHM, Application.HS256_ALGORITHM]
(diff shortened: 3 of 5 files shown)


auvipy (changes requested):
can you please add the changelog entry?

dollarklavs:
> can you please add the changelog entry?

It's been added👌

rtpg:
@dollarklavs is it safe to say that, given this is affecting the connection discovery view, that in the current release this is just outright broken (since the URLs are wrong), and so people upgrading don't need to take special precautions when upgrading if they are using this setting (since supposedly the feature just isn't working)? 

If there's some big-ish changes we might want to add something to the changelog to make it clear that this change affects people using this specific setting. 

Things otherwise look good to me as well

dollarklavs:
> @dollarklavs is it safe to say that, given this is affecting the connection discovery view, that in the current release this is just outright broken (since the URLs are wrong), and so people upgrading don't need to take special precautions when upgrading if they are using this setting (since supposedly the feature just isn't working)?
> 
> If there's some big-ish changes we might want to add something to the changelog to make it clear that this change affects people using this specific setting.
> 
> Things otherwise look good to me as well

I would think that OIDC in the current release is broken. So yeah no need to take special precautions - but I/we could probably have overlooked a use case that incorporates the previous implementation.

rtpg:
I think that would be good, yeah. So if somoene is looking over the release notes and are using OIDC they'll click through and maybe realize the change does (or more likely doesn't!) affect them. Thanks for the detailed reply.

dollarklavs:
> I think that would be good, yeah. So if somoene is looking over the release notes and are using OIDC they'll click through and maybe realize the change does (or more likely doesn't!) affect them. Thanks for the detailed reply.

Changelog updated🥳

rtpg (approved):

n2ygk:
@dollarklavs @rtpg I'm concerned that you think OIDC in the current release is broken as I've been using it as have others. I hope this PR hasn't caused a breaking change. @MattBlack85 and I will need to re-review before we publish a release.

rtpg:
Going to try and summarize the issue, because I think the issue itself is relatively limited in scope:
- if you are _not_ setting `OIDC_ISS_ENDPOINT` then the URLS listed in the discovery view will correctly list the URLs for various views according to what you have in your url configuration (so if I set the base `oauth2_providers.urls` to `foobar/` then `http://127.0.0.1:8000/foobar/.well-known/openid-configuration/` will properly report all of those)
- However, as you might know, this ends up working by getting the host from the request (so doing a request via `localhost` will return those URLS etc). Maybe this is part of the reason for this setting to exist in the first place?
- If I decided that I want people to get a canonical hostname and declare the OpenID connect absolute path explicitly, by setting:
```
OAUTH2_PROVIDER = {
   ...,
  "OIDC_ISS_ENDPOINT": "http://localhost:8000/foobar",
}
```

then what I get back from the above `.well-known` view is 

```
{"issuer": "http://127.0.0.1:8000/foobar",
 "authorization_endpoint": "http://127.0.0.1:8000/foobar/foobar/authorize/",
 "token_endpoint": "http://127.0.0.1:8000/foobar/foobar/token/",
 "userinfo_endpoint": "http://127.0.0.1:8000/foobar/foobar/userinfo/",
 ...
}
```

(note the doubling of `foobar` in the URL). To be clear, the endpoints being returned in this response are invalid, this setting doesn't change the routing or anything. So if you were getting these URLs they would be unusable to you, a priori. Now I'm not super clear on the reason one would set this `OIDC_ISS_ENDPOINT` setting in the first place (the canonical thing is a guess), but it definitely feels like in the current configuration there are few ways t set this setting that doesn't cause the discovery view to return invalid data.

But if you don't use this setting, everything seems fine! Hence this not being a major issue yet, I think. Hope this helps

PR #963, opened 2021-04-15
Handles ValueErrors with invalid hex values in query strings (#954)

### LINKED ISSUE #954: ValueError from oauthlib when validating querystrings is raised as 500 instead of 400 HTTP error
**Describe the bug**
When passing in invalid hex encoding in the querystring (for example %%2A), a `ValueError` is raised in https://github.com/oauthlib/oauthlib/blob/b69fa53fd836dc559aa7fcd78ce075bcbe361629/oauthlib/common.py#L395 and the Django application will return 500 HTTP error code.

**To Reproduce**
1. Have an endpoint that accepts the token in a querystring
2. Send an invalid querystring (in our example it was an extra % - not converted to %25) just before a hex value.

Examples:
/endpoint/?evae%%7B7%2A191%7Daith=
/endpoint/?beiy%%%%%%%%3moh=

**Expected behavior**
I think a 400 status would be more appropriate, it's the client that's sending invalid data.

**Version**
1.5.0

**Additional context**
1. First I'd like to ask some help in triaging this issue - is it best fixed in django-oauth-toolkit, or would it be better to be fixed in oauthlib?
2. More than happy to contribute to the issue with a PR

Thank you!

kreatemore:
Fixes #954

## Description of the Change
Invalid hex values result in HTTP 500 errors, the purpose of this PR is to convert them to HTTP 400 ones.

## Checklist

- [X] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/backends.py
+++ b/oauth2_provider/backends.py
@@ -17,8 +17,8 @@ class OAuth2Backend:
         if request is not None:
             try:
                 valid, request = OAuthLibCore.verify_request(request, scopes=[])
-            except ValueError as err:
-                raise SuspiciousOperation(err)
+            except ValueError as error:
+                raise SuspiciousOperation(error)
             else:
                 if valid:
                     return request.user
--- a/oauth2_provider/backends.py
+++ b/oauth2_provider/backends.py
@@ -1,4 +1,5 @@
 from django.contrib.auth import get_user_model
+from django.core.exceptions import SuspiciousOperation
 
 from .oauth2_backends import get_oauthlib_core
 
@@ -14,9 +15,14 @@ class OAuth2Backend:
 
     def authenticate(self, request=None, **credentials):
         if request is not None:
-            valid, r = OAuthLibCore.verify_request(request, scopes=[])
-            if valid:
-                return r.user
+            try:
+                valid, request = OAuthLibCore.verify_request(request, scopes=[])
+            except ValueError as err:
+                raise SuspiciousOperation(err)
+            else:
+                if valid:
+                    return request.user
+
         return None
 
     def get_user(self, user_id):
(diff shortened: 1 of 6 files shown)


oauth2_provider/backends.py
```diff
@@ -14,9 +15,14 @@ class OAuth2Backend:
 
     def authenticate(self, request=None, **credentials):
         if request is not None:
-            valid, r = OAuthLibCore.verify_request(request, scopes=[])
-            if valid:
-                return r.user
+            try:
+                valid, request = OAuthLibCore.verify_request(request, scopes=[])
+            except ValueError as error:
+                raise SuspiciousOperation(error)
```
kreatemore, line 21: As raised in https://github.com/jazzband/django-oauth-toolkit/issues/954#issuecomment-819512062 is catching all `ValueError`s too broad? I think it should be fine, but it might end up exposing error messages users were never meant to see, possibly resulting in security flaws?

rtpg: I'm also a bit worried about this catching "real issues" and causing them to end up not being logged (for example in one Django setup I know `SuspiciousOperation` won't get logged into Sentry but a raw `ValueError` would). So I could imagine some time lost as this absorbs a real error.

The minimalist approach here would be to check if `error.message == "invalid hex string"` (or whatever) and only transform the error then (otherwise just reraising the `ValueError`). But to be honest this might not be a major thing.

kreatemore: I've updated it now to check the message of the error, otherwise re-raise the exception as-is.

auvipy (changes requested):
can you please fix the conflict in changelog

auvipy (approved):

auvipy:
can you plz fix he flake8 errors in py3.8?

kreatemore:
I made some updates @auvipy , but I’m flying a bit blind, as flake8 didn’t report these issues locally. I read the errors from the CI run, and did my best to fix them.

auvipy:
thanks a lot everyone involved

PR #969, opened 2021-04-24
Batch tokens deletion in cleartokens command

### LINKED ISSUE #651: `cleartokens` management command could have some performance improvements
I'm using the `cleartokens` command to clear the expired tokens from my MySQL database and try to improve performance. However, we got to a point where we have a lot of tokens in our database (~300 MM) because of a bug in another application.

The command no longer workers because the `delete()` method from Django tries to load the entire object list to the python memory. The task will make the machine CPU goes to 100% and won't be able to delete anything.

I think we should have some alternatives to clear the tokens. Maybe some kind of pagination so it would delete X tokens at a time or maybe using the `_raw_delete()` method from Django ORM.

Is that an acceptable resolution for the problem? I can try and work in a PR to address this issue.

merito:
Fixes #651 

## Description of the Change

Use batches when deleting expired tokens using `cleartokens` management command. Introduce two setting variables to configure batch size and interval between deletions. The defaults were tested on quite weak machines and ~1.5M deleted tokens. It allows to run `cleartokens` without any downtime.

## Checklist

- [x] unit-test added
- [x] documentation updated


auvipy (approved):

auvipy (dismissed):
unit and integration tests for the changed logics are needed. also need explanation why atomic transacion is not needed here.

n2ygk (commented):
Sorry for the delayed follow up on this PR. It looks good. Please rebase and resolve conflicts. I'm targeting the next minor release for this. Thanks! (I'm afraid to check our production DB to see how many expired tokens are in it;-)

merito:
I had to exclude AUTHORS from check-merge-conflict, because it causes
```
Check for merge conflicts................................................Failed
- hook id: check-merge-conflict
- exit code: 1

Merge conflict string "=======
" found in AUTHORS:2
```
You can keep this change or change separator in AUTHORS file.

.pre-commit-config.yaml
```diff
@@ -10,6 +10,7 @@ repos:
       - id: check-ast
       - id: trailing-whitespace
       - id: check-merge-conflict
+        exclude: AUTHORS
```
n2ygk, line 13: I don't think this is correct. If you are rebasing and there's a merge conflict, fix the conflict, don't exclude the test. Please revert this change.

n2ygk (changes requested):

.pre-commit-config.yaml
merito: No, because there is no merge conflict here. Check for the merge conflicts detects "=======" string from AUTHORS file as a leftover from an unresolved conflict, but it is a regular AUTHORS file content, see AUTHORS on current master: https://github.com/jazzband/django-oauth-toolkit/blob/6017f079abc0da9f66fb0082539a88ce56b03a42/AUTHORS#L2 Without changing checks I can't commit with changes in AUTHORS file.

n2ygk: Oh yeah, I see that now. Wow.

n2ygk:
I think I'll approach this by fixing AUTHORS separately as the problem doesn't "belong" to this PR.

Also, since I'm targeting 1.7.0 for this as it's more than a patch, let's hold it off a bit until 1.6.1. gets published.

n2ygk:
https://github.com/pre-commit/pre-commit-hooks/issues/100 describes the problems with RST underline.

n2ygk (approved):

n2ygk:
@auvipy no test existed prior to this PR so let's accept it as is and put writing a test for the command on the backlog.

oauth2_provider/settings.py
```diff
@@ -101,6 +101,8 @@
     # Whether to re-create OAuthlibCore on every request.
     # Should only be required in testing.
     "ALWAYS_RELOAD_OAUTHLIB_CORE": False,
+    "CLEAR_EXPIRED_TOKENS_BATCH_SIZE": 10000,
+    "CLEAR_EXPIRED_TOKENS_BATCH_INTERVAL": 0.1,
```
n2ygk, line 105: @merito Sorry this is after the fact but wouldn't a default value of 0 be best, especially since the sleep is always executed even if the batch is tiny.
https://github.com/merito/django-oauth-toolkit/blob/725c3c9d8927379c9808abd1badb4fcd9ff1cbaa/oauth2_provider/models.py#L636

PR #975, opened 2021-05-04
Client credentials app throws assertion on auth code flow

### LINKED ISSUE #958: Client Credentials app throws assertion on auth code flow
**Describe the bug**

If you try and use a client credentials app as an authorization code app flow it will cause an assertion error to be thrown.

**To Reproduce**

- Create a client_id as a client_credentials app w/o a redirect_uri
- Attempt Authorization code flow using client_id
- observe 500 and
```AssertionError: If you are using implicit, authorization_codeor all-in-one grant_type, you must define redirect_uris field in your Application model```

**Expected behavior**

I would expect a 400 back to the client letting them know they are using the wrong client type or the 500 be handled.

**Version**

django-oauth-toolkit==1.4.1

**Additional context**

App creation form does not do any validation on `redirect_uris` for a `client_credentials` app. So it appears an end user could create a bad app and the server not really tell them what is happening. It is up to the Oauth server owner to check the logs for the exception then communicate that to the app owner. Also the error does not appear handled as it does not go though the exception block [here](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/views/base.py#L145),

### PR DESCRIPTION
**Pull request by panosangelopoulos**
Fixes #958

## Description of the Change

If you try and use a client credentials app as an authorization code app flow it will cause an assertion error to be thrown, with this PR the return response has 400 status code.

## Checklist

- [X] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -15,7 +15,7 @@ from django.contrib.auth.hashers import check_password
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
 from django.db.models import Q
-from django.http import HttpRequest
+from django.http import HttpRequest, HttpResponseBadRequest
 from django.utils import dateformat, timezone
 from django.utils.timezone import make_aware
 from django.utils.translation import gettext_lazy as _
@@ -300,7 +300,10 @@ class OAuth2Validator(RequestValidator):
         return self._load_application(client_id, request) is not None
 
     def get_default_redirect_uri(self, client_id, request, *args, **kwargs):
-        return request.client.default_redirect_uri
+        try:
+            return request.client.default_redirect_uri
+        except Exception as e:
+            raise HttpResponseBadRequest(e)
 
     def _get_token_from_authentication_server(
         self, token, introspection_url, introspection_token, introspection_credentials
--- a/tests/test_auth_backends.py
+++ b/tests/test_auth_backends.py
@@ -162,3 +162,12 @@ class TestOAuth2Middleware(BaseTest):
         response = m(request)
         self.assertIn("Vary", response)
         self.assertIn("Authorization", response["Vary"])
+
+    def test_get_default_redirect_uri_empty_redirect_uri_raise_exception(self):
+        self.application.authorization_grant_type = Application.GRANT_AUTHORIZATION_CODE
+        self.application.redirect_uris = None
+
+        with self.assertRaises(Exception):
+            self.validator.get_default_redirect_uri(
+                client_id=self.application.client_id, request=self.request
+            )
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: tests/test_auth_backends.py [hunk 1]
```diff
@@ -162,3 +162,12 @@ def test_middleware_response_header(self):
         response = m(request)
         self.assertIn("Vary", response)
         self.assertIn("Authorization", response["Vary"])
+
+    def test_get_default_redirect_uri_empty_redirect_uri_raise_exception(self):
+        self.application.authorization_grant_type = Application.GRANT_AUTHORIZATION_CODE
```
n2ygk, line 167: See the tox log. It looks like this code is failing due to an undefined variable. I suggest you run tox locally before pushing the next commit.

```
=================================== FAILURES ===================================
  _ TestOAuth2Middleware.test_get_default_redirect_uri_empty_redirect_uri_raise_exception _
  
  self = <tests.test_auth_backends.TestOAuth2Middleware testMethod=test_get_default_redirect_uri_empty_redirect_uri_raise_exception>
  
      def test_get_default_redirect_uri_empty_redirect_uri_raise_exception(self):
  >       self.application.authorization_grant_type = Application.GRANT_AUTHORIZATION_CODE
  E       NameError: name 'Application' is not defined
```

n2ygk (changes requested):
Yes, not returning a 500 is always the right thing! This is a good fix.

Please resolve the failing tests. Consider running tox locally before pushing up the next commit(s).

Please let me know if you'll have time to do this. I'm planning it for the next minor release.

n2ygk:
I've pushed some simple flake8 changes but there are still several errors in this PR.

tests/test_auth_backends.py
```diff
@@ -162,3 +162,12 @@ def test_middleware_response_header(self):
         response = m(request)
         self.assertIn("Vary", response)
         self.assertIn("Authorization", response["Vary"])
+
+    def test_get_default_redirect_uri_empty_redirect_uri_raise_exception(self):
```
n2ygk, line 166: This test case is in the wrong place: `self.application` does not exist here.

oauth2_provider/oauth2_validators.py
```diff
@@ -300,7 +300,10 @@ def validate_client_id(self, client_id, request, *args, **kwargs):
         return self._load_application(client_id, request) is not None
 
     def get_default_redirect_uri(self, client_id, request, *args, **kwargs):
-        return request.client.default_redirect_uri
+        try:
+            return request.client.default_redirect_uri
+        except Exception as e:
+            raise HttpResponseBadRequest(e)
```
n2ygk, line 306: HttpResponseBadRequest is not an Exception. You probably want to raise [BadRequest](https://docs.djangoproject.com/en/3.2/ref/exceptions/#badrequest).

n2ygk (changes requested):
The main code change is incorrect as is the test case. Please take a look.

oauth2_provider/oauth2_validators.py
n2ygk: And the test should be for a more specific Exception.

PR #978, opened 2021-05-07
Multiple rsa keys

### LINKED ISSUE #950: Multi RS256 Keys support in OpenID
hello, it is really nice to hear that we now add support for openid, we now use django-oidc-provider to build openid service but its last release was in 2019, so we try to use another active project.

After read the docs and code , i find that this project not provide multi rs256 keys support, right?

as describe in the document: https://openid.net/specs/openid-connect-core-1_0.html#SigEnc , we can manully add keys.

will this feat be considered? thanks a lot for your work 😊

dylantack:
Fixes #950 

## Description of the Change
This adds support for multiple RSA keys via the OIDC_RSA_PRIVATE_KEYS_INACTIVE setting.
Additionally, jwks_uri has Cache-Control header as recommended by the [OpenID spec](https://openid.net/specs/openid-connect-core-1_0.html#RotateEncKeys).

## Checklist

- [x] unit-test added
- [x] documentation updated


dylantack:
To be very precise about the Cache-Control header, the spec only mentions it in the context of _encryption_ keys, not signing keys. However jwks_uri would be the same resource for both, and I think even for signing the max-age is valuable information for the verifier. Without it, the verifier doesn't have any guidance on how often to re-check for revoked keys.

Here are some examples from jwks_uri on popular ID providers:

**Auth0:**
```Cache-Control: public, max-age=15, stale-while-revalidate=15, stale-if-error=15```

**Firebase Auth:**
```Cache-Control: public, max-age=22159, must-revalidate, no-transform```

**Okta:**
```cache-control: max-age=3293198, must-revalidate```

**Azure Active Directory B2C:**
```Cache-Control: max-age=86400, private```

So it seems like there's quite a bit of variation; I have defined OIDC_JWKS_MAX_AGE_SECONDS with a default of 1 hour.

n2ygk (approved):
Nice! Thanks.

PR #988, opened 2021-06-18
Optimize DB access in AccessTokenAdmin

knaperek:
## Description of the Change

This is needed to avoid 2 extra DB queries per each line in the list view, so that is usually +200 unnecessary queries.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,6 +36,7 @@ Jerome Leclanche
 Jim Graham
 Jonas Nygaard Pedersen
 Jonathan Steffan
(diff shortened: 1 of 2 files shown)


knaperek:
CHANGELOG is not updated as this is not a user relevant change. I marked that tickbox as having that step completed, that is, updated CHANGELOG in case of a user relevant change; which it wasn't, so this is a no-op :-)

auvipy:
ok

PR #1012, opened 2021-09-15
Not existing tokens should return 200 within introspection (not 403)

ZuSe:
Compare with https://datatracker.ietf.org/doc/html/rfc7662

Fixes #

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -47,6 +47,7 @@ Michael Howitz
 Paul Dekkers
 Paul Oswald
 Pavel Tvrdík
+Patrick Palacin
 Peter Carnesciali
 Petr Dlouhý
 Rodney Richardson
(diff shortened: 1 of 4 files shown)


auvipy (approved):

auvipy (dismissed):
you need to update the tests as well

n2ygk (changes requested):
Thanks for this. It looks right but needs the tests fixed. Also, you indicated that CHANGELOG was updated, AUTHORS, etc. but I don't see where you've done that.

The tests looks pretty easy to fix: 
```
  E       AssertionError: 200 != 401
```

Please let me know if you can get to this soon as I'm targeting this for the next minor release.

n2ygk:
I've updated the checklist to indicate correct status of this PR.

n2ygk (approved):

PR #1014, opened 2021-09-24
Update tutorial_03.rst to use arrays instead of tuples in the settings.py file

sdwoodbury:
`AUTHENTICATION_BACKENDS` and `MIDDLEWARE` should be arrays, not tuples. Using tuples seems to work, but everything else in the settings.py file is an array.

Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated


auvipy (approved):
list not array.

Issue #177, opened 2014-12-23
Auto increment id for accesstoken table is not sufficient for large applications


oreh:
Currently, the PK of accesstoken table is using int type (as default django table id does). But this 32-bit field can only hold around 2 billions records. The keys will be used out in a year if you have 10 millions users who generate new keys on a daily basis. 

Can we change the type from int to bigint? Or even better, can we use 'token' as the PK?

masci:
We're currently planning to support both django1.7 and south1.0 migrations, so we prefer to postpone the resolution of this issue after that.

jleclanche:
I'd like to change all IDs to use BigAutoField but this requires Django 1.10 so we might not be able to do it for a while.

PR #1029, opened 2021-11-12
Explicitly change the oauth2_provider_idtoken.application_id from bigint(20) to int(11)

jasonm:
This is a draft PR - I'm not sure this is an appropriate fix. Still investigating. I believe this is related to updating an old 0.12.0 installation to 1.5.0 and hitting an issue across #177 

Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/migrations/0004_auto_20200902_2022.py
+++ b/oauth2_provider/migrations/0004_auto_20200902_2022.py
@@ -42,6 +42,17 @@ class Migration(migrations.Migration):
                 'swappable': 'OAUTH2_PROVIDER_ID_TOKEN_MODEL',
             },
         ),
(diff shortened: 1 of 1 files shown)


jasonm:
I see... I think the issue is that we have an existing 0.12.0 installation and want to update to 1.5.0. I see the changelog note about the merged migrations in 1.2.0. There is also the change from int PKs to bigint. Updating to 1.2.0 and running `manage.py migrate` doesn't do the trick, and I went down a small rabbit hole of writing SQL to manually move our PK/FK fields to bigint. That got stuck on having unpredictable FK constraint identifiers, so I'm going to back up and try an approach for our project that does not require upgrading this library right now.

PR #1041, opened 2021-12-16
upgrades to admin: search_fields, list_filters and raw_id_field

PetrDlouhy:
Fixes:

* long loading times of admin edit pages when many User/Token models

## Description of the Change

I defined `search_fields`, `list_filters` and `raw_id_fields` for easier management of the tokens in admin.

Not sure what to test in this.
Also I think, this doesn't need any documentation, it is pretty self-explanatory to users.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/admin.py
+++ b/oauth2_provider/admin.py
@@ -14,7 +14,8 @@ from oauth2_provider.models import (
     get_refresh_token_model,
 )
 
-has_email = hasattr(get_user_model(), 'email')
+
+has_email = hasattr(get_user_model(), "email")
 
 
 class ApplicationAdmin(admin.ModelAdmin):
@@ -42,11 +43,9 @@ class GrantAdmin(admin.ModelAdmin):
 
 class IDTokenAdmin(admin.ModelAdmin):
     list_display = ("jti", "user", "application", "expires")
-    raw_id_fields = ("user", )
+    raw_id_fields = ("user",)
     search_fields = ("token",) + (("user__email",) if has_email else ())
     list_filter = ("application",)
-
-
 
 
 class RefreshTokenAdmin(admin.ModelAdmin):
--- a/oauth2_provider/admin.py
+++ b/oauth2_provider/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django.contrib.auth import get_user_model
 
 from oauth2_provider.models import (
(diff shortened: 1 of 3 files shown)


PetrDlouhy:
This is my second take on this, I fixed the User without e-mail problem, that was in #723

auvipy (approved):

auvipy:
thank you

n2ygk:
In hindsight, it would probably be a good thing for the Admin UI to be documented somewhere....

PetrDlouhy:
@n2ygk Which part of the Admin UI do you think should be documented. Do you mean the UI as general? Or just the search by e-mail functionality (which don't work if there is no e-mail)?

n2ygk:
@PetrDlouhy not here, but as an update to the Tutorial it would probably be helpful to show users around the admin UI. I'll make an issue.

oauth2_provider/admin.py
```diff
@@ -38,11 +44,15 @@ class GrantAdmin(admin.ModelAdmin):
 class IDTokenAdmin(admin.ModelAdmin):
     list_display = ("jti", "user", "application", "expires")
     raw_id_fields = ("user",)
+    search_fields = ("token",) + (("user__email",) if has_email else ())
```
danlamanna, line 47: Is this valid? Searching from the admin now gives me `django.core.exceptions.FieldError: Cannot resolve keyword 'token' into field. Choices are: access_token, application, application_id, created, expires, id, jti, scope, updated, user, user_id`

PetrDlouhy:
@danlamanna You are right. Fixed in PR #1085

charettes:
FWIW it might have been better to search by `USERNAME_FIELD` of the user model instead as nothing guarantees that the email field will be indexed `USERNAME_FIELD` must be to the unique constraint defined on it.

n2ygk:
> FWIW it might have been better to search by `USERNAME_FIELD` of the user model instead as nothing guarantees that the email field will be indexed while `USERNAME_FIELD` must be to support the unique constraint defined on it.

Yes, that makes sense @charettes. A PR to change this would be appreciated.

charettes:
@n2ygk done in https://github.com/jazzband/django-oauth-toolkit/pull/1428.

PR #1046, opened 2021-12-20
support django 4.0.0

dulmandakh:
## Description of the Change

When I try do install **django-oauth-toolkit** using Poetry I get following error. So this PR fixes django 4.0.0 support.

```python
SolverProblemError

  Because no versions of django-oauth-toolkit match >1.6.0,<2.0.0
   and django-oauth-toolkit (1.6.0) depends on django (>=2.2,<4.0.0 || >4.0.0), django-oauth-toolkit (>=1.6.0,<2.0.0) requires django (>=2.2,<4.0.0 || >4.0.0).
  And because no versions of django match >4.0.0,<5.0, django-oauth-toolkit (>=1.6.0,<2.0.0) requires Django (>=2.2,<4.0.0 || >=5.0).
  So, because backend depends on both Django (^4.0) and django-oauth-toolkit (^1.6.0), version solving failed.
```

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   -->
 
 ## [Unreleased]
+* Fix Django 4.0.0 support
 
 ## [1.6.0] 2021-12-19
 ### Added
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Requirements
 ------------
 
 * Python 3.7+
-* Django 2.2, 3.2, or >=4.0.1
+* Django 2.2, 3.2, or >=4.0
 * oauthlib 3.1+
 
 Installation
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -22,7 +22,7 @@ Requirements
 ------------
 
 * Python 3.7+
-* Django 2.2, 3.2, 4.0.1+
+* Django 2.2, 3.2, 4.0+
 * oauthlib 3.1+
 
 Index
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ zip_safe = False
 # jwcrypto has a direct dependency on six, but does not list it yet in a release
 # Previously, cryptography also depended on six, so this was unnoticed
 install_requires =
-	django >= 2.2, != 4.0.0
+	django >= 2.2
 	requests >= 2.13.0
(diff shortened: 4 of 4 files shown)


Andrew-Chen-Wang:
Django 4.0.0 support will not be included due to a regression that caused a migration file to be created. My past experience indicates many devs actually create and migrate these in their virtual environments **on production**. When Django 4.0.1 is released, those who upgrade their Django will have migration errors if we allow support of Django 4.0.0.

Thanks for understanding.

n2ygk:
@Andrew-Chen-Wang  the problem is that we are incorrectly reporting 4.0 support in setup.cfg:
```
        Framework :: Django :: 4.0
```
So we should undo that and anywhere else where it's happening.

Also, how is it that the 4.0 tox tests are succeeding?

Andrew-Chen-Wang:
tl;dr 4.0.0 is technically compatible, and that's why tox still runs fine with 4.0.0. What  will be a massive future headache is when people make migrations on production in a virtual environment. I'd like to avoid that. [Ref past headache; this package will have it worse because it's a foreign key](https://github.com/jazzband/djangorestframework-simplejwt/pull/411)

---

@n2ygk there is no user behavior that makes 4.0.0 incompatible with django-oauth-toolkit. The problem is that a new migration file can be created. If you have a custom User model, the new migration will cause the foreign key on one of the django-oauth-toolkit tables to reference `auth.user`. But if you don't use that table, then your production environment is broken (unless the package makes changes so that your accidental migration file won't break on a future package release, [ref](https://github.com/jazzband/djangorestframework-simplejwt/pull/411)). If you're careful, you won't create the migration file; if not, your production will go down on Django 4.0.1+ release.

In our tox environment, we don't have a custom user model. There is no API that has changed, so I left the 4.0 tox running.

> So we should undo that and anywhere else where it's happening.

Perhaps...? It definitely causes confusion given 4.0.1 isn't released, but I don't really think a patch version just to say "4.0 is supported" in the trove classifier.

Andrew-Chen-Wang:
@dulmandakh Apologies for the poetry resolving problem. Poetry should have a feature to allow you to install django-oauth-toolkit from git. Install at a commit that does not include our new setup.cfg change and you should be good to go with your current production. After Django 4.0.1 release, revert back to package version number in poetry config. Sorry!

n2ygk:
Closing this PR as it won't be merged.

PR #1054, opened 2021-12-22
Clarify why Django 4.0.0 isn't supported

Andrew-Chen-Wang:
Fixes #1051 

## Description of the Change

Clarifies why Django 4.0.0 specifically isn't supported but 4.0.1 will be supported.

Updates README (for upfront dev purposes) and docs (updated CHANGELOG since that goes into our RTD for SEO/Google and RTD search engine purposes).

## Checklist

- [X] PR only contains one change (considered splitting up PR)
- [X] unit-test added
- [X] documentation updated
- [X] `CHANGELOG.md` updated (only for user relevant changes)
- [X] author name in `AUTHORS`

PR #1055, opened 2021-12-22
WIP: Add test to check whether a migration can be created

Andrew-Chen-Wang:
Fixes #1054 

## Description of the Change

~~Adds a test case to see if a Django version creates a new migration. If it does, decide whether we incorporate the change into the package itself or to simply exclude the Django version that created this regression.~~

I completely forgot this isn't a Django project, it's a package :)

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,6 +1,7 @@
 import pytest
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ImproperlyConfigured, ValidationError
+from django.core.management import call_command
 from django.test import TestCase
 from django.test.utils import override_settings
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


Andrew-Chen-Wang:
Interestingly enough, I can't actually seem to produce the error in test cases:

```python
    def test_makemigrations(self):
        try:
            call_command("makemigrations", "oauth2_provider", "--check", "--noinput", "--dry-run")
        except SystemExit:
            self.fail("Migration should be created or there is a regression in Django")
        try:
            with self.settings(AUTH_USER_MODEL="auth.User"):
                call_command("makemigrations", "oauth2_provider", "--check", "--noinput", "--dry-run")
        except SystemExit:
            self.fail("Migration should be created or there is a regression in Django")
```

Run:
```shell
tox -e py39-dj40 tests/test_models.py
```

Output:
```
tests/test_models.py ...No changes detected in app 'oauth2_provider'
No changes detected in app 'oauth2_provider'
..............^X.......^CERROR: got KeyboardInterrupt signal
```

n2ygk:
I don't think you can exec a check migrations command because the database context for tests is an in-memory database create just for the test cases to use.

n2ygk:
Closing this as no longer needed

PR #1056, opened 2021-12-23
Add migration due to noop FK from Django 4.0

Andrew-Chen-Wang:
## Description of the Change

I wanted to double check that 4.0.1 wouldn't create a new migration. Turns out, I missed a comment from the PR that mentions how a migration would still be created. The reason is because of the related_name using the placeholder itself now rather than using "oauth2_provider".

Fortunately, this migration isn't as drastic as the `auth.User` one which would have deleted data... Let me know if I'm missing anything else.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 * #1012 Return status for introspecting a nonexistent token from 401 to the correct value of 200 per [RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).
 * #1068 Revert #967 which incorrectly changed an API. See #1066.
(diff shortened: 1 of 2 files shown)


n2ygk:
I'm marking this as a draft until I get a chance to dig in deeper to better understand the impact.

Unfortunately swappable models and migrations in DOT have always been a problem, with lots of manual tweaking of auto migrations required in the past. I've never completely understood the reasoning....

n2ygk:
@Andrew-Chen-Wang I'm reading the PR commit https://github.com/django/django/commit/43289707809c814a70f0db38ca4f82f35f43dbfd and don't see where it says the new migration should be created....

Andrew-Chen-Wang:
https://github.com/django/django/pull/15205#issuecomment-995474183 and its reply

n2ygk:
> [django/django#15205 (comment)](https://github.com/django/django/pull/15205#issuecomment-995474183) and its reply

Ah so it's in 4.0 release notes. The 4.0.1 thing talked about a different fix that you reported the bug for.  Got it. So we'll have this noop migration....

n2ygk (approved):

PR #1057, opened 2021-12-23
Replace RST header underline with one that pre-commit check-merge won't break on.

n2ygk:
Fixes #

## Description of the Change
Workaround problem where the precommit merge hook breaks on the 7-character doulbe-underline for AUTHORS by having a false-positive broken commit error. 

It pains me to do this but I'd rather catch actual merge conflicts than exclude checking this file.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,11 +1,11 @@
 Authors
-=======
+-------
 
 Massimiliano Pippi
 Federico Frenguelli
 
 Contributors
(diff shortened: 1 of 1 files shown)


n2ygk:
See https://github.com/pre-commit/pre-commit-hooks/issues/100

PR #1068, opened 2022-01-04
#1066: Revert #967 which incorrectly breaks API.

### LINKED ISSUE #1066: Backwards-incompatible API change in OAuth2Validator in 1.6.0
**Describe the bug**

PR #967 changed the API for `OAuth2Validator.get_additional_claims` in a backwards-incompatible way (and the project major version has not been bumped, which would have been required by semantic versioning).

Specifically, the `request` argument was removed, which we relied upon downstream. It seems that the functionality has remained (in a slightly more complex way), but such breakage is not expected between minor versions.

I propose to revert the change in 1.6.1, and make a new release 2.0.0 introducing the breaking change.

**Expected behavior**

We expect the (documented) API not to break backwords-incompatibly in minor versions.

**Version**
1.6.0

n2ygk:
Fixes #1066 

## Description of the Change
PR #967 should never have been merged as it changes the signature of a published API.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,6 +65,5 @@ pySilver
 Łukasz Skarżyński
 Shaheed Haque
 Peter Karman
-Andrea Greco
 Vinay Karanam
 Eduardo Oliveira
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #1012 Return status for introspecting a nonexistent token from 401 to the correct value of 200 per [RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).
+* #1068 Revert #967 which incorrectly changed an API. See #1066.
 
 ## [1.6.1] 2021-12-23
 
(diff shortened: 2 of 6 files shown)


Natureshadow:
@n2ygk Thanks for looking into this!

By now, we have a compatibility layer downstream. If you are interested, isntead of removing the feature the PR introduced, I could try to make a PR adding this compatibility layer to django-oauth-toolkit itself.

On the other hand, I think the feature was implemented incompletely, because publishing all available claims in the discovery info might be a privacy or security risk for some applications. So I think the API should provide for a new validator method, e.g. `get_discovery_claims`, which can override the list of published claims.

What do you think?

Natureshadow (approved):

n2ygk:
> @n2ygk Thanks for looking into this!
> 
> By now, we have a compatibility layer downstream. If you are interested, isntead of removing the feature the PR introduced, I could try to make a PR adding this compatibility layer to django-oauth-toolkit itself.
> 
> On the other hand, I think the feature was implemented incompletely, because publishing all available claims in the discovery info might be a privacy or security risk for some applications. So I think the API should provide for a new validator method, e.g. `get_discovery_claims`, which can override the list of published claims.
> 
> Additionally, the validator implementation should be free to return plain data instead of a callable, like before (both should be supported).
> 
> I could make these additions on top of @AndreaGreco's PR.
> 
> What do you think?

It was done incorrectly and the release has been out such a short time that we can just put it back the way it was for now. No need to add compat layer.

n2ygk:
@Natureshadow Also, feel free to start a new issue/PR. I don't quite understand what you are getting at but would love to review it. I didn't want that conversation to be in this thread though;-)

PR #1070, opened 2022-01-05
Add Celery task for cleantokens

Natureshadow:
## Description of the Change

This adds a Celery task to clean expired tokens. It can be used to schedule token cleanup in the background, for applications using Celery and Celery beat.

The `tasks` module is autodiscovered in projects that use Celery, and should not be loaded in other projects, so this does not introduce any new dependency,

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,3 +66,4 @@ pySilver
 Shaheed Haque
 Vinay Karanam
 Eduardo Oliveira
+Dominik George
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * #651 Batch expired token deletions in `cleartokens` management command
 * Added pt-BR translations.
+* #1070 Add a Celery task for clearing expired tokens, e.g. to be scheduled as a [periodic task](https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html)
 
 ### Fixed
 * #1012 Return status for introspecting a nonexistent token from 401 to the correct value of 200 per [RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).
--- a/docs/management_commands.rst
+++ b/docs/management_commands.rst
@@ -21,3 +21,8 @@ To prevent the CPU and RAM high peaks during deletion process use ``CLEAR_EXPIRE
 
 Note: Refresh tokens need to expire before AccessTokens can be removed from the
 database. Using ``cleartokens`` without ``REFRESH_TOKEN_EXPIRE_SECONDS`` has limited effect.
+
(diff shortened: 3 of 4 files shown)


Natureshadow:
> Hey, this is cool. Is there a way to provide some test coverage of this?

Sure there is. But it requires to run Celery alongside the test suite. I don't think it is worth adding that complexity to the test suite, given that the test would ensure nothing more that a simple function call calls a function, and Celery can run Python code (which is hopefully tested in Celery itself).

n2ygk:
Can this functionality be accommodated external to DOT such as in your Django app? I'm not sure we should be adding code that's not core DOT functionality here. 
@MattBlack85 what do you think?

Natureshadow:
> Can this functionality be accommodated external to DOT such as in your Django app? I'm not sure we should be adding code that's not core DOT functionality here.

Sure it can. As can the management command – that is just the same thing, only that it is called from the command line.

It's nice to have things that make sense to be run periodically pop up in django-admin automatically. Tasks are reusable by design.

n2ygk:
@Natureshadow I'm trying to understand what to expect with Celery. Is the defined task supposed to show up in Django Admin? What additional configuration is needed to make this happen in, for example, https://github.com/n2ygk/dot-tutorial?

I made a few changes but am missing something fundamental I think. Do some templates need to be added oauth2_provider?

Natureshadow:
> I'm trying to understand what to expect with Celery. Is the defined task supposed to show up in Django Admin?

Depends on how the project orchestrates tasks. Most will probably resort to using django-celery-beat.

> What additional configuration is needed to make this happen in, for example, https://github.com/n2ygk/dot-tutorial?
> 
> I made a few changes but am missing something fundamental I think. Do some templates need to be added oauth2_provider?

> 

I will look into that tomorrow.

auvipy (commented):
if the defined scheduled task's to be shown in admin might need to use django-celery-beat & django-celery-results to be used/configured to specific project. [I actively maintain both but more work need to be done before deciding to add celery as dependency]

n2ygk:
> if the defined scheduled task's to be shown in admin might need to use django-celery-beat & django-celery-results to be used/configured to specific project. [I actively maintain both but more work need to be done before deciding to add celery as dependency]

I'm looking for specific configuration changes to my test DOT app so that I can see the celery tasks in the admin console.

n2ygk:
> I will look into that tomorrow.

Hi @Natureshadow, just checking in to see if you've had time to look at this.

Natureshadow:
I checked what we did in AlekSIS for this, and it seems [the steps laid out in the Celery documentation](https://docs.celeryproject.org/en/latest/userguide/periodic-tasks.html#using-custom-scheduler-classes) are sufficient.

Natureshadow:
> Please rebase this PR and I'll merge it.
> 
> Thanks.

Done!

n2ygk (approved):

PR #1080, opened 2022-01-09
Fix bug with hashed client secret passed verbatim

### LINKED ISSUE #1077: client secret hashing allows use of the hash value of a hashed secret.
**Describe the bug**

You can use the hashed value of a client secret as the secret. This eliminates the protection supposedly afforded by #1020 storing client secrets as hashes. 

**To Reproduce**

1. Set up a simple DOT provider as described in the [tutorial](https://django-oauth-toolkit.readthedocs.io/en/latest/tutorial/tutorial_01.html)
2. Add `'CLIENT_SECRET_HASHER': 'pbkdf2_sha256'` to `settings.OAUTH2_SETTINGS`.
3. `./manage.py runserver`
4. Go to http://127.0.0.1:8000/admin/oauth2_provider/application/add/ and add a client: with:
     - client_id: "hash1"
     - client_secret: "hashed"
     - client_type: public
     - authorization grant type: client credentials
    This will generate a hash like: `pbkdf2_sha256$320000$hash2$k1C/17bZRVQGuqpiAF5lR5E9XXt7+/0i5l4QQxdq2uQ=`
5. Use Postman to request a new client credentials token using client_id and secret as above. Confirm it works.
6. Use Postman with an incorrect secret. Confirm it fails.
7. Use Postman with the hash string as the client secret. Confirm it works.

**Expected behavior**

Final step above should fail.

**Version**

Master at #1020

**Additional context**

pkarman:
Fixes #1077 

## Description of the Change

Passing a hashed secret in the basic auth header should not be treated as legitimate.

In addition to the `$` sentinel character, I also discovered that sometimes the hashed secret contains a `+` character which would be decoded as a space ` `. This meant that the new test was flakey (e.g. passed locally but not in CI).

This change checks for exactly 3 `$` characters as a stronger indicator of a hashed secret, and explicitly disallows passing hashed secrets via basic auth.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -129,15 +129,16 @@ class OAuth2Validator(RequestValidator):
         # used by our default generator does not include the "$" character.
         # However, if a different character set was used to generate the secret, this sentinel
         # might be a false positive.
-        elif (
-            "$" in request.client.client_secret
-            and "$" not in client_secret
-            and request.client.client_secret != client_secret
-        ):
-            if not check_password(client_secret, request.client.client_secret):
-                log.debug("Failed basic auth: wrong hashed client secret %s" % client_secret)
-                return False
-            return True
+        # If a hashed secret were passed verbatim in the header and it contained a "+" it will be decoded
+        # as a space, which is also not part of the default character set.
+        elif "$" in client_secret or " " in client_secret:
+            log.debug("client_secret passed in HTTP header appears to be hashed secret")
+            return False
+        elif "$" in request.client.client_secret and request.client.client_secret != client_secret:
+            if check_password(client_secret, request.client.client_secret):
+                return True
+            log.debug("Failed basic auth: wrong hashed client secret %s" % client_secret)
+            return False
         elif request.client.client_secret != client_secret:
             log.debug("Failed basic auth: wrong client secret %s" % client_secret)
             return False
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -129,7 +129,11 @@ class OAuth2Validator(RequestValidator):
         # used by our default generator does not include the "$" character.
         # However, if a different character set was used to generate the secret, this sentinel
         # might be a false positive.
-        elif "$" in request.client.client_secret and request.client.client_secret != client_secret:
+        elif (
+            "$" in request.client.client_secret
+            and "$" not in client_secret
+            and request.client.client_secret != client_secret
+        ):
             if not check_password(client_secret, request.client.client_secret):
                 log.debug("Failed basic auth: wrong hashed client secret %s" % client_secret)
                 return False
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -127,14 +127,15 @@ class OAuth2Validator(RequestValidator):
         # whether a secret has been hashed like a Django password or not.
         # We can do this because the default oauthlib.common.UNICODE_ASCII_CHARACTER_SET
         # used by our default generator does not include the "$" character.
-        # However, if a different character set was used to generate the secret, this sentinel
+        # If a different character set was used to generate the secret, this sentinel
         # might be a false positive.
-        # If a hashed secret were passed verbatim in the header and it contained a "+" it will be decoded
-        # as a space, which is also not part of the default character set.
-        elif "$" in client_secret or " " in client_secret:
+        # However, the django password storage pattern contains exactly 3 "$" characters.
+        # <algorithm>$<iterations>$<salt>$<hash>
+        # The 3 $ pattern is therefore a strong indication of a hashed secret.
+        elif client_secret.count("$") == 3:
             log.debug("client_secret passed in HTTP header appears to be hashed secret")
             return False
-        elif "$" in request.client.client_secret and request.client.client_secret != client_secret:
+        elif request.client.client_secret.count("$") == 3 and request.client.client_secret != client_secret:
             if check_password(client_secret, request.client.client_secret):
                 return True
             log.debug("Failed basic auth: wrong hashed client secret %s" % client_secret)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * #651 Batch expired token deletions in `cleartokens` management command
 * Added pt-BR translations.
 * #729 Add support for [hashed client_secret values](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html#client-secret-hasher).
+* #1077 Bug fix for hashed client secrets
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


pkarman:
@n2ygk 2nd effort here. The test flakiness was a real puzzler. It would pass/fail about 50/50 on repeated runs. Turned out the `+` character was the culprit. It ended up being a good thing to catch, as my original fix attempt was wrong.

oauth2_provider/oauth2_validators.py
```diff
@@ -127,13 +127,19 @@ def _authenticate_basic_auth(self, request):
...
         # might be a false positive.
-        elif "$" in request.client.client_secret and request.client.client_secret != client_secret:
-            if not check_password(client_secret, request.client.client_secret):
-                log.debug("Failed basic auth: wrong hashed client secret %s" % client_secret)
-                return False
-            return True
+        # However, the django password storage pattern contains exactly 3 "$" characters.
+        # <algorithm>$<iterations>$<salt>$<hash>
+        # The 3 $ pattern is therefore a strong indication of a hashed secret.
+        elif client_secret.count("$") == 3:
```
n2ygk, line 135: Here's the problem with this: While autogenerated secrets don't include '$' in them, there's nothing to stop one from manually entering a cleartext secret containing three '$' causing a failure, including when stored as a hash. Simple test case: manually set the application client secret to a string containing three $.

Why not never compare request.client.client_secret to client_secret at all and make sure any previously-created cleartext secrets are upgraded as part of a migration?  See https://docs.djangoproject.com/en/4.0/topics/auth/passwords/#password-upgrading-without-requiring-a-login

If the proffered client secret is a stolen hash string, it won't matter since it will be treated as cleartext and rehashed.

n2ygk (changes requested):

oauth2_provider/oauth2_validators.py
pkarman: I don't use this library, so I was making some assumptions about how it is used in the wild when I first implemented this feature. My original design goals were to be the least intrusive possible and backwards-compatible with existing installations, where the `client_secret` is stored in plaintext. That's why `oauth2_settings.CLIENT_SECRET_HASHER` was made optional.

So that's what sparked the pattern checking as an indicator of whether a `client_secret` was hashed or not. I had 2 use cases in mind:

* a new installation where there is no existing oauth data stored
* an existing installation where there is existing oauth data stored

For the first case, just set `oauth2_settings.CLIENT_SECRET_HASHER` and all client secrets are hashed correctly from the start.

For the second case, if you set `oauth2_settings.CLIENT_SECRET_HASHER` then you immediately have the possibility of a mixed data set, some secrets hashed and some not. And, if you ever change the hashing scheme in that config, you have the possibility of an even more mixed set with multiple schemes.

It was this second case I had in mind with the character checks, since relying solely on `oauth2_settings.CLIENT_SECRET_HASHER` to determine whether to compare with `check_password` rather than simple `!=` meant that existing secrets would no longer work.

I considered the approach you suggest, with a required migration of all existing secrets to use the hashing scheme. That just felt more intrusive to me. But again, I don't use the library, so it wasn't clear to me what was considered "normal" and what was not.

It seems highly unlikely that anyone would be manually setting application client secrets, since they are by default randomly generated strings with a constrained character set, and it seems even more unlikely that someone would deliberately use three `$` characters. So that was not a use case I considered. It's true that it's a security risk, but such a small one (after all, if a bad actor can manually set client secrets, your system has bigger problems) that I didn't consider it worth spending time on.

If you believe it's appropriate to add a migration to change all existing client secrets to use hashing, rather than checking string patterns at run time, I can do that. That's what I would call the "all or nothing" approach. There are a couple risks:

* I believe we should also then make `oauth2_settings.CLIENT_SECRET_HASHER` required rather than optional, so that we know what scheme to use, and relatedly,
* If someone were to set it, or change it, after the migration has run, then they would need to manually migrate all their secrets again

In other words, crossing the border from plaintext to hashed via a config setting means that individual installations take on the burden of maintaining their own data sets, in a way similar to passwords, with custom migrations, etc. That feels way outside the scope of this library, just as it is outside the scope of Django's password feature. If you want to change the scheme for your installation, you also assume the burden of migrating data.

If `oauth2_settings.CLIENT_SECRET_HASHER` were not only required but also hard coded to a particular scheme like `pbkdf2_sha256` as we do in the tests (i.e. not configurable in settings but a constant defined somewhere), then I think it's fair that this library also maintain its own migrations.

As I said at the top, one of my design biases was to be the least intrusive possible, so creating a new optional config was more in line with my bias than forcing a setting and adding a migration (and assuming the burden of that setting and subsequent migrations in the future).

Thanks for reading. Interested to hear what you think.

n2ygk: The ability to set client IDs and secrets and not use the default values is in the UI:
Screen Shot 2022-01-09 at 9 49 56 AM
I don't want to have to assume that someone hasn't or won't ever set a secret that has 3 $'s in it. Who knows? `Show$me$the$money`? 

The concern I have is that improving security of client secrets is a great idea but want to make sure users of this new feature don't assume that it is a complete solution when it isn't for existing cleartext secrets. I'd rather have it documented that it will fail on all existing unhashed secrets. I suggest that you let `check_password` fail on an unhashed secret, keeping the code simple, and not worry about backwards compatibility but instead add a further improvement: 

Rather than putting hashing of cleartext secrets in a migration and having it silently hash old cleartext secrets -- since the hashes would not be reversable -- add a management command called something like `hash_secrets` or `upgrade_secrets` which runs through the `oauth2_provider.applications` and hashes any cleartext secrets.

To determine what is or is not cleartext is still potentially an issue of the three '$' match but the management command and user documentation could flag those concerns, perhaps using a regex pattern or list of configured algorithm names.

Also, anytime someone uses the UI to edit an existing application's client_secret, it would get hashed when stored with whatever the current hasher is.

The documentation and application add UI would also need to be enhanced to remind the user to grab a copy of the client secret before hitting Save. This is consistent with how most systems I'm aware of work: You get one chance to grab the secret.

This approach would result in great a security upgrade as it won't let old, stale cleartext secrets that are in the database forever become an exposure.

After a while of running with this feature, me might even consider deprecating cleartext secrets and having a future major release with a breaking change to require hashing.

pkarman: I clearly approached this feature implementation with a naive understanding of how this library is used, especially with regard to how client credentials are managed via the UI. My original motivation was a "quick fix" as a result of evaluating the library for a $work project, and anticipating the security review would flag the plaintext secrets as a blocker.

I agree that a more mature feature implementation should be opinionated about the hashing (always require it) and make the necessary UI adjustments to ensure hashing is done consistently for all secrets. Such an implementation scope is beyond what I can commit to doing right now, especially as I don't use this library. So I'm going to create a PR to revert the changes on master. I have a branch on my fork that does the simpler path you suggested, without the data migration or administrative support. If I ever time, and it hasn't been finished, I'd be happy to return to it, but in the meantime, I assume you'd prefer not to have a half-baked implementation in the `master` branch.

Thanks for your time and review -- sorry not to be able to see it through to the end right now.

n2ygk:
OK, sorry to hear that but I understand.

PR #1093, opened 2022-01-20
WIP: Hash application client secrets using Django password hashing

### LINKED ISSUE #729: Why are secrets in plain text?
Hello all,
I was digging in the data and I found out that all the secrets are stored in plain text in the DB. 

Why so?

reading this https://github.com/jazzband/django-oauth-toolkit/issues/276 it seems that the client_secret is used to generate/calculate some value, which one precisely? so for that, there's no way to hash it.

However, what about access_token and refresh tokens? why are not they hashed? in the end, they should be, somehow, as a password, so the real value is not needed (although you will have access to the real value each request if needed).

n2ygk:
Fixes #729

## Description of the Change
Pick up where #1020 left off and simplify to always hash the client_secret using the Django default hasher.
The migration is not reversible as it hashes cleartext application client secrets.

## Checklist

- [x] unit-test added
- [x] documentation updated


n2ygk:
@pkarman - Thanks for the idea regarding client secret hashing. I've simplified by turning this into a breaking change (migration not reversible) that will always hash the client secret.

tests/test_client_credential.py
```diff
@@ -55,35 +57,28 @@ def tearDown(self):
...
-        }
-        request = self.factory.get("/fake-resource", **auth_headers)
-        request.user = self.test_user
-
-        view = ResourceView.as_view()
-        response = view(request)
-        self.assertEqual(response, "This is a protected resource")
+        # secret mismatch should return a 401
+        auth_headers = get_basic_auth_header(self.application.client_id, "not-the-secret")
+        response = self.client.post(reverse("oauth2_provider:token"), data=token_request_data, **auth_headers)
```
Andrew-Chen-Wang, line 74: I'd still prefer using RequestFactory for speed of testing and more manageable configuration. Only reason I'd use client is for middleware; is there middleware enabled here?

If we're testing for a 401 through an exception, just use `pytest.raises`?

tests/test_introspection_view.py
```diff
@@ -301,9 +304,7 @@ def test_view_post_valid_client_creds_basic_auth(self):
 
     def test_view_post_invalid_client_creds_basic_auth(self):
         """Must fail for invalid client credentials"""
-        auth_headers = get_basic_auth_header(
-            self.application.client_id, self.application.client_secret + "_so_wrong"
-        )
+        auth_headers = get_basic_auth_header(self.application.client_id, CLEARTEXT_SECRET + "_so_wrong")
```
Andrew-Chen-Wang, line 307:
```suggestion
        auth_headers = get_basic_auth_header(self.application.client_id, f"{CLEARTEXT_SECRET}_so_wrong")
```

tests/test_introspection_view.py
```diff
@@ -340,7 +341,7 @@ def test_view_post_invalid_client_creds_plaintext(self):
             {
                 "token": self.valid_token.token,
                 "client_id": self.application.client_id,
-                "client_secret": self.application.client_secret + "_so_wrong",
+                "client_secret": CLEARTEXT_SECRET + "_so_wrong",
```
Andrew-Chen-Wang, line 344:
```suggestion
                "client_secret": f"{CLEARTEXT_SECRET}_so_wrong",
```

CHANGELOG.md
```diff
@@ -14,7 +14,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Security
   -->
 
-## [Unreleased]
+## [2.0.0] unreleased
+
+### Changed
+* #1093 Changed to implement [hashed](https://docs.djangoproject.com/en/stable/topics/auth/passwords/)
```
Andrew-Chen-Wang, line 20:
```suggestion
* #1093 (**Breaking**) Changed to implement [hashed](https://docs.djangoproject.com/en/stable/topics/auth/passwords/)
```

I understand it already said "Breaking change" but I typically like it when I can see it immediately. Just personal opinion and no need to merge.

Andrew-Chen-Wang (changes requested):
In general, it LGTM!

tests/test_client_credential.py
n2ygk: My goal was to do the minimum needed to get what was started and reverted in #1020 simplified and merged. Cleaning up these tests and making them consistent is a much larger issue that I'm not comfortable with tackling; there are something like 3 different styles used.

A (series of) PR(s) addressing refactoring the test suite would be appreciated!

Andrew-Chen-Wang (approved):

MT-Cash:
If feels wrong to see "WIP" in the title of a merged PR / to see WIP in a breaking change in the changelog.

Issue #971, opened 2021-04-30
Allow "urn:ietf:wg:oauth:2.0:oob:auto" as a redirect_uri in the Aplication admin


manelclos:
Hi, the `Application` model admin is not currently allowing to set `redirect_uris` to `urn:ietf:wg:oauth:2.0:oob:auto`.

The `Application` model `clean` method does not take it into account, currently only allowing `http` and `https` schemes.

It would be great if this could be improved. Would a patch be accepted? or is this due to a design decision?

Also, trying to access the `authorize` url fails by default, we had to manually add the `crsf_exempt` decorator, as the `AuthorizationView` class is not decorated with `csrf_exemp` as the `TokenView` and `TokenRevokeView` are. Not a problem, but unexpected in the JSON use case.

vector-kerr:
+1 to this request - if my understanding is correct, this would allow for native app URIs with registered custom schemes.

manelclos:
Hi @vector-kerr, we are using this to get an authorization code from the authorize endpoint from a SPA, this is, no navigation to the authorize endpoint, no user confirmation.

PoDuck:
Is this not already implemented in the setting `ALLOWED_REDIRECT_URI_SCHEMES` as documented [here](https://django-oauth-toolkit.readthedocs.io/en/latest/settings.html?highlight=settings#allowed-redirect-uri-schemes)?

dmvol81:
> Hi, the `Application` model admin is not currently allowing to set `redirect_uris` to `urn:ietf:wg:oauth:2.0:oob:auto`.
> 
> 
> 
> The `Application` model `clean` method does not take it into account, currently only allowing `http` and `https` schemes.
> 
> 
> 
> It would be great if this could be improved. Would a patch be accepted? or is this due to a design decision?
> 
> 
> 
> Also, trying to access the `authorize` url fails by default, we had to manually add the `crsf_exempt` decorator, as the `AuthorizationView` class is not decorated with `csrf_exemp` as the `TokenView` and `TokenRevokeView` are. Not a problem, but unexpected in the JSON use case.
> 
> 
> 
> 

I knew nothing of what's going on. This is pure BS.

n2ygk:
> I knew nothing of what's going on. This is pure BS.

Not sure what you mean by this. This appears to have been solved at https://github.com/jazzband/django-oauth-toolkit/issues/971#issuecomment-948779327

Feel free to reopen if not.

[...]

PR #1097, opened 2022-01-23
Accept oob values as valid redirect URIs

danlamanna:
This addresses [this comment](https://github.com/jazzband/django-oauth-toolkit/issues/971#issuecomment-1016139005) made in #971.

## Description of the Change

The out of band (oob) protocol is supported within django-oauth-toolkit but paradoxically can't be configured using the forms, either in the Django admin or in the views provided. The tests test oob by creating an Application at the model layer, bypassing the typical validation. Because the allowed oob values don't look like a URI, they had to be shoehorned into the URI validator with a short circuit return, and `urn` had to be added as an "always valid" scheme.

This is a little bit hackier than I'd like, however oob is still widely used and solves a valuable niche in OAuth (CLI applications). This was the least hacky patch I could come up with to maintain backwards compatibility.

## Checklist

- [X] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,6 +21,7 @@ Ash Christopher
 Asif Saif Uddin
 Bart Merenda
 Bas van Oostveen
+Dan LaManna
 Dave Burkholder
 David Fischer
 David Smith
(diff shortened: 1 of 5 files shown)


n2ygk (commented):
oob support in DOT is slated to be deprecated per #1104. Google documents an alternative approach [here](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html?m=1#disallowed-oob) including an [end date](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html?m=1#dates-oob) of October 3, 2022 when they will no longer support oob. As of February they've already stopped allowing new oob clients.

n2ygk:
Will not add more oob.

PR #1108, opened 2022-01-29
OIDC: Standard scopes to determine which claims are returned

### LINKED ISSUE #1107: OIDC: 5.4.  Requesting Claims using Scope Values

Per the [OIDC spec 5.4](https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims) one should be able to use OIDC standard scopes to limit which additional claims are returned in the ID Token and Userinfo response. The current implementation is unable to filter which additional claims to return in a reasonable way.

The ability to extend `get_oidc_claims()` to use the granted access token scopes. This does not work because upstream oauthlib [does not pass the correct request.scopes](https://github.com/oauthlib/oauthlib/issues/799) instead hardcoding the scope list as `["openid"]`

A small fix to `get_userinfo_claims()` will make the `access_token.scopes` available. PR coming soon;-)

I've prototyped this [here](https://github.com/n2ygk/dot-tutorial/blob/main/my_oidc/oauth_validator.py).

**Additional context**

I believe the "right" way to get access to the scopes might require a fix to oauthlib, but the workaround above seems to be a pretty good start.

n2ygk:
Fixes #1107

## Description of the Change

Only return standard OIDC claims that are permitted by the granted OIDC scopes (profile, email, address, etc.)

## Checklist

- [x] unit-test added
- [x] documentation updated


oauth2_provider/oauth2_validators.py
```diff
@@ -760,7 +788,13 @@ def get_oidc_claims(self, token, token_handler, request):
         claims = {}
 
         for k, v in data.items():
-            claims[k] = v(request) if callable(v) else v
+            if not self.oidc_claim_scope or (
+                token
+                and hasattr(token, "scopes")
```
n2ygk, line 793: I'm not super-happy about having to test for token and token.scopes. It's not clear if this is a real possibility or if the test cases aren't providing enough of a dummy token.

n2ygk:
TODO: Need to confirm if the claims parameter is actually supported as stated in the [docs](https://github.com/jazzband/django-oauth-toolkit/blob/master/docs/oidc.rst#adding-claims-to-the-id-token). The openid-configuration does not assert `"claims_parameter_supported". Per the [spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) this means it does: "...if omitted, the default value is false."

Andrew-Chen-Wang:
Are you still looking to merge upstream and close this?

n2ygk:
> Are you still looking to merge upstream and close this?

Maybe merge here and then look at what upstream stuff can be done and remove this and other stuff that should be upstream later? There's other stuff like oidc metadata that can also potentially go upstream.

Andrew-Chen-Wang (approved):
sure. lgtm as far as I can tell

PR #1113, opened 2022-02-09
Topic/1112

daffyd:
Fixes # 
1112

## Description of the Change
Modifying the tutorial to suggest how to extend the rest_framework class based views tutorial to use oauth-toolkit.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/tutorial/tutorial_03.rst
+++ b/docs/tutorial/tutorial_03.rst
@@ -104,9 +104,9 @@ The key is setting a class attribute to override the default *permissions_classe
         ...
         permission_classes = [TokenHasReadWriteScope]
 
-Note that this example overrides the Django default permission class setting. There are several other 
+Note that this example overrides the Django default permission class setting. There are several other
 ways this can be solved. Overriding the class function *get_permission_classes* is another way
-to solve the problem. 
+to solve the problem.
 
 A detailed dive into the `Dango REST framework permissions is here. <https://www.django-rest-framework.org/api-guide/permissions/>`_
 
--- a/docs/tutorial/tutorial_03.rst
+++ b/docs/tutorial/tutorial_03.rst
@@ -103,3 +103,10 @@ The key is setting a class attribute to override the default *permissions_classe
     class SnippetDetail(generics.ListCreateAPIView):
         ...
         permission_classes = [TokenHasReadWriteScope]
+
+Note that this example overrides the Django default permission class setting. There are several other 
+ways this can be solved. Overriding the class function *get_permission_classes* is another way
+to solve the problem. 
+
+A detailed dive into the `Dango REST framework permissions is here. <https://www.django-rest-framework.org/api-guide/permissions/>`_
+
--- a/docs/tutorial/tutorial_03.rst
+++ b/docs/tutorial/tutorial_03.rst
@@ -86,16 +86,15 @@ If you have completed the `Django REST framework tutorial
 <https://www.django-rest-framework.org/tutorial/3-class-based-views/#using-generic-class-based-views>`_,
 you will be familiar with the 'Snippet' example, in particular the SnippetList and SnippetDetail classes.
 
-It would be really nice to reuse those views, but also support token handling. Instead of reworking
+It would be nice to reuse those views **and** support token handling. Instead of reworking
 those classes to be ProtectedResourceView based, the solution is much simpler than that.
 
-Assuming you have already modified the settings as was already shown.
-
-The key is setting a class variable to override the default *permissions_classes* with something that will use our :term:`Access Token` properly.
+Assume you have already modified the settings as was already shown.
+The key is setting a class attribute to override the default *permissions_classes* with something that will use our :term:`Access Token` properly.
 
 .. code-block:: python
 
-    from django.contrib.auth.decorators import login_required
+    from oauth2_provider.contrib.rest_framework import TokenHasReadWriteScope
 
     class SnippetList(generics.ListCreateAPIView):
         ...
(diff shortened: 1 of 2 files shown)


n2ygk:
Thanks for this first PR. Hopefully one of the review team will take a look shortly.

docs/tutorial/tutorial_03.rst
```diff
@@ -78,3 +78,29 @@ Now supposing your access token value is `123456` you can try to access your aut
...
+-----------------------------------------------------
+
+If you have completed the `Django REST framework tutorial
+<https://www.django-rest-framework.org/tutorial/3-class-based-views/#using-generic-class-based-views>`_,
+you will be familiar with the 'Snippet' example, in particular the SnippetList and SnippetDetail classes.
+
+It would be really nice to reuse those views, but also support token handling. Instead of reworking
+those classes to be ProtectedResourceView based, the solution is much simpler than that.
+
+Assuming you have already modified the settings as was already shown.
```
Andrew-Chen-Wang, line 92:
```suggestion
Assume you have already modified the settings as was already shown.
```

It might be better to combine this sentence with the next paragraph.

docs/tutorial/tutorial_03.rst
```diff
@@ -78,3 +78,29 @@ Now supposing your access token value is `123456` you can try to access your aut
...
+
+The key is setting a class variable to override the default *permissions_classes* with something that will use our :term:`Access Token` properly.
```
Andrew-Chen-Wang, line 94:
```suggestion
The key is setting a class attribute to override the default *permissions_classes* with something that will use our :term:`Access Token` properly.
```

docs/tutorial/tutorial_03.rst
```diff
@@ -78,3 +78,29 @@ Now supposing your access token value is `123456` you can try to access your aut
...
+
+.. code-block:: python
+
+    from django.contrib.auth.decorators import login_required
+
+    class SnippetList(generics.ListCreateAPIView):
+        ...
+        permission_classes = [TokenHasReadWriteScope]
```
Andrew-Chen-Wang, line 102: I believe this will override other default permission classes. Perhaps you can override the  `get_permission_classes` method

docs/tutorial/tutorial_03.rst
```diff
@@ -78,3 +78,29 @@ Now supposing your access token value is `123456` you can try to access your aut
...
+It would be really nice to reuse those views, but also support token handling. Instead of reworking
+those classes to be ProtectedResourceView based, the solution is much simpler than that.
+
+Assuming you have already modified the settings as was already shown.
+
+The key is setting a class variable to override the default *permissions_classes* with something that will use our :term:`Access Token` properly.
+
+.. code-block:: python
+
+    from django.contrib.auth.decorators import login_required
```
Andrew-Chen-Wang, line 98: Random import?

Also if login is required, then it is necessary not to override the class attribute and instead override the method.

docs/tutorial/tutorial_03.rst
```diff
@@ -78,3 +78,29 @@ Now supposing your access token value is `123456` you can try to access your aut
...
     curl -H "Authorization: Bearer 123456" -X GET http://localhost:8000/secret
+
+Working with Rest_framework generic class based views
+-----------------------------------------------------
+
+If you have completed the `Django REST framework tutorial
+<https://www.django-rest-framework.org/tutorial/3-class-based-views/#using-generic-class-based-views>`_,
+you will be familiar with the 'Snippet' example, in particular the SnippetList and SnippetDetail classes.
+
+It would be really nice to reuse those views, but also support token handling. Instead of reworking
```
Andrew-Chen-Wang, line 89:
```suggestion
It would be nice to reuse those views **and** support token handling. Instead of reworking
```

Andrew-Chen-Wang (changes requested):
Thanks for the PR!

daffyd:
Adding in the suggested changes.
And yes - copied the wrong import and of course my eyes saw the wrong thing
when I reviewed the result. sigh.

As far as overriding the default behavior of permission_classes - I am up
to alternate suggestions, but that was the only thing that worked for me. I
am using
application_c.GRANT_CLIENT_CREDENTIALS to create the application.
(maybe improperly.... as I don't set a user as this is m2m)
Certainly could suggest for more complex operations overriding
get_permission_classes() would make more sense ?
There is a lot of chatter on this I found while fighting the issue, and
this was the only suggestion that handled the case. But django/rest/oauth2
newbie here.

Andrew-Chen-Wang:
On my phone so not sure if this is the right suggestion, but to override get permission classes, you can just call super() and extend with the remaining permission classes to be added. Then return.
________________________________

docs/tutorial/tutorial_03.rst
daffyd: added

docs/tutorial/tutorial_03.rst
daffyd: changed

docs/tutorial/tutorial_03.rst
daffyd: doh - fixed

docs/tutorial/tutorial_03.rst
daffyd: fixed

docs/tutorial/tutorial_03.rst
daffyd: Yes - it overrides the default classes. And maybe there is a better way to do this, but it was the way that worked for me. 
Overriding  get_permission_classes would basically do the same thing with more code - unless you need more complicated logic. Again - open to suggestions.

daffyd: I added another paragraph that hopefully resolves your concern. There is so many ways to "solve" the issue of a newbie who starts with the Snippets tutorial. I like the "simple/fast" solution presented, but now have references additional information that will guide the reader.

Andrew-Chen-Wang (approved):
LGTM thanks!

PR #1115, opened 2022-02-13
Serve discovery view on both routes

### LINKED ISSUE #1089: Improve reach of the discovery endpoint
When using django-oauth-toolkit with default discovery options, the discovery URL is mounted under `OIDC_ISS_ENDPOINT` + `/.well-known/openid-configuration/`.

django-oauth-toolkit serves the discovery endpoint with a trailing slash an Django will automatically redirect requests without such trailing slash to the former with a HTTP 301 redirect.

We have observed clients to implement a single GET request and not to follow redirects. As a result the discovery process terminates with failure. I am proposing to implement both.

The following request will fail without additional options.
```sh
$ curl https://server.example.net/o/.well-known/openid-configuration
$
```

Expected behavior:
```sh
$ curl https://server.example.net/o/.well-known/openid-configuration
{"issuer": "https://server.example.net/o"...}
```

Create an additional url pattern in my stack where the non-trailing slash is also served.

**Additional context**
The [OpenID Connect standard](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) has registered the URI `/.well-known/openid-configuration` as per [RFC5785](https://datatracker.ietf.org/doc/html/rfc5785#section-3). The URI-encoded application name is a `segment-nz` as defined in [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).

The standard defines a URL without trailing slash as the entry point. As such OIDC application developers construct those and issue a GET request which may or may not follow redirects (unspecified in the standard). I would propose to implement the canonical form where no redirects are required and the same content is served by the application.

Some working examples on commercial providers:
* https://accounts.google.com/.well-known/openid-configuration
* https://appleid.apple.com/.well-known/openid-configuration
 https://appleid.apple.com/.well-known/openid-configuration/
* https://cognito-idp.region.amazonaws.com/userpool-id/.well-known/openid-configuration
 https://cognito-idp.region.amazonaws.com/userpool-id/.well-known/openid-configuration/
* https://custom-domain.auth0.com/.well-known/openid-configuration
* https://custom-domain.okta.com/.well-known/openid-configuration
 https://custom-domain.okta.com/.well-known/openid-configuration/
* https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration
 https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration/
* https://www.facebook.com/.well-known/openid-configuration
 https://www.facebook.com/.well-known/openid-configuration/
* [Forgerock /openam/oauth2/realms/root/realms/subrealm1/.well-known/openid-configuration](https://backstage.forgerock.com/docs/am/7/oidc1-guide/rest-api-oidc-discovery-configuration.html)
* [Gluu /.well-known/openid-configuration](https://gluu.org/docs/gluu-server/4.0/admin-guide/openid-connect/)
* [Keycloack /realms/{realm-name}/.well-known/openid-configuration](https://www.keycloak.org/docs/latest/securing_apps/#endpoints)

timptner:
Serving the discovery view on both routes enables usage of third-party clients, which only request the view without a trailing slash and doesn't follow a 301 redirect.

Fixes #1089

## Description of the Change

Making the trailing slash optional allows serving the ConnectDiscoveryInfoView on both routes. The oidc guideline doesn't specify if 301 redirects should be followed and the default route to access a discovery info is without the trailing slash. Since Django defaults to redirecting all routes returning a 404 to its corresponding route with an appended slash (typically for a folder) one can circumstance this behavior by responding to the specific route.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/urls.py
+++ b/oauth2_provider/urls.py
@@ -32,7 +32,7 @@ management_urlpatterns = [
 
 oidc_urlpatterns = [
     re_path(
-        r"^\.well-known/openid-configuration/$",
+        r"^\.well-known/openid-configuration/?$",
         views.ConnectDiscoveryInfoView.as_view(),
         name="oidc-connect-discovery-info",
     ),


n2ygk (changes requested):
Hey - Thanks for this. 

Please complete the rest of the checklist:

-  unit-test added: Please add a test to `test_oidc_views.py` to demonstrate that both with and without trailing slash works.
-  documentation updated
-  CHANGELOG.md updated (only for user relevant changes)
-  author name in AUTHORS

PR #1118, opened 2022-02-16
Prevent the tests/migrations directory from getting packaged

brianhelba:
Simply excluding `tests` is not enough to prevent subpackages from also being excluded. If a wildcard is provided to `find_packages(exclude=`, it will only match a single hierarchy level.

Without this, a directory `site-packages/tests/migrations` gets created upon package install.

## Checklist
- [ ] ~unit-test added~
- [ ] ~documentation updated~
- [ ] ~`CHANGELOG.md` updated (only for user relevant changes)~

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,6 +21,7 @@ Ash Christopher
 Asif Saif Uddin
 Bart Merenda
 Bas van Oostveen
+Brian Helba
 Dave Burkholder
 David Fischer
 David Smith
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,4 +38,6 @@ install_requires =
 	jwcrypto >= 0.8.0
 
 [options.packages.find]
-exclude = tests
+exclude =
+	tests
+	tests.*


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #1108 OIDC: Fix `validate_bearer_token()` to properly set `request.scopes` to the list of granted scopes.
+* Miscellaneous packaging issues.
```
n2ygk, line 38:
```suggestion
```
Probably not necessary to explicitly say this as it doesn't change the user-visible functionality.

n2ygk (dismissed):
Thanks for this! No need for the CHANGELOG update.

CHANGELOG.md
brianhelba: @n2ygk Fixed.

n2ygk (approved):

Issue #1114, opened 2022-02-11
Dependency on celery lacks documentation


sumpfralle:
**Describe the bug**

Recently (commit a6a21d3566d634e1bc016c3146fe1560469d9813) a dependency on celery was introduced. Probably it was meant to be triggered only, if celery is active (which would cause the import of `tasks.py`). But in reality it is also triggered, if other task queue handlers are running (in my case: [huey](https://github.com/coleifer/huey)).

**To Reproduce**
Add `oauth2_provider` and `huey` to `INSTALLED_APPS` and ensure, that the `celery` module is not available.
`python3 manage.py runserver` will fail due to the missing `celery` module.

**Expected behavior**
I could imagine the following behaviors to be more desirable:
* A) a missing `celery` module is silently ignored
* B) `celery` is only imported, if `celery` is found in `settings.INSTALLED_APPS`
* C) the *active* task queue manager (e.g. `celery` or `huey`) is detected (e.g. by inspecting `settings.INSTALLED_APPS`) and the relevant decorator is selected

**Version**
1.7.0

- [x] I have tested with the latest published release and it's still a problem.
- [ ] I have tested with the master branch and it's still a problem.

**Additional context**

n2ygk:
Thanks for this report. Can you submit a PR which would fix this?

sumpfralle:
To be honest, I cannot tell, what the original commit is supposed to do exactly.

I *think*, it is supposed to wrap the function in order to enforce its execution in a task queue.
But I have no idea, how the task is actually *scheduled*.

In [huey](https://huey.readthedocs.io/en/latest/django.html) there are two types of function decorators:
* `db_task` (no parameters): for tasks to be called explicitly by code from within the django application
* `db_periodic_task` (parameters specify the wanted period): for tasks, which should be executed periodically (they are not called explicitly)

The original commit, which introduced the `celery` dependency (a6a21d3566d634e1bc016c3146fe1560469d9813) references the [celery documentation for periodic tasks](https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html), but the `shared_task` decorator is not described there. Thus I guess, that the `shared_task` decorator is similar to the `db_task` decorator in huey (see above), i.e. it is *not* executed periodically.

In this case I do not understand the purpose of the current snippet. Maybe there is something missing?

@Natureshadow: maybe you could clarify, how exactly your addition should be used?

MichaelVoelkel:
[...]

PR #1123, opened 2022-03-18
fix(tasks): fix error caused by relative import

alexei:
## Description of the Change

Running `oauth2_provider.tasks.clear_tokens` results in an error e.g.:
```python
>>> from oauth2_provider.tasks import clear_tokens
>>> clear_tokens()
Traceback (most recent call last):
  File "[python3.9]/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "[site-packages]/celery/local.py", line 188, in __call__
    return self._get_current_object()(*a, **kw)
  File "[site-packages]/celery/app/task.py", line 392, in __call__
    return self.run(*args, **kwargs)
  File "[site-packages]/oauth2_provider/tasks.py", line 6, in clear_tokens
    from ...models import clear_expired  # noqa
ImportError: attempted relative import beyond top-level package
```

This update fixes the import path.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/tasks.py
+++ b/oauth2_provider/tasks.py
@@ -3,6 +3,6 @@ from celery import shared_task
 
 @shared_task
 def clear_tokens():
-    from ...models import clear_expired  # noqa
+    from oauth2_provider.models import clear_expired
 
     clear_expired()


n2ygk (commented):
I don't quite understand what is broken and that you are fixing. That task works as designed when using [celery-beat](https://github.com/celery/django-celery-beat). It works fine relative as it is part of the package. If you are not using celery-beat you should never invoke that code but instead directly call the [cleartokens](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/management/commands/cleartokens.py) management command or invoke the [clear_expired function](https://github.com/jazzband/django-oauth-toolkit/blob/c48b7516fe6aebf89c5964e4855ecf129274c3c3/oauth2_provider/models.py#L643)

n2ygk:
Closing as not applicable.

alexei:
I first saw this failing in production running inside celery. From Sentry:

Screenshot 2022-03-18 at 20 28 56

My example was to illustrate what's going on. `clear_tokens` is a _normal_ function, so I see no reason why it should fail when called inside the shell.

Regardless, the reason why the import is failing is because `clear_expired` is imported from `...models`, but `tasks` and `models` are on the same level. At worst, the path would be `.models` i.e. `from .models import clear_expired`.

n2ygk:
clear_tokens works as designed for use with celery-beat. I've tested it as such even though I don't use celery. It is not "regular" function as it is preceded by the celery `@shared_task` decorator. It is not meant to be invoked directlyl.

alexei:
@n2ygk why are you arguing it works as indended when I just offered proof that it doesn't, and even upon reviewing the code it becomes clear the import can't possibly work since it looks up the function outside of the package?

n2ygk:
> @n2ygk why are you arguing it works as indended when I just offered proof that it doesn't, and even upon reviewing the code it becomes clear the import can't possibly work since it looks up the function outside of the package?

Good question. So how come it worked for me and the original contributor @Natureshadow? Can you please help me out here and document how this is supposed to be configured with celery? I had asked @Natureshadow to do so but he just provided a link so I configured celery in a test app as best I could and didn't see any errors and that the task showed up in the django admin intervface. I suspect that the [@shared_task](https://github.com/celery/celery/blob/25ca389039f667cdf940c9efe44bdefd1cf70f30/celery/app/__init__.py#L24) decorator is changing the relative path location somehow but it's a bit over my head when reading it.

Perhaps you can share a full sample app configuration that demonstrates the problem since it is not reproducible given this report alone and I frankly don't have unlimited time or knowledge to debug this. Cutting and pasting an error message without context is not enough.

Given the other conflict with Huey #1114 I'm seriously considering reverting #1070 and leaving it up to someone to document how to add a Celery or Huey autoloaded tasks rather than including it in this package.

alexei:
> Good question. So how come it worked for me and the original contributor @Natureshadow?

I don't think it did, really. I don't think you're liars either, but I checked https://github.com/jazzband/django-oauth-toolkit/pull/1070 and I see no proof it ever worked. Both you and @Natureshadow confirm the task is discovered, but none of you attested it actually ran. For example, at one point you're [saying](https://github.com/jazzband/django-oauth-toolkit/pull/1070#issuecomment-1015804567):

> And see the admin interface and was able to autodiscover the task. I didn't get it completely working (econnrefused) but close enough to understand what I needed to.

---

Detective work aside, in my experience with Celery I'd say simply importing the task and running it inside the shell ought to do it. [unbound] celery tasks are regular functions and can be called directly like any other function.

Initially I started writing a smoke test, but I was faced with the choice of adding celery to the requirements or manipulating `sys.modules`, so I abandoned the idea. I agree to the point made by @Natureshadow in https://github.com/jazzband/django-oauth-toolkit/pull/1070#issuecomment-1006073286:

>  I don't think it is worth adding that complexity to the test suite

But an alternative that would work as a smoke test would be to move the import statement at the top of the file. If there was an issue it would fail a lot sooner.

---

I don't have experience with Huey. Regardless, personally I'm ok with reverting the original commit if it's causing problems.

n2ygk:
> > Good question. So how come it worked for me and the original contributor @Natureshadow?
> 
> I don't think it did, really. I don't think you're liars either, but I checked #1070 and I see no proof it ever worked. Both you and @Natureshadow confirm the task is discovered, but none of you attested it actually ran. For example, at one point you're [saying](https://github.com/jazzband/django-oauth-toolkit/pull/1070#issuecomment-1015804567):

You will note that the import statement is *inside* the function definition not top-level in the tasks.py file. As such, it is not executed in the context of importing tasks.py but in the context of the decorator function for which the function definition is an argument.

alexei:
@n2ygk I reported a bug, showed proof of it and offered instructions to reproduce it, showed there was no proof that it ever worked, and offered to fix it. But rather than listening and possibly executing the code, it seems to me you're looking to argue and trying to be right or something like that. Whatever it is, it's not helping advance this conversation and not making the celery task run either. As such, I'll take a step back until we can have a reasonable conversation. Do let me know if you’re willing to do that.

Luckily, in my project I'm not using celery’s autodiscover feature, so I'll set up a custom task that imports `clear_expired` from `oauth2_provider.models`, where it actually lives, not the root of the virtual env like the builtin `clear_tokens` does.

n2ygk:
@alexei I asked you to provide an example that reproduces the bug in context. You did not.

alexei:
@n2ygk I showed you a screenshot with the error in production running in celery beat, and offered simple steps to reproduce it in the shell. Calling an unbound celery task directly is reasonable use. That said, I think I did my part of the deal. Besides, the error literally says you're importing from the outside of the package and the extra dots confirm it. If you don't connect the dots, there's nothing I can do about it.

n2ygk:
@alexei With **tremendous amounts of egg on my face**, I have connected the dots and apologize profusely. It seems @Natureshadow supplied an untested PR #1070.  I was sticking by his word over yours. That was a mistake.

I've configured Celery and reproduced the error and applied your fix and have documented how to use it [here](https://github.com/n2ygk/dot-tutorial/blob/a3cb8ae62772b11e078b82e21c05edfe31021e95/README.md#add-celery-automated-task-discovery).

I'll be reverting #1070 and documenting how to do instead add the `clear_tokens` task into an app that uses DOT.

alexei:
Apology accepted. I appreciate your looking into this. 

What do you think about documenting the `clear_expired` function in the [official docs](https://django-oauth-toolkit.readthedocs.io/)? My thinking is, the `cleartokens` management command is documented, but that's only useful in a setup where one's using cron for periodic tasks. For those using Celery, Huey etc. that's not really useful, and sadly Django doesn't provide a mechanism to execute management commands programatically. Sure enough, writing a task to call `clear_expired` is trivial for either Celery or Huey, but a newcomer reading the official docs is unaware of it unless they check out the [source](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/management/commands/cleartokens.py) or the [DOT tutorial](https://github.com/n2ygk/dot-tutorial) which btw I didn't know existed before this.

n2ygk:
> Apology accepted. I appreciate your looking into this. 
> 
> 
> 
> What do you think about documenting the `clear_expired` function in the [official docs](https://django-oauth-toolkit.readthedocs.io/)? My thinking is, the `cleartokens` management command is documented, but that's only useful in a setup where one's using cron for periodic tasks. For those using Celery, Huey etc. that's not really useful, and sadly Django doesn't provide a mechanism to execute management commands programatically. Sure enough, writing a task to call `clear_expired` is trivial for either Celery or Huey, but a newcomer reading the official docs is unaware of it unless they check out the [source](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/management/commands/cleartokens.py) or the [DOT tutorial](https://github.com/n2ygk/dot-tutorial) which btw I didn't know existed before this.

See #1128

n2ygk:
@alexei PS: that dot-tutorial is just my sandbox. As you can see now at https://django-oauth-toolkit--1128.org.readthedocs.build/en/1128/tutorial/tutorial.html I've attempted to document this.

alexei:
Both look fine by me 👍

PR #1173, opened 2022-06-12
Add required_scopes view decorator for function-based views

TurnrDev:
## Description of the Change

Adds a decorator for adding the required_scopes parameter to function-based views for permission classes such as [TokenHasScope](https://github.com/jazzband/django-oauth-toolkit/blob/40b0de1df9000136e4a3ee8c81c027aca0b9def5/oauth2_provider/contrib/rest_framework/permissions.py#L14)

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -79,3 +79,4 @@ Dominik George
 David Hill
 Darrel O'Pry
 Jordi Sanchez
+Jay Turner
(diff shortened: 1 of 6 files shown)


TurnrDev:
Despite confirming that the required_scopes attribute is on the view with vars(view), it seems that when `TokenHasScope` touches it, it does not have it. Not sure what's causing it, but until I figure it out, this will stay a draft.

TurnrDev:
Seems this is blocked by Django Rest Framework's WrappedAPIView, which converts a function-based view to a class-based view. Without rewriting the api_view decorator, I won't get this to work and I don't fancy going to that much effort. (And the effort of maintaining it.)

PR #1175, opened 2022-06-19
Release 2.1.0

n2ygk:
Fixes #

## Description of the Change

Release 2.1.0

It's past due time to get this out there for the hashed migrations fix especially.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,13 +16,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [2.1.0] 2022-06-19
+
+### WARNING
+
+Issues caused by **Release 2 breaking changes** continue to be logged. Please **make sure to carefully read these release notes** before
+performing a MAJOR upgrade to 2.x.
+
+These issues both result in `{"error": "invalid_client"}`:
+
+1. The application client secret is now hashed upon save. You must copy it before it is saved. Using the hashed value will fail.
+
+2. `PKCE_REQUIRED` is now `True` by default. You should use PKCE with your client or set `PKCE_REQUIRED=False` if you are unable to fix the client.
+
 ### Added
-* Support `prompt=login` for the OIDC Authorization Code Flow end user [Authentication Request](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest).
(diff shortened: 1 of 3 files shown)


n2ygk:
@Andrew-Chen-Wang pinging you since you seem to be the only one helping out here;-)

CHANGELOG.md
```diff
@@ -16,13 +16,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [2.1.0] 2022-06-19
+
+### WARNING
+
+Issues caused by **Release 2 breaking changes** continue to be logged. Please **make sure to carefully read these release notes** before
```
Andrew-Chen-Wang, line 23:
```suggestion
Issues caused by **Release 2.0.0 breaking changes** continue to be logged. Please **make sure to carefully read these release notes** before
```

CHANGELOG.md
```diff
@@ -16,13 +16,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [2.1.0] 2022-06-19
```
Andrew-Chen-Wang, line 19: We prob don't want to bump major v, but I highly recommend you add a setting saying "USE_HASHED_VALUES=True" or something like that and set that to false by default. Users cannot migrate unless that setting value is true. If not, raise an error during migration.

I've had poor experiences with breaking changes being hidden in minor versions, and many people don't unit test this pkg , so highly recommend

Andrew-Chen-Wang (changes requested):

CHANGELOG.md
n2ygk: The breaking change happened with 2.0.0 with hashing and PKCE required. There's no breaking change w.r.t hashing here -- a bugfix for migrations which would have been a patch release (2.0.1) but then I included a new backwards-compatible feature (prompting) which resulted in 2.1.0. 

Given the repeated issues raised with `{"error": "invalid_client"}` caused generally by not using the cleartext client secret or not using PKCE I decided to update the README in the hopes that people would see that before raising the issue.

Andrew-Chen-Wang (approved):
got it sounds good!

PR #1177, opened 2022-06-21
Added list_select_related to reduce duplicate SQL queries in admin UI

phith0n:
## Description of the Change

Added `list_select_related` in the admin `IDTokenAdmin` to avoid duplicate SQL queries. 

Before adding the `list_select_related`, every 100 items in the `IDToken` 206 SQL queries is executed:

image

After adding the `list_select_related`, every 100 items in the `IDToken` list 6 SQL queries is executed:

image

The total page loading time is 7305 milliseconds and 698 milliseconds respectively in my computer. It expects to reduce 10x the time.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -79,3 +79,4 @@ Dominik George
 David Hill
 Darrel O'Pry
 Jordi Sanchez
+Owen Gong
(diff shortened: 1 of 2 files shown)


n2ygk (approved):
Thanks for this!

PR #1182, opened 2022-07-01
Add 'code_verifier' parameter to token request

### LINKED ISSUE #1178: can't make tutorial "getting started" , django-oauth-toolkit 	2.0.0
hi good day

I can't make work the tutorial "getting started" from the docs
(sorry I know there are a couple of similars issues rised but still it seems there is some differences)

Django               	       4.0.5
django-oauth-toolkit 	2.0.0
OS					Ubuntu 20.04
python				3.8

I've been following the steps as appears in the docs but at the end the system delivers me the line

`{"error": "invalid_request", "error_description": "Code verifier required."}`

instead of the expected:

view8

These where my steps:

view1
view3
view4

BTW, when you enters the url
[http://127.0.0.1:8000/o/applications/register/](http://127.0.0.1:8000/o/applications/register/)

the form has another field called "Algorithm" with 3 options:
- No OIDC support
- RSA with SHA-2 256
- HMAC with SHA-2 256

Which option should I choose at this step? The docs do not refers to it.
I tried with the first and the third but the steps didn't work with any of them. The second option gives me this strange error when I try to selected it.

view2

Finally, and don't know if it was important, but when I tried to create the code_challenge, I had to change the line
`code_verifier = base64.urlsafe_b64encode(code_verifier)`
to
`code_verifier = base64.urlsafe_b64encode(code_verifier.encode('utf-8'))`
because I had an error that requests me to use a string.
Later the line
`code_challenge = hashlib.sha256(code_verifier.encode('utf-8')).digest()`
had to change to
`code_challenge = hashlib.sha256(code_verifier).digest()`

view7

What am I doing wrong?

(Thanks in advance)

porowns:
Fixes #1178

## Description of the Change
The getting started tutorial excludes the 'code_verifier' when fetching the access token. This is causing confusion, example https://github.com/jazzband/django-oauth-toolkit/issues/1178

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -49,6 +49,7 @@ Joseph Abrahams
 Jozef Knaperek
 Julien Palard
 Jun Zhou
+Kaleb Porter
 Kristian Rune Larsen
 Michael Howitz
 Paul Dekkers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   -->
 
 ## [unreleased]
+* Add 'code_verifier' parameter to token requests in documentation
 
 ## [2.1.0] 2022-06-19
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -305,7 +305,7 @@ Export it as an environment variable:
 
 Now that you have the user authorization is time to get an access token::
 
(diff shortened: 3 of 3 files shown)


n2ygk (changes requested):
Thanks for this!

Please add your name to AUTHORS along with a short description of the improvement to CHANGELOG (or I can do that).

n2ygk (approved):

PR #1191, opened 2022-08-04
Remove duplicate OAuthLibMixin from base classes

fdemmer:
## Description of the Change

The generic view classes `ProtectedResourceView` and `ClientProtectedResourceView` inherit from `ProtectedResourceMixin` and `ClientProtectedResourceMixin` respectively, which both inherit `OAuthLibMixin` already.

It seems redundant to me to have the views inherit `OAuthLibMixin` again.

If there is no reason I am missing, this PR fixes that.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -51,6 +51,7 @@ Eduardo Oliveira
 Egor Poderiagin
 Emanuele Palazzetti
 Federico Dolce
+Florian Demmer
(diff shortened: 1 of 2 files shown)


Andrew-Chen-Wang (approved):
thanks!

PR #1192, opened 2022-08-06
Fix tox env used for flake8 in contributing docs

fdemmer:
## Description of the Change

The tox environment to run flake8 changed in https://github.com/jazzband/django-oauth-toolkit/commit/6af081c8053dc9712cb4822f5c876d18269b7851.

## Checklist

- [x] documentation updated


rcmurphy (approved):
Thanks for the fix!

PR #1196, opened 2022-08-14
Split Migration 0004 into two migrations

abu-siddique-shipsy:
Citus Postgres DB doesnt allow adding of new field with constrains on it.  Example Error attached

You can issue each command separately such as ALTER TABLE oauth2_provider_accesstoken ADD COLUMN id_token data_type; ALTER TABLE oauth2_provider_accesstoken ADD CONSTRAINT constraint_name CHECK (check_expression);

Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/migrations/0004_auto_20200902_2022.py
+++ b/oauth2_provider/migrations/0004_auto_20200902_2022.py
@@ -45,6 +45,11 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='accesstoken',
             name='id_token',
+            field=models.IntegerField(),
+        ),
+        migrations.AlterField(
+            model_name='accesstoken',
+            name='id_token',
             field=models.OneToOneField(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='access_token', to=oauth2_settings.ID_TOKEN_MODEL),
         ),
         migrations.AddField(


dopry (changes requested):
Citus postgres isn't technically a supported database. That said I don't see a reason not to do our best to support it. 

However we need to document why we're doing this here. Please add comment explaining why, with links to Citus DB documentation that demonstrates what is not supported and the specific error that you get from Citus if this is not done in 2 steps.

n2ygk:
@abu-siddique-shipsy will you be doing this?

> Citus postgres isn't technically a supported database. That said I don't see a reason not to do our best to support it.
> 
> However we need to document why we're doing this here. Please add comment explaining why, with links to Citus DB documentation that demonstrates what is not supported and the specific error that you get from Citus if this is not done in 2 steps.

n2ygk:
@abu-siddique-shipsy Have you opened an issue against Citus that shows this error that makes it non-compliant with Django Postgres support? It seems like that's where the appropriate compatibility fix should live given that DOT is not the only place where migrations like 0004 happen.

Also you say "Example Error attached" above but I don't see the attachment.

n2ygk:
closing as stale. Feel free to reopen with the requested changes.

PR #1039, opened 2021-12-14
Add support for Dj40, drop Py36 and Dj31

Andrew-Chen-Wang:
Fixes #1037 

## Description of the Change

Add support for Dj40, drop Py36 and Dj31

Dj31 and Py36 will be EOL end of this December of 2021. I do not want to hold back any PRs that need to support old versions of stuff, thus I have dropped those two.

## Checklist

- [X] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [ ] documentation updated
- [ ] `CHANGELOG.md` updated (only for user relevant changes)
- [X] author name in `AUTHORS`

PR #1203, opened 2022-09-05
Test Django 4.1 and remove upper version bound

### LINKED ISSUE #1202: Incorrect upper bound prevents use of Django 4.1.1+
**Describe the bug**

setup.cfg constrains the maximum Django version:

```
install_requires =
	django >= 2.2, <= 4.1
```

With this upper bound it disallows today's [Django 4.1.1 bugfix release](https://www.djangoproject.com/weblog/2022/sep/05/bugfix-release/0, and later security releases to come.

Upper bound version constraints are highly discouraged by many - see [this blog post](https://iscinumpy.dev/post/bound-version-constraints/). They introduce unnecessary slowness and blocking into the softwawre ecosystem.

I recommend dropping the upper bound entirely.

**To Reproduce**

```
$ pip install django==4.1.1 django-oauth-toolkit==2.1.0
...
ERROR: Cannot install django-oauth-toolkit==2.1.0 and django==4.1.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested django==4.1.1
    django-oauth-toolkit 2.1.0 depends on django<=4.1 and >=2.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

**Expected behavior**

Allow upgrades of Django

**Version**
2.1.0

- x ] I have tested with the master branch and it's still a problem.

**Additional context**
n/a

adamchainz:
Fixes #1202

## Description of the Change

Remove upper version bound, as I recommended on the ticket.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,7 @@ classifiers =
 	Framework :: Django :: 2.2
 	Framework :: Django :: 3.2
 	Framework :: Django :: 4.0
+	Framework :: Django :: 4.1
 	Intended Audience :: Developers
 	License :: OSI Approved :: BSD License
 	Operating System :: OS Independent
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,7 @@ zip_safe = False
 # jwcrypto has a direct dependency on six, but does not list it yet in a release
 # Previously, cryptography also depended on six, so this was unnoticed
 install_requires =
-	django >= 2.2, <= 4.1
+	django >= 2.2, != 4.0.0
 	requests >= 2.13.0
 	oauthlib >= 3.1.0
 	jwcrypto >= 0.8.0
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@ Contributors
 ------------
 
 Abhishek Patel
+Adam Johnson
 Alan Crosswell
 Alejandro Mantecon Guillen
 Aleksander Vaskevich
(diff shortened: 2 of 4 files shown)


adamchainz:
@n2ygk @Andrew-Chen-Wang , since you worked on #1175 which added this version bound.

setup.cfg
```diff
@@ -32,7 +32,7 @@ zip_safe = False
 # jwcrypto has a direct dependency on six, but does not list it yet in a release
 # Previously, cryptography also depended on six, so this was unnoticed
 install_requires =
-	django >= 2.2, <= 4.1
+	django >= 2.2, != 4.0.0
```
adamchainz, line 35: restoring the block on 4.0.0 which had a bug, as noted here:

https://github.com/jazzband/django-oauth-toolkit/pull/1039#pullrequestreview-834284574

Andrew-Chen-Wang (approved):
@n2ygk I don't think an upper bound is necessary either. also LGTM

setup.cfg
```diff
@@ -32,7 +33,7 @@ zip_safe = False
...
```
n2ygk, line 36:
```suggestion
	django >= 2.2, != 4.0.0, < 4.2
```

n2ygk (dismissed):
Please leave the upper bound test in. Making it open-ended is doing a disservice to library users. We just need to be more on top of updating the versions that we test against and resolving discrepancies.

n2ygk:
> @n2ygk I don't think an upper bound is necessary either. also LGTM

I kinda disagree given the tox tests are failing with djmain.

aaronmader:
Any idea when we can expect the next release to pypi, including this fix?

PR #1223, opened 2022-11-17
Clear expired ID tokens from database

### LINKED ISSUE #1222: Expired ID tokens are not removed
**Describe the bug**

When running `django-admin cleartokens`, access tokens and refresh tokens are cleared as documented, but expired ID tokens stick around in the DB until they are removed manually.

**To Reproduce**

1. Set  `OIDC_ENABLED = True`
2. Set `REFRESH_TOKEN_EXPIRE_SECONDS` to a low value
3. Set `ACCESS_TOKEN_EXPIRE_SECONDS` to a low value
4. Set `ID_TOKEN_EXPIRE_SECONDS` to a low value
6. Get a set of tokens
7. Wait until all of the tokens expired
8. Run `django-admin cleartokens`
9. The access and refresh token will be removed, the expired ID token is still in the DB

**Expected behavior**

I'd expect the ID token to be removed alongside the access token which holds the reference OR the ID token to be removed from the database when it expired.

**Version**

Tested with version 2.1.0.

**Additional context**

More generally, I wonder if revoking an access token should clear the ID token (if there is one), similar to how revoking a refresh token clears the associated access token?

I'd be willing to contribute a test and patch to the issue if a patch would be welcome for this issue (please let me know if that's the case).

Thanks for your consideration!

Pankrat:
The `cleartokens` management command removed expired refresh tokens and associated access tokens but kept expired ID tokens in the database.

Remove ID tokens when the associated access and refresh tokens are cleared.

Preserve expired tokens until the associated access token is deleted to keep relationships intact and not trigger delete cascades.

Fixes #1222

## Checklist

- [x] unit-test added
- [x] documentation updated


n2ygk (approved):
Great stuff. Thanks for this!

PR #1229, opened 2022-12-11
Fix CORS by passing 'Origin' header to OAuthLib

akanstantsinau:
Fixes #

## Description of the Change
Pass DJango `HTTP_ORIGIN` header to underlying OAuthLib

It is possible to control CORS by overriding is_origin_allowed method of RequestValidator class.
OAuthLib allows origin if:
- `RequestValidator.is_origin_allowed` returns `True` for particular request
- Request connection is secure
- Request has `Origin` header

## Checklist

- [x] unit-test added
- [x] documentation updated. No documentation. This is the fix for the feature documented in [OAuthLib](https://oauthlib.readthedocs.io/en/latest/oauth2/validator.html#oauthlib.oauth2.RequestValidator.is_origin_allowed)
- [x] `CHANGELOG.md` updated (only for user relevant changes) - change is not user relevant


n2ygk (commented):
Can you explain the use case for this given apps using Django CORS settings have been using this library successfully  without it?

dopry:
@akanstantsinau Are you unable to meet your CORS needs using the CORS middleware? The docs recommend using cors-headers for this purpose.  I'm wary of adding CORS handling. The scope of oauth support is pretty broad and there is already a lot of surface area in oauth_toolkit. I wouldn't want to add things we didn't have to and that weren't explicitly defined in OAuth/OIDC specifications. I'm willing to hear more about your case for adding this, but convenience isn't going to be a convincing argument.

akanstantsinau:
@dopry Thank your for looking into this PR
I can meet the same functionality using CORS middleware with a bit more coding.
The use case is pretty typical:
- We have a Django website  with some settings in CORS middleware
- We want to share user details with another service via OIDC (lets refer hose third party service as  https://example.com)
- OIDC client on example.com is web-based, so it requires CORS headers to be set on our `/o/token` endpoint which is provided by Django OAuth Toolkit
- We want to allow origin https://example.com *only* for url `/o/token`

The solution would be trivial if we do not have existing CORS middelware configs.
```
CORS_ALLOWED_ORIGINS=["https://example.com"]
CORS_URLS_REGEX = r"^/o/token$"
```

In our case we already have set `CORS_ALLOWED_ORIGINS` for some origins for *all* urls. So to allow https://example.com for `/o/token` we may implement own handler and attach  it to `corsheaders.signals.check_request_enabled` Django signal. 
No issues with it, if you recommend this approach.

The only motivation for this PR was to reuse already existing CORS feature in `oauthlib` project. It does exactly what we expect - set CORS headers only on `/o/token` url.
But I see you concerns regarding the scope of support.

oauth2_provider/oauth2_backends.py
```diff
@@ -75,6 +75,8 @@ def extract_headers(self, request):
             del headers["wsgi.errors"]
         if "HTTP_AUTHORIZATION" in headers:
             headers["Authorization"] = headers["HTTP_AUTHORIZATION"]
+        if "HTTP_ORIGIN" in headers:
+            headers["Origin"] = headers["HTTP_ORIGIN"]
```
dopry, line 79: Let's document why we're passing origin, we should link to the appropriate OAuthLib docs or code.

dopry (changes requested):
I think this is a good start to improving our CORS support. 

The default OauthLib request validator always returns false, https://github.com/oauthlib/oauthlib/blob/d4b6699f8ccb608152b764919e0bd3d38a7b171f/oauthlib/oauth2/rfc6749/request_validator.py#L656

I worry the change might cause cors failures for some consumers. It would be nice to also add origins to the application model and create a base validator that implements a lookup against the application model for allowed origins.

oauth2_provider/oauth2_backends.py
tonial: OAuthLib looks for "origin" in the headers : https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/base.py#L253-L268
We need to translate Django's value

dopry:
@akanstantsinau do you have the bandwidth to continue working on this issue?

akanstantsinau:
@dopry Apologize for long delay. Yes I do have bandwidth now to continue work on this. 
Your changes requested make sense for me. Will come back soon with updates.

rcoup:
FWIW, we implemeted this, and the approach is working well: It allows us to customise `is_origin_allowed()` to only be enabled for specific clients for specific origins.

akanstantsinau:
@dopry 
- implemented  `is_origin_allowed` in `oauth2_provider.oauth2_validators.OAuth2Validator`.
- added `allowed_origins` field for the allowed origins on the application model
 - only `token` endpoint uses `is_origin_allowed` according to `oauthlib` implementation.  Added tests for grant and refresh token
 - OIDC public endpoints (`oidc-connect-discovery-info`, `jwks-info`) use `Access-Control-Allow-Origin:*` 
 - OIDC `user-info` ignores `is_origin_allowed`

oauth2_provider/models.py
```diff
@@ -132,6 +132,10 @@ class AbstractApplication(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     updated = models.DateTimeField(auto_now=True)
     algorithm = models.CharField(max_length=5, choices=ALGORITHM_TYPES, default=NO_ALGORITHM, blank=True)
+    allowed_origins = models.TextField(
+        blank=True,
+        help_text=_("Allowed origins list to enable CORS, space separated"),
```
dopry, line 137: This help_text could be more informative.  Here is a suggestion. The details may not be accurate so it should be updated to reflect the actual requirements.

> Space-separated list of allowed origins for the token endpoint. The origin must be in the form of <scheme> "://" <host> [ ":" <port> ], such as https://login.mydomain.com or http://localhost:3000.  Query strings and hash information are not taken into account when validating these URLs. This does not include the 'Redirect URIs' or 'Post Logout Redirect URIs', if those domains will also use the token endpoint, they must be included in this list. 

We should decide whether the redirect uris are impicitly allowed origins or whether all allowed origins need to be explicitly specified. Also

tests/test_cors.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+        self.oauth2_settings.ALLOWED_REDIRECT_URI_SCHEMES = ["https"]
+
+    def tearDown(self):
+        self.application.delete()
+        self.test_user.delete()
+        self.dev_user.delete()
+
+    def test_cors_header(self):
```
dopry, line 59: The test name should be clear about the case you are testing something more like  `test_valid_origin_with_https`. The endpoint and Cors is already captures in the class name.

dopry (changes requested):
Awesome progress. This is great work and we're super close. I have a few small asks. Mostly simple renaming and docs, otherwise, from a code review perspective, we're pretty much there. After we can get those changes in I'll setup a test env and put it through it's paces!

oauth2_provider/models.py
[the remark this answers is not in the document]
dopry: I'm personally all for explicit and not making assumptions for the user at this juncture, even if it may lead to more support overhead.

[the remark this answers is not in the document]
akanstantsinau: Makes sense. Done.

oauth2_provider/oauth2_validators.py
```diff
@@ -958,3 +958,13 @@ def get_userinfo_claims(self, request):
...
+
+    def is_origin_allowed(self, client_id, origin, request, *args, **kwargs):
+        """Indicate if the given origin is allowed to access the token endpoint
+        via Cross-Origin Resource Sharing (CORS).  CORS is used by browser-based
+        clients, such as Single-Page Applications, to perform the Authorization
+        Code Grant.
+
+        Verifies if request's origin is within Application's allowed origins list.
+        """
+        return request.client.origin_allowed(origin)
```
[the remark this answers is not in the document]
akanstantsinau: Done

tests/test_cors.py
akanstantsinau: Done

dopry:
@akanstantsinau This looks great. Can you look at the coverage issue in Codecov? It looks like your tests should be covering those cases. Can you look into it and confirm those cases are covered? Also double check the doc changes to make sure there isn't a typo that is throwing sphinx for a loop.  I'll try to do a local test setup tomorrow.

akanstantsinau:
@dopry  I believe sphinx issue is something not related with changes in this PR:
- According to [build log ](https://readthedocs.org/api/v2/build/22123643.txt), dependency `sphinx-rtd-theme` is not installed
- `sphinx_rtd_theme` extension is required in [docs/conf.py](https://github.com/jazzband/django-oauth-toolkit/blob/9aa27c7528cdeda0b85bac5a8a00b39d696a43f9/docs/conf.py#L45)
- That causes build failure with ` No module named 'sphinx_rtd_theme'` error

The fix might be adding `sphinx-rtd-theme` dependency to `docs/requirements.txt`
New versions of sphinx and readthedocs-sphinx-ext were released in September 2023, so it may make things changed.

I'd prefer not to fix it in this PR, because its a different issue. Also my findings can be not correct.
 
Working on coverage issue.

rcoup:
ah, don't suppose django-oauth-toolkit is hit by the recent RTD changes? https://blog.readthedocs.com/defaulting-latest-build-tools/ or https://blog.readthedocs.com/newsletter-september-2023/

dopry:
@rcoup looks like you're right. I'll try to get a PR up to fix that today so we can get CI working again.

dopry:
@akanstantsinau I updated the branch with a rebase in the Github UI to see if the RTD fixes are good. Feel free to force push your local over it.

dopry:
@akanstantsinau it looks like everything is passing. I'll do some manual testing later today and hopefully we'll get this in.

akanstantsinau:
@dopry that are awesome news! Thank you!

oauth2_provider/models.py
```diff
@@ -132,6 +132,10 @@ class AbstractApplication(models.Model):
...
+    )
```
tonial, line 138: Don't we need to add a default value ?
I recently had to add one to another field in #1319

akanstantsinau: @tonial definitely! Missed that migration will fail if row exist in `application` table.
Thank you!

dopry: @tonial is this resolved to your satisfaction?

oauth2_provider/models.py
```diff
@@ -202,6 +215,13 @@ def clean(self):
                     grant_type=self.authorization_grant_type
                 )
             )
+        allowed_origins = self.allowed_origins.strip().split()
+        if allowed_origins:
+            # oauthlib allows only https scheme for CORS
+            validator = URIValidator({"https"})
```
dopry, line 221: The URIValidaor({"https"}) isn't a good fit here.  We need to allow testing on localhost for developers. IIRC origins should not have a path or fragment component but should only be <scheme>://<hostname>:<port>.  We probably need to implement an AllowedOriginValidator that extends the RedirectURIValidator with and allow_paths property and check. 

I'm a little confused as to why the `RedirectUriValidator` isn't instantiated as `RedirectURIValidator(WildcardSet(), allowed_schemes=allowed_schemes)` and `allowed_schemes` are being checked explicitly when scheme validation is part of `URLValidator`. We don't need to fix it as a part of this PR but it would be nice to put a 
```python 
# TODO:  Document why we're are not passing  allowed_schemes to RedirectURI Validation and explicitly checking. 
# If there isn't a reason we should rewrite this to use the allowed_schemes validation built into the parent class.
```

dopry (changes requested):
I wasn't able to do my local testing due to validation issues.

oauth2_provider/models.py
akanstantsinau: @dopry good point for localhost testing. Initially I was supposing that oauthlib allows only `https` origins.  (https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/base.py#L256) :
```
if not is_secure_transport(origin):
    log.debug('Origin "%s" is not HTTPS, CORS not allowed.', origin)
    return {}
```
But then discovered that setting `OAUTHLIB_INSECURE_TRANSPORT` environment variable will allow any schema. (https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/utils.py#L81)
This should be mentioned in documentation, otherwise user experience will be confusing, because non-https origins will be ignored by default.

dopry: @akanstantsinau thanks for finding that.  We should definitely mention it in the docs. Right now I'm wondering what is a good approach for allowed schemes. Application.get_allowed_schemes seems to be universal, but the setting ALLOWED_REDIRECT_URI_SCHEMES seems to be redirect specific.  It also enables ['http'] by default, which I feel is something a consumer should need to have to manually setup since it is insecure...   Part of me thinks we should add an ALLOWED_SCHEMES setting that is universal and use it for origins and set OAUTHLIB_INSECURE_TRANSPORT based on ALLOWED_SCHEMES...  In the future we can look at deprecating ALLOWED_REDIRECT_URI_SCHEMES in favor of an ALLOWED_SCHEMES....  We allow other schemes mainly for native app support, see: https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/

rcoup: Unless we're promoting really insecure defaults, IMO:

1. only allow `https`; but _also_ `http://localhost` / `http://127.0.0.1` / `http://::1` (plus dynamic tcp ports). MS Azure AD, for example, operates this way.
3. or someone very manually opts into `http` for non-localhost

I don't think people should have to allow all `http` to enable development/test setups using `localhost`. I can't see why CORS should be any different to redirect URIs in this regard?

dopry: As a sysadmin I want to ensure that no one can use an http:// origin, localhost or otherwise, on my production server.   We only allows testing against our production Auth server from SSL secured UAT environments.

rcoup: So do I in production, but I don't have external developer apps either :-) 

So feels like 3 options? 
* `https`: nothing but
* `http`: yolo security
* https with localhost-http: `auto`, `https+local`, `balance`, `default`, `mixed`... ? naming is hard.

Anyway, that change could land separately, since it presumably would apply to both origins+redirect URIs via ALLOWED_SCHEMES

dopry: Maybe we can open a feature request to discuss those settings. Personally I tend to prefer setting explicit values like `ALLOWED_SCHEMES=http,https,myapp` , `ALLOWED_DOMAINS=/regex/` over magic values like `auto`, `https+local`, `balance`, `default`, `mixed`. I find presets creating code that we need to maintain that isn't directly related to the business of authenticating, and they hide what's going on under the hood from consumers of the lib.

dopry: @akanstantsinau sorry about going on a tangent. We still need to resolve the Origin Validation issue. Are you up to adding a OriginURIValidator and oauth2_settings.ALLOWED_SCHEMES that will ensure the original is just <scheme>://<host>:<port>?

akanstantsinau:
@dopry Added `oauth_settings.ALLOWED_SCHEMES` with default set to `["https"]` and updated validation for `allowed_origins`

dopry (approved):

PR #1232, opened 2022-12-14
log output from management command to stderr instead of stdout

### LINKED ISSUE #1231: Management commands should write logs to stderr not stdout
**Describe the bug**
The `createapplication` management command writes log output to `stdout` instead of using `stderr` like regular python logging does. This makes it difficult to script around the management command when it clutters `stdout`.

**To Reproduce**
Run the `manage.py createapplication` command and see that output is to `stdout`

**Expected behavior**
Log output should go to `stderr`

**Version**
django-oauth-toolkit==1.3.2

tykling:
log output from management command to stderr instead of stdout to match normal python logging behaviour, update tests and CHANGELOG.md to match for #1231

Fixes #1231 

## Description of the Change
Change from self.stdout to self.stderr for output in management command

## Checklist

- [x] unit-test added (changed)
- [X] documentation updated


n2ygk:
I am concerned that this could be a breaking change for anyone currently capturing stdout for example with
```
Result = `manage.py createapplication ...`
```
 
Can't you work around by redirecting stdout to stderr?

tykling:
Hello :)

Yes, it would of course be a breaking change for anyone already scripting around the management command, if those wrapper scripts rely on `createapplication` outputting (or not outputting) like it does now. We could introduce an "output logs to stdout" switch to keep the current behavior, meaning the change to existing scripts would be minimal. But they would still have to be changed.

Of course I can work around it (and have done) in my wrapper script. But thinking about the big picture doing the inverse of what regular Python logging does just feels wrong.

Keeping the status quo means all future travelers will also have to work around this in scripts. So you have a some script wrapping a number of python scripts, and this one goes against what all the others do. It was surprising to me and I am sure it has been/will be to others.

I believe changing this is the right thing to do, but it should probably be done in a major release to please the semver gods.

n2ygk:
New functionality should not break things for existing users. Why not instead add a switch to change the logging to stderr? This is a management command and not a running service where syslogging is used....

tykling:
The default should be changed because the current behaviour is wrong, it outputs log messages to stdout which breaks scripts.

This has nothing to do with syslog though. When the Python logging framework logs to console it outputs to stderr simply because you want to leave stdout free for useful stuff, not clutter it with log messages like "The application was created successfully". This management command goes against that principle and should be fixed.

tykling:
Of course, this is all IMO, and if you disagree you can just close this PR and I will keep my workaround :+1: :smile:

n2ygk:
Yeah I think I will close it. After all stdout is for "normal" output and stderr is for errors. I'd rather not cause (unknown) other users of the management command to deal with a breaking change.

Thanks!

tykling:
I will leave this here for future explorers https://docs.python.org/3/howto/logging.html - it describes the default Python logging behaviour of sending logs to stderr. Bye for now :)

PR #1244, opened 2023-01-19
Implement OIDC RP-Initiated Logout

### LINKED ISSUE #1131: OpenID Connect RP-Initiated Logout
implement [OpenID Connect RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)

* [ ] add post_logout_redirect_uri to application model
* [ ] implement `end_session` endpoint conforming to [Section 2](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#rfc.section.2)
* [ ] add end_session_endpoint to discovery metadata per [Section 2.1](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#rfc.section.2.1).

Qup42:
Fixes #1131

## Description of the Change
Implement [OIDC RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html).

Regarding functionality the code is mostly finished. I am not yet satisfied with the structure of some parts of the code and would appreciate feedback or ideas to improve these points. Most functionality that `django-oauth-toolkit` provides is backed by `oauthlib`. This is not the case with the RP-Initiated Logout and I think that most of my uncertainty comes from that.
- The handling of exceptions. I am not quite happy with the new exception classes I introduced. But on the other hand the errors in other parts of `django-oauth-toolkit` are just passed on from `oauthlib`. The exceptions that are passend on don't require new exception classes.
- Location of `validate_logout_request`. For RP-Initiated Logout the actual core logic has to be implement in `django-oauth-toolkit`. Most other functionality can be forwarded to `oauthlib` via already existing classes (`OAuthLibCore`, etc.). But maybe there is a better place for this function.

## Checklist

- [x] unit-test added
- [x] documentation updated
- [x] Review Exceptions
- [x] Review `validate_logout_request`
- [x] Test two untested cases in `views/oidc.py`
- [x] Adapt code for Django 2.2


Qup42:
There are two things causing the tests to fail:
1. Django has dropped support for Python >3.10 on the main branch with a commit from 18/01/2023. Fixing that is probably out of scope of this PR.
2. Django 2.2 has a different access to the Response Headers than what I am used. I will fix that.

Andrew-Chen-Wang:
@Qup42 simply run `black` and your lint errors will probably be fine? Would be helpful to post what you get.

Qup42:
> @Qup42 simply run `black` and your lint errors will probably be fine? Would be helpful to post what you get.

Yes I have fixed them locally. I just wanted to state why the builds fail, as the lint errors are in locations that didn't change with this PR. So I was hesitant to add them to this PR and push the changes yet.

dopry:
@Qup42 Thank you so much for submitting this PR. I can't test it today, but I'll review first thing next week. I did do a quick scan over the code and the general gist looks good.

Qup42:
1. I will change it to `logout/`

2. That's an aspect I haven't thought about until now. It seems to me that the specs only considered the case of enabling it globally. I would prefer to stick to global enabling unless there is a common use case where this is required. If it is configurable per app some questions would arise as this situation is not standardized. E.g.

- The Discovery Endpoint is global and has no way to discern apps that may or may not have logout enabled. But optimally it should only show the logout URL to apps that also have it enabled. Alternatively should there be a way for an application to test whether remote logout is enabled for it?
- What is the behavior if an app tries to logout a user but does not have logout enabled? Automatic logout should probably not be possible. But should the user still be asked to logout and the return URL is also ignored or just show an error? 

3. Not if we want to be compliant with the specs. Specs:
> At the Logout Endpoint, the OP SHOULD ask the End-User whether to log out of the OP as well. Furthermore, the OP MUST ask the End-User this question if an id_token_hint was not provided or if the supplied ID Token does not belong to the current OP session with the RP and/or currently logged in End-User.

and

> 6.  Security Considerations
> Logout requests without a valid id_token_hint value are a potential means of denial of service; therefore, OPs should obtain explicit confirmation from the End-User before acting upon them.

dopry:
@Qup42 I've shared a gist of an end session endpoint I implemented for a project. I didn't implement the user confirmation part of the workflow, but I do have a sample to expired token validation you can grab and maybe a few other things... https://gist.github.com/dopry/36e8cb676b08697ad26612521993b57d

n2ygk:
@Qup42  FYI I've just rebased this

Qup42:
> @Qup42 I've shared a gist of an end session endpoint I implemented for a project. I didn't implement the user confirmation part of the workflow, but I do have a sample of expired token validation you can grab and maybe a few other things... https://gist.github.com/dopry/36e8cb676b08697ad26612521993b57d

Looking at your gist I see that you delete the access tokens. Previously libraries built on top of django-oauth-toolkit would do the actual logout. Now that we actually also logout a user this has to be considered.

I'll integrate the token handling and token deletion from your gist in about three weeks once I'm done with my exams.

Qup42:
>     3. I wasn't asking about whether to prompt or not. I was wondering what happens when an id_token and client_hint are not provided. I see the pass the request through so it looks like prompting is required in that case. What happens if a user is not authenticated and provides a hint and token? Can a user be maliciously logged out by an actor that has exfiltrated an expired token. does django's logout function verify there the session is valid before logging out a user or should this implementation do that?

- RP-Initiated Logout should be idempotent. When hint and token are provided logout can happen without interaction. So if a user is already logged out nothing should happen (it is probably a good idea to try to delete the access tokens nonetheless?).
- Yes. The specs allow this case and even encourage accepting time-expired tokens. But there is no obligation to accept expired tokens.
- Django's logout does nothing when no user is logged in. The `user_logged_out` signal seems to be invoked regardless of whether someone was actually logged out.

Qup42:
> @Qup42 I've shared a gist of an end session endpoint I implemented for a project. I didn't implement the user confirmation part of the workflow, but I do have a sample of expired token validation you can grab and maybe a few other things... https://gist.github.com/dopry/36e8cb676b08697ad26612521993b57d

Regarding expired tokens: have you tested that the code in your snipped does actually accept expired tokens? (Shouldn't `JWTExpired` be raised via `JWT::init -> JWT::deserialize -> JWT::validate -> JWT::_check_provided_claims -> JWT::_check_default_claims -> JWT::_check_exp`?)

Anyways ... if the code already accepts expired tokens - great. Otherwise `check_claims=False` in the constructor of `JWT` disables the check for the expiry date and the not before timestamp.

dopry:
Thanks for the QC on the JWT check,,, I'll look into it. I recall that it supported expired tokens, but it's also been over a year since I wrote that and it was on a time crunch so I very well could have made some flawed assumptions when writing the code. :), 

I think deleting tokens would probably be good. I used Oauth2AccessToken.objects.filter(user=request.user).delete(). I'd want to check that request wouldn't be too greedy and delete things like client credentials created by a user that should be persistent. I don't have an m2m or client credentials in my environment.

dopry (commented):
Looking good, Overloading _load_id_token with both validation and token loading caught my attention at first. Applying the single resposibility prinicpal here would lead us to decompose validating the provided token and loading a token into to two separate functions.  It is paralell to the ` oauth2_validators._load_id_token`, so there is a strong case to maintain consistency here so I would approve this.

@Qup42 any plans on implementing token removal along side logout?

Qup42:
> Looking good, Overloading _load_id_token with both validation and token loading caught my attention at first. Applying the single resposibility prinicpal here would lead us to decompose validating the provided token and loading a token into to two separate functions. It is paralell to the ` oauth2_validators._load_id_token`, so there is a an argument for maintaining consistency, so I would approve this.
> 
> @Qup42 any plans on implementing token removal along side logout?

Thanks for having a look. I also wasn't quite satisfied with that function. I have now split it into two: loading the IDToken and validating the claims. The difference to the validator is that the validation doesn't happen on the `IDToken` object. For validation `jwt.JWT` or the claims are required. Both can't be extracted from the `IDToken`.

No worries - I haven't forgotten the token removal ;) I will probably implement it in the next couple days.

Qup42:
It can now be configured whether tokens should be deleted at all and for which types of clients (public/confidential) and authorization grant types.

Deleting tokens just because a User logs seems to be a bit against the idea of the tokens to me, especially for IDTokens. Currently I also delete the IDTokens of the user that is being logged out. I'm unsure about that aspect. Do you have any input on whether the IDTokens should also be deleted or kept? @dopry

dopry:
@Qup42 thanks for your work on this. I'm really looking forward to getting it in. I think a lot of people can benefit from this work. 

> Deleting tokens just because a User logs seems to be a bit against the idea of the tokens to me, especially for IDTokens. Currently I also delete the IDTokens of the user that is being logged out. I'm unsure about that aspect. Do you have any input on whether the IDTokens should also be deleted or kept? @dopry

I'm not working from the concept of a token and how they're intended to be used, but from the Users objective of being logged out. If a user has requested to be logged out of a specific RP or at the IDP, I suspect their expectation is that their session is ended and that their 'live' PII is fully invalidated and wiped.  When I log out of a server side session based app, my session is fully deleted along with any identification info contained there in. It may still live at rest in the DB, but the sessions and live copy of the PII is gone.  From a more practical perspective, I've seen a number of RP implementations use an IDToken as an auth token. Not saying it's right or should be actively supported, but I've seen it and we should consider it. Not all RPs with validate the token on every request anyway if they're still holding it, but I feel like it's best to tell RPs that the ID token is no longer valid of they try to check one once a use has actively ended their session. This is just my opinion. 

Since we have opposing views, I think we should try to lean back on spec to determine how to handle this scenario. Based on your reading of OAUTH related specs and tokens, why do you feel that ID Tokens shouldn't be deleted?

dopry:
@Qup42 I wanted to check in. I've been busy the last few weeks. I think it's okay to delete the tokens. In what cases would you not want to delete them?

Qup42:
@dopry Hmm I guess you're right. Doing what most users expect is probably the better approach here. The deletion of the tokens can also be turned of should that be required.

I also merged the master branch once more. What would be the next step going forward?

dopry:
@Qup42 I'm starting into testing today. I rebased your PR so the feature history will be linear.

docs/oidc.rst
```diff
@@ -147,6 +149,23 @@ scopes in your ``settings.py``::
...
+~~~~~~~~~~~~~~~~~~~
+This feature has to be enabled separately as it is an extension to the core standard.
+
+.. code-block:: python
+
+   OAUTH2_PROVIDER = {
+       # OIDC has to be enabled to use RP-Initiated Logout
+       "OIDC_ENABLED": True,
+       # Enable and configure RP-Initiated Logout
+       "OIDC_RP_INITIATED_LOGOUT": True,
```
dopry, line 163: Should be "OIDC_RP_INITIATED_LOGOUT_ENABLED"

dopry (changes requested):
Found a small thing. I'll get it if all goes well in further testing.

docs/oidc.rst
Qup42: Good catch, Thanks!

oauth2_provider/views/oidc.py
```diff
@@ -120,3 +147,246 @@ def _create_userinfo_response(self, request):
...
+        application = id_token.application
+
+    # Validate `post_logout_redirect_uri`
+    if post_logout_redirect_uri:
+        if not application:
+            raise InvalidOIDCClientError()
+        scheme = urlparse(post_logout_redirect_uri)[0]
+        if not scheme:
+            raise InvalidOIDCRedirectURIError("A Scheme is required for the redirect URI.")
+        if scheme == "http" and application.client_type != "confidential":
```
dopry, line 255: We might want to consider skipping this when running in DEBUG on localhost to facilitate local integration testing... @Qup42  what do you think?

Qup42: I understand the use case. But I couldn't find any other parts of the library where the logic is changed when running in `DEBUG`. The only thing I found is that more information is shown. I don't know if the convenience for local testing outweighs the reduced security if some deploys in `DEBUG` mode. Personally I would lean against doing that but I am also not completely opposed.

Qup42:
@dopry I'll wait for you to finish reviewing and integrate the changes after that in bulk.

dopry:
@Qup42 I've been testing, so far everything looks good. I'm still working on getting an RP setup to do a full integration test with a  3rd party oidc client library.

n2ygk:
@dopry This might help with testing: https://github.com/n2ygk/dot-tutorial/blob/main/certification/notes.md

The heroku app is not currently set up but I can do that.

dopry:
I was setting up an example IDP (django + django_oauth_toolkit) and RP (https://github.com/dopry/svelte-oidc) locally per https://github.com/jazzband/django-oauth-toolkit/issues/1266 to facilitate testing. My svelte-oidc project needed some love and I thought I'd kill two bird with one stone.

oauth2_provider/views/oidc.py
```diff
@@ -120,3 +147,246 @@ def _create_userinfo_response(self, request):
...
+            )
+            for token in access_tokens_to_delete:
+                # Delete the token and its corresponding refresh and IDTokens.
+                if token.id_token:
+                    token.id_token.revoke()
+                token.revoke()
+            for refresh_token in refresh_tokens_to_delete:
+                refresh_token.revoke()
+        # Logout in Django
+        logout(self.request)
```
tonial, line 374: There's an issue here if the RP sent the request from it's server.

Django logout won't be able to remove the session created in the login view which will allow the user to login again from RP without entering their credentials.
 
We can clear the sessions with the following code (not really satisfying, but it works)
```
[
    s.delete()
    for s in Session.objects.filter(expire_date__gte=timezone.now())
    if s.get_decoded().get("_auth_user_id") == str(id_token.user_id)
]
```

Another way might be to link the Session to one of the tokens so that we can retrieve them.

tonial: This also raises the question of `REFRESH_TOKEN_EXPIRE_SECONDS` vs `SESSION_COOKIE_AGE` : 
If Django session bigger than `REFRESH_TOKEN_EXPIRE_SECONDS` then we can be in the case where the tokens were deleted with `cleartokens` management command, but the session still exists.

tonial (commented):
Hi.

First of all, I'm really glad to that you worked on this feature for DOT.
It was clearly not a small task, and I'm sure it will help a lot of people (including me !) 

I've done some testing and I think there's a issue regarding logout without prompt sent from the RP server, so I added some comments in your pull request.

oauth2_provider/views/oidc.py
```diff
@@ -120,3 +147,246 @@ def _create_userinfo_response(self, request):
...
+    id_token = None
+    must_prompt_logout = True
+    if id_token_hint:
+        # Only basic validation has been done on the IDToken at this point.
+        id_token, claims = _load_id_token(id_token_hint)
+
+        if not id_token or not _validate_claims(request, claims):
+            raise InvalidIDTokenError()
+
+        if id_token.user == request.user:
```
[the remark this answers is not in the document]
Qup42: 1. This case is not planed for by the specifications. We also added this functionality in our own deployment. If this is added, it should probably be a separate PR.
> An RP requests that the OP log out the End-User by redirecting the End-User's User Agent to the OP's Logout Endpoint.

2. This is the intended behavior from my reading of the specifications. 
> Furthermore, the OP MUST ask the End-User this question [asking to confirm the logout] if an id_token_hint was not provided or if the supplied ID Token does not belong to the current OP session with the RP and/or currently logged in End-User.

My interpretation is that the user should be asked for confirmation if any of the conditions is not fulfilled. If a user does not have a session then the session does not match with the ID Token. But I am also no expert in reading and understanding specifications. It might be worthwhile to have a look at the behavior of already existing implementations (e.g. Auth0, Zitadel)

oauth2_provider/views/oidc.py
```diff
@@ -120,3 +147,246 @@ def _create_userinfo_response(self, request):
...
+
+    The `id_token_hint` will be validated if given. If both `client_id` and `id_token_hint` are given they
+    will be validated against each other.
+    """
+
+    id_token = None
+    must_prompt_logout = True
+    if id_token_hint:
+        # Only basic validation has been done on the IDToken at this point.
+        id_token, claims = _load_id_token(id_token_hint)
```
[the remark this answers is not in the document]
Qup42: Thanks for the hint. I'll add that to the list of things to change.

oauth2_provider/views/oidc.py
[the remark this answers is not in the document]
tonial: English isn't my native language, that might be why I didn't understand the specs the same way you did.

1) I'm unclear if "End-User's User Agent" also matches the RP server.
On [Wikipedia](https://en.wikipedia.org/wiki/User_agent) we can see :
> In computing, a user agent is any software, acting on behalf of a user, which "retrieves, renders and facilitates end-user interaction with Web content".
Isn't the RP sever performing an action on behalf of the user ?
If not, I agree on the separated PR.

2) You are right. If the django session is expired, then we need to prompt the user before removing the tokens.
[Keycloak does work this way](https://www.keycloak.org/docs/latest/securing_apps/index.html#logout) (supplying the id_token_hint is enougth to skip confirmation)  but maybe the case doesn't exists in Keycloak because both (session and id token) have the same lifespan.

oauth2_provider/views/oidc.py
Qup42: Although I don't want to cater to the use case of a back channel logout, deleting all sessions might be usefull. It is a shame that Django only allows such a hacky way to achieve this. I might add this together with a setting to enable/disable it.

dopry: @tonial per https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

> An RP requests that the OP log out the End-User by redirecting the **End-User's User Agent** to the OP's Logout Endpoint.

RP initiated logout does not allow an RP's server will make a direct call to the end_session_endpoint on behalf the user. The scenario you are describing is outside of the bounds of protocol and not applicable.

oauth2_provider/views/oidc.py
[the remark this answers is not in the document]
dopry: @tonial  the spec specifically refers to https://www.rfc-editor.org/rfc/rfc7230.html for the definition of user agent.  In this case "End User Agent", is referring to something use directly by the end user, where a server relaying a request for a user would be a intermediate user agent or proxy.   The interpretation in this case should be an application directly under the user control such as a web browser or application such as a mobile app or desktop app that can act as an agent.

oauth2_provider/views/oidc.py
dopry: I do extensive local development with docker hosting services. It is a very important use case for me. However I've been looking closely at https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

```
post_logout_redirect_uri - OPTIONAL. URI to which the RP is requesting that the End-User's User Agent be redirected after a logout has been performed. This URI SHOULD use the https scheme and MAY contain port, path, and query parameter components; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of [OAuth 2.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RFC6749) [RFC6749], and provided the OP allows the use of http RP URIs. The URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application. The value MUST have been previously registered with the OP, either using the post_logout_redirect_uris Registration parameter or via another mechanism. An id_token_hint is also RECOMMENDED when this parameter is included.
```

technically 'SHOULD' is recommended with caveats... MUST is not used in this situation. So as long as one understands and has carefully weighed the impact it's okay to use HTTP instead of https. 


This is very similar to https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

```
REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in Section 6.2.1 of [[RFC3986]](https://openid.net/specs/openid-connect-core-1_0.html#RFC3986) (Simple String Comparison). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of OAuth 2.0, and provided the OP allows the use of http Redirection URIs in this case. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
```

The scheme is not validated on the authorization request.  

What is validated on the request is the ALLOWED_REDIRECT_URI_SCHEMES and that the redirect_uri is present in the applications redirect_uris.  I feel like we could remove any explicit check for 'scheme' here and instead validate post_logout_redirect_uri on the application model and show a big warning note near the field saying that http schemes are strongly discouraged by the spec and pointing someone to the examples. 

For now I think we can just remove the https check.

oauth2_provider/views/oidc.py
dopry: @tonial also thank your for taking the time to review. The catch on the token parsing was particularly nice.

oauth2_provider/views/oidc.py
Qup42: Your argument has convinced me enough that I would add this and make it active when in `DEBUG` mode.

dopry: I have use cases where I use http://localhost:5147/ etc  as valid redirect URIs for doing local frontend development in react/svelte/vue etc against a production IDP and API. I would recommend we forego the https validation for post_logout_redirect_uri in the RPInitiatedLogoutView / validate_logout_request

dopry: I've also confirmed that Auth0, Okta, and IdentityServer  allow http:// post_logout_redirect_uris, and based on code review solely, so does Keycloak.

dopry:
@Qup42 I've completed my review... The last changes are
- [ ] handling token parsing exceptions and noted by @tonial 
- [ ] documentation typo I noted 
- [ ] remove https validation from post_logout_redirect_uri and explain the trade off in the help text for AbtractApplication.post_logout_redirect_uri

oauth2_provider/views/oidc.py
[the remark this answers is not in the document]
tonial: Thanks for the explanation !

oauth2_provider/views/oidc.py
Qup42: `http` is now allowed per default. Strict Checking can be enabled with `OIDC_RP_INITIATED_LOGOUT_STRICT_REDIRECT_URIS`.

dopry (changes requested):
l love the solution by using a strict setting. There is one typo in the docs, If you can get around to fixing it that would be awesome. I gotta run to a meeting, if you haven't gotten to it before I'm back I'll update it when I do a final testing pass.

docs/settings.rst
```diff
@@ -313,6 +313,41 @@ this you must also provide the service at that endpoint.
...
+Default: ``True``
+
+Whether to always prompt the :term:`Resource Owner` (End User) to confirm a logout requested by a
+:term:`Client` (Relying Party). If it is disabled the :term:`Resource Owner` (End User) will only be prompted if required by the standard.
+
+OIDC_RP_INITIATED_LOGOUT_STRICT_REDIRECT_URIS
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Default: ``False``
+
+Enable to only allow the `http` scheme in post logout redirect URIs when a :term:`Client` is `confidential`.
```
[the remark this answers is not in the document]
Qup42: `http` is correct there. I have reworded the sentence in [8413052](https://github.com/jazzband/django-oauth-toolkit/pull/1244/commits/84130520c7e1627b4a0eea17234ef75404d74f2d) to make it clearer.

dopry (approved):
LGTM.

dopry:
@Qup42 Thank you so much for all your work on this. It is very appreciated.

PR #1245, opened 2023-01-20
Remove incompatible python+django version combinations from CI

Qup42:
## Description of the Change
The pipeline tests are failing for `djmain` with python < 3.10 because the main branch of django requires python >= 3.10. This PR removes the incompatible python versions from the `djmain` step.

### Why did the tests break?
Django has released 4.2 alpha 1[^1] with a feature freeze on 17/01/2023. This means that the main branch is now working towards 5.0 and has thus dropped python < 3.10[^2] on 18/01/2023.

[^1]: https://www.djangoproject.com/weblog/2023/jan/17/django-42-alpha-1-released/
[^2]: https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Joseph Abrahams
 Josh Thomas
 Jozef Knaperek
 Julien Palard
+Julian Mundhahs
 Jun Zhou
 Kaleb Porter
 Kristian Rune Larsen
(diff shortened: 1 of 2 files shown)


n2ygk (approved):
Thanks!

PR #1264, opened 2023-04-14
Test on Django 4.2

adamchainz:
## Description of the Change

Test with the new Django! 🎉

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -1,3 +1,5 @@
+import django
+
 ADMINS = ()
 
 MANAGERS = ADMINS
@@ -23,7 +25,8 @@ LANGUAGE_CODE = "en-us"
 SITE_ID = 1
 
 USE_I18N = True
-USE_L10N = True
+if django.VERSION < (4, 0):
+    USE_L10N = True
 USE_TZ = True
 
 MEDIA_ROOT = ""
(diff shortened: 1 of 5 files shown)


tests/settings.py
```diff
@@ -23,7 +25,8 @@
 SITE_ID = 1
 
 USE_I18N = True
-USE_L10N = True
+if django.VERSION < (4, 0):
+    USE_L10N = True
```
adamchainz, line 30: This change avoids a warning on Django 4.0+:

```
/.../site-packages/django/conf/__init__.py:267: RemovedInDjango50Warning: The USE_L10N setting is deprecated. Starting with Django 5.0, localized formatting of data will always be enabled. For example Django will display numbers and dates using the format of the current locale.
    warnings.warn(USE_L10N_DEPRECATED_MSG, RemovedInDjango50Warning)
```

tox.ini
```diff
@@ -9,6 +9,7 @@ envlist =
     py{37,38,39,310}-dj32,
     py{38,39,310}-dj40,
     py{38,39,310,311}-dj41,
+    py{38,39,310,311}-dj42,
```
adamchainz, line 12: https://docs.djangoproject.com/en/4.2/releases/4.2/#python-compatibility

n2ygk (approved):
Thanks!

PR #1270, opened 2023-05-15
Fix RP-initiated Logout with expired Django session

### LINKED ISSUE #1269: RP-initiated Logout with expired django session not revoking the tokens
**Describe the bug**
If there's not active Django session when calling the RP-initiated Logout view with a valid `id_token_hint`, then the tokens won't be revoked. 

Since there's not active session, the code will always prompt a confirmation form : https://github.com/jazzband/django-oauth-toolkit/blob/11294ab5678691fb6bc21ecf917dca3099311a9e/oauth2_provider/views/oidc.py#L234-L237

But in `do_logout()` the code relies on `self.request.user` which is an `AnonymousUser`, therefore no tokens will be revoked : https://github.com/jazzband/django-oauth-toolkit/blob/11294ab5678691fb6bc21ecf917dca3099311a9e/oauth2_provider/views/oidc.py#L364-L368

The app may even crash (it's my case because the User PK is a UuidField, and `"AnonymousUser"` isn't a valid uuid) 

**To Reproduce**

Have a Django application with django-oauth-toolkit installed
Set `settings.SESSION_COOKIE_AGE=10` (seconds)
Perform an oauth login
Call RP-initiated Logout view passing the retrieved id_token in `id_token_hint` params.

**Expected behavior**
The tokens for the user found with `id_token_hint` should be revoked.

**Version**
master branch on commit https://github.com/jazzband/django-oauth-toolkit/commit/11294ab5678691fb6bc21ecf917dca3099311a9e

**Additional context**

tonial:
Fixes #1269

## Description of the Change

Fix RP initiated logout when there's no django session available.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/tests/test_oidc_views.py
+++ b/tests/test_oidc_views.py
@@ -536,7 +536,7 @@ def test_token_deletion_on_logout_expired_session(oidc_tokens, client, rp_settin
     )
     assert rsp.status_code == 200
     assert not is_logged_in(client)
-    # Check that all tokens have either been deleted or expired.
+    # Check that all tokens are active.
     assert AccessToken.objects.count() == 1
     assert not any([token.is_expired() for token in AccessToken.objects.all()])
     assert IDToken.objects.count() == 1
--- a/tests/test_oidc_views.py
+++ b/tests/test_oidc_views.py
@@ -1,3 +1,4 @@
+from pytest_django.asserts import assertRedirects
 import pytest
 from django.contrib.auth import get_user
 from django.contrib.auth.models import AnonymousUser
@@ -552,7 +553,7 @@ def test_token_deletion_on_logout_expired_session(oidc_tokens, client, rp_settin
             "allow": True,
         },
     )
-    assert rsp.status_code == 302
+    assertRedirects(rsp, "http://testserver/", fetch_redirect_response=False)
     assert not is_logged_in(client)
     # Check that all tokens have either been deleted or expired.
     assert all(token.is_expired() for token in AccessToken.objects.all())
--- a/tests/test_oidc_views.py
+++ b/tests/test_oidc_views.py
@@ -1,10 +1,10 @@
-from pytest_django.asserts import assertRedirects
 import pytest
 from django.contrib.auth import get_user
 from django.contrib.auth.models import AnonymousUser
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
 from django.utils import timezone
+from pytest_django.asserts import assertRedirects
 
 from oauth2_provider.exceptions import ClientIdMissmatch, InvalidOIDCClientError, InvalidOIDCRedirectURIError
 from oauth2_provider.models import get_access_token_model, get_id_token_model, get_refresh_token_model
--- a/tests/test_oidc_views.py
+++ b/tests/test_oidc_views.py
@@ -555,9 +555,9 @@ def test_token_deletion_on_logout_expired_session(oidc_tokens, client, rp_settin
     assert rsp.status_code == 302
     assert not is_logged_in(client)
     # Check that all tokens have either been deleted or expired.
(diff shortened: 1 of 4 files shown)


tests/test_oidc_views.py
```diff
@@ -519,6 +519,47 @@ def test_token_deletion_on_logout(oidc_tokens, loggend_in_client, rp_settings):
...
+    rsp = client.get(
+        reverse("oauth2_provider:rp-initiated-logout"),
+        data={
+            "id_token_hint": oidc_tokens.id_token,
+            "client_id": oidc_tokens.application.client_id,
+        },
+    )
+    assert rsp.status_code == 200
+    assert not is_logged_in(client)
+    # Check that all tokens have either been deleted or expired.
```
francoisfreitag, line 539:
```suggestion
    # Check that all tokens are active.
```

tests/test_oidc_views.py
```diff
@@ -519,6 +519,47 @@ def test_token_deletion_on_logout(oidc_tokens, loggend_in_client, rp_settings):
...
+    assert AccessToken.objects.count() == 1
+    assert not any([token.is_expired() for token in AccessToken.objects.all()])
+    assert IDToken.objects.count() == 1
+    assert not any([token.is_expired() for token in IDToken.objects.all()])
+    assert RefreshToken.objects.count() == 1
+    assert not any([token.revoked is not None for token in RefreshToken.objects.all()])
```
francoisfreitag, line 545:
```suggestion
    access_token = AccessToken.objects.get()
    assert not access_token.is_expired()
    id_token = IDToken.objects.get()
    assert not id_token.is_expired()
    refresh_token = RefreshToken.objects.get()
    assert refresh_token.revoked is None
```

tests/test_oidc_views.py
```diff
@@ -519,6 +519,47 @@ def test_token_deletion_on_logout(oidc_tokens, loggend_in_client, rp_settings):
...
+            "client_id": oidc_tokens.application.client_id,
+            "allow": True,
+        },
+    )
+    assert rsp.status_code == 302
+    assert not is_logged_in(client)
+    # Check that all tokens have either been deleted or expired.
+    assert all([token.is_expired() for token in AccessToken.objects.all()])
+    assert all([token.is_expired() for token in IDToken.objects.all()])
+    assert all([token.revoked <= timezone.now() for token in RefreshToken.objects.all()])
```
francoisfreitag, line 560: Can remove unnecessary list comprehension.
```suggestion
    assert all(token.is_expired() for token in AccessToken.objects.all())
    assert all(token.is_expired() for token in IDToken.objects.all())
    assert all(token.revoked <= timezone.now() for token in RefreshToken.objects.all())
```

tests/test_oidc_views.py
```diff
@@ -519,6 +519,47 @@ def test_token_deletion_on_logout(oidc_tokens, loggend_in_client, rp_settings):
...
+
+    rsp = client.post(
+        reverse("oauth2_provider:rp-initiated-logout"),
+        data={
+            "id_token_hint": oidc_tokens.id_token,
+            "client_id": oidc_tokens.application.client_id,
+            "allow": True,
+        },
+    )
+    assert rsp.status_code == 302
```
francoisfreitag, line 555: Perhaps verify where the client is redirected?

francoisfreitag (approved):

dopry:
@tonial good catch.  I'm not super jazzed about the complex return type from _validate_logout_request.  I feel like that could be broken up into a few discrete methods on the RPInitiatedLogoutView  and more of the validation could be lifted up into form_valid. I'm gonna go ahead and merge so we fix the breakage.

dopry (approved):
LGTM.

PR #1281, opened 2023-06-02
🐞 fix(oidc-logout): 修复一个引发500 error的问题

### LINKED ISSUE #1280: Route "/o/logout" returns 500 Internal Server Error
**Describe the bug**
When the logout request comes from an anonymous user and the ID Token is empty, clicking the Logout button will return 500 Internal Server Error

**To Reproduce**
In the case of not logging in, open "/o/logout" directly, make sure that the parameter id_token_hint is not carried, and then click the "Logout" button to reproduce this problem

**Expected behavior**
I think that when no id_token is provided and it is impossible to confirm which user the request came from, a 400 error should be returned to remind the user that we cannot perform an exit operation for him/her

**Version**
V2.3.0

Polaris-d:
Fixes #1280

## Description of the Change

Returns a 400 error when no id_token_hint is provided and it is impossible to confirm who the request came from

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -94,3 +94,4 @@ Víðir Valberg Guðmundsson
 Will Beaufoy
 pySilver
 Łukasz Skarżyński
+Dawei Liu
--- a/oauth2_provider/exceptions.py
+++ b/oauth2_provider/exceptions.py
@@ -60,6 +60,10 @@ class InvalidIDTokenError(InvalidRequestFatalError):
     description = "The ID Token is expired, revoked, malformed, or otherwise invalid."
 
 
+class EmptyIDTokenError(InvalidRequestFatalError):
+    description = "The current request is from an anonymous user and the ID Token is empty."
+
+
 class LogoutDenied(OIDCError):
     error = "logout_denied"
     description = "Logout has been refused by the user."
--- a/oauth2_provider/views/oidc.py
+++ b/oauth2_provider/views/oidc.py
@@ -2,6 +2,7 @@ import json
 from urllib.parse import urlparse
 
 from django.contrib.auth import logout
+from django.contrib.auth.models import AnonymousUser
 from django.http import HttpResponse, JsonResponse
 from django.urls import reverse
 from django.utils.decorators import method_decorator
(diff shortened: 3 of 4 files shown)


tonial:
Hi, 

The OIDC RP-initiated logout specifications (https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) states that `id_token_hint` is only recommended.

This means the view should allow a RP to send a user without `id_token_hint`, and the code should not raise an error if the user session already expired.

I think a better way to fix the issue would be to improve again `do_logout()` by skipping the `if oauth2_settings.OIDC_RP_INITIATED_LOGOUT_DELETE_TOKENS:` part if no user was found (`token_user == None` and `self.request.user` is anonymous)

Polaris-d:
> Hi,
> 
> The OIDC RP-initiated logout specifications (https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) states that `id_token_hint` is only recommended.
> 
> This means the view should allow a RP to send a user without `id_token_hint`, and the code should not raise an error if the user session already expired.
> 
> I think a better way to fix the issue would be to improve again `do_logout()` by skipping the `if oauth2_settings.OIDC_RP_INITIATED_LOGOUT_DELETE_TOKENS:` part if no user was found (`token_user == None` and `self.request.user` is anonymous)

I read it again carefully and found some very interesting places

1. It has never been mentioned that the OP is not logged in, does this mean that the author may not be aware of this situation, or does it mean that the OP must be logged in when using the logout endpoint
2. `At the Logout Endpoint, the OP SHOULD ask the End-User whether to log out of the OP as well.` It is specially emphasized here that when logging out, the user should be prompted whether to log out the OP at the same time, which should mean that RP and OP should be processed separately according to the user's choice, instead of logging out at the same time

There is no doubt that it is precisely because of the ambiguity of the documentation that we cannot clarify the processing flow of this situation. I also tried to find some sites that already support logout and refer to their practices, but so far I have nothing reward

tonial:
Thanks @Polaris-d 

SInce I've got some time and also had to mitigate the issue on my project, I've opened # 1284 to fix the issue

PR #1284, opened 2023-06-12
Fix RP Initiated Logout when no user is found

### LINKED ISSUE #1280: Route "/o/logout" returns 500 Internal Server Error

tonial:
Fixes #1280

## Description of the Change

Don't crash when we only have a AnonymousUser to logout (no id_token_hint or existing session), 

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * #1273 Add caching of loading of OIDC private key.
 
+- ### Fixed
+* #1284 Allow to logout whith no id_token_hint even if the browser session already expired
+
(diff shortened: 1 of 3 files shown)


n2ygk (approved):
LGTM. Thanks!

PR #1285, opened 2023-06-12
Add post_logout_redirect_uris field to application views

tonial:
## Description of the Change

These are missing since RP-initiated logout was added. 

I added more tests on the applications views
- check that no posted data is dropped by the form on update and registration views,
- tests the displayed forms on the detail view.

## Checklist

- [x] unit-test added
- [ ] documentation updated


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * #1273 Add caching of loading of OIDC private key.
+* #1285 Add post_logout_redirectè_uris field in application views.
```
n2ygk, line 21:
```suggestion
* #1285 Add post_logout_redirect_uris field in application views.
```

oauth2_provider/views/application.py
```diff
@@ -37,6 +37,7 @@ def get_form_class(self):
                 "client_type",
                 "authorization_grant_type",
                 "redirect_uris",
+                "post_logout_redirect_uris",
```
n2ygk, line 40: Is meant to be this a plural list of uris or singular? If the latter, please change the name to singular `post_logout_redirect_uri`

n2ygk (changes requested):
This looks like new functionality that needs to be documented. Please update the docs. See [docs related to redirect_uris](https://django-oauth-toolkit.readthedocs.io/en/latest/search.html?q=redirect_uris&check_keywords=yes&area=default)

Looks like you may want to also update the create_application management command.

oauth2_provider/views/application.py
n2ygk: Also the [test is failing](https://github.com/jazzband/django-oauth-toolkit/actions/runs/5246478141/jobs/9475365332?pr=1285#step:7:129)

tonial:
The management command already allows to specify post_logout_redirect_uris : [see here](https://django-oauth-toolkit.readthedocs.io/en/latest/management_commands.html?highlight=redirect_uris#id2)

I've added the new fields in the documentation

PR #1286, opened 2023-06-12
Allow to disable applications management views

tonial:
## Description of the Change

Add a setting to disable management views.

Without this, anybody can use the urls to add an application, which could lead to a breach in security (typically if the application can be used to retrieve data from an API).

I think it would be better to make the default False, but It would be a backward-incompatible change for every project using these URLs. 

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * #1273 Add caching of loading of OIDC private key.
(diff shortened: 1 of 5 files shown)


tonial:
On second though, I'm dropping this.

I think a better way to do it would be to have the different urls groups (base, management and oidc) imported by the projects.
This way, we only import management urls if we want.

And using a setting to add or not some urls in the urlpatterns is just awful to test...

dopry:
@tonial I agree grouped urls that consumers can conditionally include would be ideal.

PR #1276, opened 2023-05-31
Allow Authorization Code flow without a client_secret - Initial commit with Patch & testcases

bull500:
Hello @n2ygk 
Here's the PR

Fixes #1092

## Description of the Change
Patch to address -> Allow Authorization Code flow without a client_secret #1092
Edited oauth2_provider/oauth2_validators.py and added testcases to tests/test_oauth2_validators.py
Ran tox. Seems alright

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [x] unit-test added
- [ ] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #1288, opened 2023-06-14
Fix the invalid_client error when request token without the client_secret field

glaucojunior22:
Fixes #1276 

Even after the PR #1276 if the request don't have the `client_secret` field the response will be:
status_code: 401 
{"error": "invalid_client"}
[Demo image](https://i.imgur.com/ADNyQBX.png)

The only way to retrieve the token is sending the client_secret field as an empty string:
[Demo image](https://i.imgur.com/Q5pGvqz.png)

This PR is to allow the first case (request without a client_secret field) to work accordingly to the [RFC specification](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1)

## Reason
The [Oauthlib Request object](https://github.com/oauthlib/oauthlib/blob/master/oauthlib/common.py#L363) always have the client_secret param initialized as None, so `getattr(request, "client_secret", "")` will return None intead of the empty string even if the original request doesn't send the parameter, this is why I added the `or ""` at the end of this line.

## Checklist

- [x ] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [ ] documentation updated
- [x ] author name in `AUTHORS`


n2ygk (approved):
This looks good. Sorry for the delay in reviewing this.

PR #1292, opened 2023-06-22
Expect the remote exp to be defined in time zone UTC conform rfc (Fix…

### LINKED ISSUE #1291: Wrong timezone is applied to expiry token
**Describe the bug**
The `_get_token_from_authentication_server` interprets the expire date wrong. The token is read from `content['exp']` which is defined as NumericDate which in turn is defined as UNIX **UTC** Timestamp (seconds since epoch). The `exp` is converted to a date using `utcfromtimestamp` followed by `make_aware` to turn it into a timezone aware date.  The latter uses the timezone as chosen for you project, while it should use the UTC timezone. Now given a timezone +2 and an access token which expires after an hour, the access token will be invalid because it's in the past.

**To Reproduce**
1. Setup identity provider that serves access token with ACCESS_TOKEN_EXPIRE_SECONDS set to 3600
2. Setup a resource  server with timezone set to TIME_ZONE = "Europe/Amsterdam" (+2)
3. Request Access Token and try to access Resource Server
4. Access Token is always expired

**Expected behavior**
Access token is invalid

**Version**
django-oauth-toolkit==2.3.0

**Additional context**
It's rather easy to fix changing:

```
- expires = make_aware(expires) if settings.USE_TZ else expires`
+ expires = make_aware(expires, timezone=pytz.UTC) if settings.USE_TZ else expires`
```

For backwards compatibility it would be better to has a setting which allows you to changes the timezone specifically for the expire token.

wkleinheerenbrink:
Fixes #1291

## Description of the Change
Interpret the EXP in access tokens as being UTC. Introduces an extra setting `AUTHENTICATION_SERVER_EXP_TIME_ZONE` to allow for the workaround in case of a remote authentication server that, wrongly, does not server the EXP in UTC.

## Checklist

- [x] unit-test added
- [x] documentation updated


dopry:
@wkleinheerenbrink  could you rebase this PR to resolve the conflicts? I'll prioritize reviewing it in the next few days.

makeevolution:
I have this issue too; can I continue this?

n2ygk:
https://docs.python.org/3/library/zoneinfo.html was added in Python 3.9 so the 3.8 tests fail.

setup.cfg
```diff
@@ -40,6 +40,7 @@ install_requires =
 	requests >= 2.13.0
 	oauthlib >= 3.1.0
 	jwcrypto >= 0.8.0
+	pytz >= 2024.1
```
n2ygk, line 43: pytz required when Python < 3.9

n2ygk (approved):
Thanks for this. I've made a few fixes that were causing tests to fail for Python < 3.9

PR #1296, opened 2023-06-30
Issue 1295. Added revert action for migration

yurasavin:
Fixes #
Added reverse action for migration 0006

## Description of the Change
Without reverse action it's impossible to revert applied migration

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/migrations/0006_alter_application_client_secret.py
+++ b/oauth2_provider/migrations/0006_alter_application_client_secret.py
@@ -26,5 +26,5 @@ class Migration(migrations.Migration):
             name='client_secret',
             field=oauth2_provider.models.ClientSecretField(blank=True, db_index=True, default=oauth2_provider.generators.generate_client_secret, help_text='Hashed on Save. Copy it now if this is a new secret.', max_length=255),
         ),
-        migrations.RunPython(forwards_func),
+        migrations.RunPython(forwards_func, migrations.RunPython.noop),
     ]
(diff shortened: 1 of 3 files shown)


oauth2_provider/migrations/0006_alter_application_client_secret.py
```diff
@@ -26,5 +26,5 @@ class Migration(migrations.Migration):
             name='client_secret',
             field=oauth2_provider.models.ClientSecretField(blank=True, db_index=True, default=oauth2_provider.generators.generate_client_secret, help_text='Hashed on Save. Copy it now if this is a new secret.', max_length=255),
         ),
-        migrations.RunPython(forwards_func),
+        migrations.RunPython(forwards_func, migrations.RunPython.noop),
```
n2ygk, line 29: Please add a reverse_func that logs a warning that previously-hashed secrets will fail to revert.

Also please fill out the checklist items (add name to Authors, etc.), update the CHANGELOG, perhaps add a warning about reverse migration of hash secrets, etc.

n2ygk (changes requested):
Thanks for this. See the request to add a reverse_func, documentation, etc. per https://django-oauth-toolkit.readthedocs.io/en/latest/contributing.html#pull-requests

oauth2_provider/migrations/0006_alter_application_client_secret.py
yurasavin: @n2ygk Sorry for the so long delay in response from me. I've added warning log and updated changelog files. Could you check it please?

n2ygk (approved):

PR #1306, opened 2023-08-12
Add an authorize endpoint that uses JSON instead of a Django template…

jhnbyrn:
…/HTML form

Fixes #

## Description of the Change

For single page applications it would be handy to be able to get the data for the authorization page as JSON and then render the authorization page on the client side, and similarly post the results as JSON rather than as a HTML form. 

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -12,6 +12,11 @@ from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.debug import sensitive_post_parameters
 from django.views.generic import FormView, View
 
+# JB
+from django.forms import model_to_dict
+from django.core.serializers.json import DjangoJSONEncoder
+from django.db.models import Model
+
 from ..exceptions import OAuthToolkitError
 from ..forms import AllowForm
 from ..http import OAuth2ResponseRedirect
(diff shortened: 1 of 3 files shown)


oauth2_provider/views/base.py
```diff
@@ -66,7 +66,83 @@ def redirect(self, redirect_to, application):
...
+            if application.skip_authorization:
+                uri, headers, body, status = self.create_authorization_response(
+                    request=self.request, scopes=" ".join(scopes), credentials=credentials, allow=True
+                )
+                return self.redirect(uri, application)
+
+            elif require_approval == "auto":
+                tokens = (
+                    get_access_token_model()
+                    .objects.filter(
```
dashdanw, line 123: you could potentially avoid iteration by filtering by your desired scopes (assuming the `scopes` var here is the required scopes of the view)

```suggestion
                    scopes_filter = { 'scopes__icontains': scope for scope in scopes.split(' ')
                    .objects.filter(
                        user=request.user, 
                        application=kwargs["application"], 
                        expires__gt=timezone.now(),
                        **scopes_filter
                    ).first()
```

dopry:
@jhnbyrn can you rebase this.. i'm open to using a mixin here if it enables you use case and doesn't break existing functionality.

oauth2_provider/views/base.py
```diff
@@ -66,7 +66,83 @@ def redirect(self, redirect_to, application):
 RFC3339 = "%Y-%m-%dT%H:%M:%SZ"
 
 
-class AuthorizationView(BaseAuthorizationView, FormView):
+class AuthorizationMixin:
```
dopry, line 69: Rename AuthorizationMixin to AuthorizationViewMixin.

oauth2_provider/views/base.py
```diff
@@ -66,7 +66,83 @@ def redirect(self, redirect_to, application):
...
+                    if token.allow_scopes(scopes):
+                        uri, headers, body, status = self.create_authorization_response(
+                            request=self.request,
+                            scopes=" ".join(scopes),
+                            credentials=credentials,
+                            allow=True,
+                        )
+                        return self.redirect(uri, application)
+
+        except OAuthToolkitError as error:
```
dopry, line 140: Having the function return the context or response is awkard. It should just return the context. 

let the exception throw and hoist the try/except to the `get` method, so the get method is returning the error_response

dopry (changes requested):

PR #1311, opened 2023-09-07
Allow the use of unhashed secrets

### LINKED ISSUE #1239: Invalid JWT signature because of hashed client secret
**Describe the bug**

I'm using [next-auth](https://next-auth.js.org/) on the frontend and the latest django-oauth-toolkit with OIDC enabled and HS256 as the algorithm on the backend. After authorizing my oauth application I get the error "failed to validate JWT signature" in next-auth.

The JWT token provided by django-oauth-toolkit seems to be incorrect. It's prepared in `oauth2_provider.oauth2_validators.finalize_id_token`. The culprit is the `jwk_key` that's used for signing the token. It's generated with the client secret, which is being hashed since `v2.0.0`.

[https://github.com/jazzband/django-oauth-toolkit/blob/04b5b3e49020168d8decc2c536f3287ad40bcfc3/oauth2_provider/oauth2_validators.py#L845](https://github.com/jazzband/django-oauth-toolkit/blob/04b5b3e49020168d8decc2c536f3287ad40bcfc3/oauth2_provider/oauth2_validators.py#L845)

I'm assuming that hashing was added under the assumption that the raw secret isn't needed anymore after it's created, but that doesn't seem to be the case (see line 224).

[https://github.com/jazzband/django-oauth-toolkit/blob/04b5b3e49020168d8decc2c536f3287ad40bcfc3/oauth2_provider/models.py#L217-L225](https://github.com/jazzband/django-oauth-toolkit/blob/04b5b3e49020168d8decc2c536f3287ad40bcfc3/oauth2_provider/models.py#L217-L225)

**To Reproduce**

I've created a test repo: https://github.com/dnsv/auth-test

1. Setup Django (I'm using Python 3.11):
    1. `poetry install`
    2. `poetry run python manage.py runserver`
    3. `poetry run python manage.py migrate`
    4. `poetry run python manage.py createsuperuser`.
 2. Login into admin.
 3. Create a new Oauth application:
    - Redirect urls: [http://localhost:3000/api/auth/callback/test](http://localhost:3000/api/auth/callback/test)
    - Client type: Confidential
    - Authorization grant type: Authorization code
    - Algorithm: HMAC with SHA-2 256
 4. Create the file `frontend/.env`:
 
    ```
    OAUTH_CLIENT_ID=...
    OAUTH_CLIENT_SECRET=...
    ```

6. Setup Next.js (I'm using node v18.12.1, but i probably works the same with lower versions).
    1. `cd frontend`
    3. `yarn install`
    4. `yarn dev` 
7. Go to http://localhost:3000/ and try to sign it. You'll get the error shown in the console after the authorization step.

Highlights from the backend setup:

```python
# backend/base/settings.py

OAUTH2_PROVIDER = {
    "OIDC_ENABLED": True,
    "OAUTH2_VALIDATOR_CLASS": "backend.oauth2.utils.CustomOAuth2Validator",
    "SCOPES": {
        "openid": "OpenID Connect scope",
    },
}

# backend/oauth2/utils.py

from oauth2_provider.oauth2_validators import OAuth2Validator

class CustomOAuth2Validator(OAuth2Validator):
    oidc_claim_scope = None

    def get_additional_claims(self, request):
        return {
            "id": request.user.id,
            "given_name": request.user.first_name,
            "family_name": request.user.last_name,
            "name": f"{request.user.first_name} {request.user.last_name}",
            "email": request.user.email,
        }

```

**Expected behavior**

Being able to login :)

**Version**

- [x]  I have tested with the latest published release and it's still a problem.
- [ ]  I have tested with the master branch and it's still a problem.

**Additional context**

N/A

gardenerik:
Fixes #1239 

## Description of the Change

Adds an option to disable client secrets hashing on save. Also allows the use of applications with unhashed secrets in the HTTP API. This makes it possible to verify OIDC JWTs with HS256.

## Checklist

- [x] unit-test added
- [x] documentation updated


n2ygk:
Please request a review when you are ready....

gardenerik:
@n2ygk it seems I am unable to request a review through the UI, but feel free to review the PR

n2ygk:
> @n2ygk it seems I am unable to request a review through the UI, but feel free to review the PR

Looks like you added them (or someone did)

gardenerik:
I am not sure what you are talking about :( The reviewers section is empty.

n2ygk:
> I am not sure what you are talking about :( The reviewers section is empty. image

Haha I was looking at *my* suggestions:

image

n2ygk:
@Andrew-Chen-Wang I've added you as a reviewer mostly as an FYI.

gardenerik:
@n2ygk I have updated your points, thanks for the feedback

docs/getting_started.rst
```diff
@@ -244,7 +244,9 @@ Start the development server::
 
 Point your browser to http://127.0.0.1:8000/o/applications/register/ lets create an application.
 
-Fill the form as show in the screenshot below and before save take note of ``Client id`` and ``Client secret`` we will use it in a minute.
+Fill the form as show in the screenshot below and before save take note of ``Client id`` and ``Client secret``, we will use it in a minute.
+
+If you want to use this application with OIDC and ``HS256`` (see :doc:`OpenID Connect <oidc>`), uncheck ``Hash client secret`` to allow verifying tokens using JWT signatures.
```
n2ygk, line 249:
```suggestion
If you want to use this application with OIDC and ``HS256`` (see :doc:`OpenID Connect <oidc>`), uncheck ``Hash client secret`` to allow verifying tokens using JWT signatures. This means your client secret will be stored in cleartext but is the only way to successfully use signed JWT's.
```

n2ygk (changes requested):
Thanks for leaving the campground cleaner than when you found it:-)

I have proposed one minor wording change just to make clear that there's a security exposure to not hashing client secrets.

n2ygk (approved):

n2ygk:
Looks like pytest broke probably due to some version issue. Will look at it tomorrow or Thursday

n2ygk:
Hmm a migrations conflict: 
```
E           django.core.management.base.CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0008_alter_accesstoken_token, 0008_add_hash_client_secret in oauth2_provider).
E           To fix them run 'python manage.py makemigrations --merge'
```

n2ygk:
My fault. I merged #1312 which created the migration conflict. Fixing it now.

hugochinchilla:
I know I can install this from git but I think this fix is important enought to create a new release on PyPI.

Thank you for your work :)

relrod:
@n2ygk (from one ham radio op to another :smiling_face_with_tear:) any chance of a release with this fix in the near-ish future?

n2ygk:
FYI - as part of the release, I'm strengthening the language to recommend using RS256 keys which is a more secure approach than HS256 and does not require retaining the `client_secret` as cleartext.

blag:
> This may lead to incompatibilities with some OIDC Relying Party libraries.

Although I think I know the answer, I have a clarification question. Does the "This" in this sentence mean that:

1. The case when the box is checked (and the client secret is hashed) may lead to incompatibilities with some OIDC RP libraries or...
2. The case when the box is unchecked (and the client secret is left unhashed) may lead to incompatibilities with some OIDC RP libraries

?

Because it isn't 100% perfectly clear here to me, and I'm not an expert on OIDC.

Once I get an answer I'll provide a PR to tweak/fix the language.

gardenerik:
When the box is checked and HS256 is used with OIDC.

HS256 requires that the token is signed by a symmetric key that is shared between the Relying Party and the Authorization Server. The application's secret is used for this signature.

When you leave "hash secret" checked, the Authorization Server will lose the original secret value and will use the hashed secret when signing the token. But this breaks the OIDC's contract as the Relying Party is no longer able to verify the token's signature.

The token is completely valid from the OAuth Toolkit's viewpoint - you can use it to authenticate against Django views, etc.
Good OIDC libraries should check the signature by default, but when doing so, they will reject the provided token, as the signatures mismatch.

You can leave "hash secret" checked without any problems if you:

- use RS256 (the token is signed by `OIDC_RSA_PRIVATE_KEY` and thus verified using the toolkit's public key)
- use HS256, but do not use OIDC, only OAuth2 - OAuth2 does not have any specifications on the token, so it is just treated as a string

blag:
@gardenerik Thanks for the explanation, that's what I thought. I'll whip up a PR to tweak the help text to be a tiny bit more clear about what state it's referring to.

As a slightly separate issue: when a user unchecks the box and uses HS256 with OIDC, does it make sense to leave the (unhashed) symmetric key visible to the user on future page loads? It seems to me that a lot of secret values are only shown once to the user (this is exactly how GitHub works with developer keys), and I assume that this is done to make it more difficult for an attacker to steal or the secret from the OIDC provider and/or spoof the provider entirely (with the correct shared secret).

When a user chooses to leave the "hash secret" option checked when registering an application with django-oauth-toolkit, it looks a bit unpolished to me to display the hashed value in the form again on subsequent page loads, so I'm considering a second PR to only show the secret value once, when it is generated. But I want to make sure I support the HS256+OIDC flow as well in that PR.

Thanks!

PR #1315, opened 2023-09-12
Update RFC URLs to modern location

### LINKED ISSUE #1313: Outdated RFC link in README.rst and elsewhere
**Describe the bug**
The `README.rst` and other files contain an outdated link to RFC 6749.

**To Reproduce**
Visit the github page and note that the URL for the statement about being rfc-compliant leads to a URL at `tools.ietf.org`, a now deprecated domain. Search the source code and find other mentions of `tools.ietf.org`.

**Expected behavior**
The link should point to the current location of the resource, https://www.rfc-editor.org/info/rfc6749 (for the info page with various formats and metadata about the RFC) or https://www.rfc-editor.org/rfc/rfc6749.html (for a direct link to the HTML version).

**Version**
current main branch

**Additional context**
The `tools.ietf.org` site has been retired / deprecated and, while it redirects to a working location, should no longer be used.

jennifer-richards:
Fixes #1313

## Description of the Change
This updates RFC references that pointed at now-deprecated `tools.ietf.org` to use `rfc-editor.org` instead. Several links were updated from http to https. A reference in `oauth2_validators.py` was still referring to an old revision of the pre-RFC draft, so I've updated that to the corresponding section of the published RFC (the text in that section looked unchanged at a glance).

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,7 +33,7 @@ If you are facing one or more of the following:
 Django OAuth Toolkit can help you providing out of the box all the endpoints, data and logic needed to add OAuth2
 capabilities to your Django projects. Django OAuth Toolkit makes extensive use of the excellent
 `OAuthLib <https://github.com/idan/oauthlib>`_, so that everything is
-`rfc-compliant <http://tools.ietf.org/html/rfc6749>`_.
+`rfc-compliant <https://rfc-editor.org/rfc/rfc6749.html>`_.
 
 Reporting security issues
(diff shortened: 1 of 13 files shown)


jennifer-richards:
n.b., I did not update URLs in the changelog on the theory that it's archival

n2ygk (approved):
Thanks!

PR #1316, opened 2023-09-13
Updated Insert user example with scope.

Bellaby:
The Insert user example is the same as the last example saying it will present a permission error. People reading this will get an error trying what is shown as a positive result example.

Fixes #

## Checklist

- [ ] unit-test added
- [x] documentation updated


n2ygk (approved):
Thanks

PR #1333, opened 2023-10-05
fix: AUTHORS alpha sort

dopry:
- L with stroke should be treated as L
- pySilver out of place, do we need a real name here for posterity?

## Description of the Change
fix sort in AUTHORS file

## Checklist

- [N/A] unit-test added
- [N/A] documentation updated
- [N/A] `CHANGELOG.md` updated (only for user relevant changes)


tonial:
I don't think it's an issue if some user stay anonymous.

tonial (approved):

PR #1334, opened 2023-10-06
Fix unhashed secret to work with request body authentication.

amanning9:
## Description of the Change
#1311 allowed the use of unhashed client secrets, but if a secret is unhashed it does not work if the client includes it's client_id and client_secret in the request body, instead of using basic auth- only hashed secrets work in that case.

This PR allows hashed or unhashed secrets in _authenticate_request_body too, and adds a test for this case.

I think this is small enough that it can really just be considered part of 1311 in terms of documentation and changelog- but if I'm wrong please let me know and I can come back.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@ Alan Crosswell
 Alejandro Mantecon Guillen
 Aleksander Vaskevich
 Alessandro De Angelis
+Alex Manning
 Alex Szabó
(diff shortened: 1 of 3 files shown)


n2ygk (approved):
Thanks for catching this and especially for improving code coverage.

Issue #1318, opened 2023-09-18
Refresh Token 500 Internal Server Error instead of the expected 401 Unauthorized


discobeta:
**Describe the bug**
While trying to refresh an OAuth token using an older refresh token, the server returns a 500 Internal Server Error instead of the expected 401 Unauthorized response. This occurs when an attempt is made to refresh the token with an older refresh token after it has been invalidated, resulting in a oauth2_provider.models.AccessToken.DoesNotExist error message.

**To Reproduce**
1. Obtain a new token and a corresponding refresh token.
2. Invalidate the new token by removing it (i.e., leave the "Authorization: Bearer" line empty) and attempt to refresh the token using the refresh token obtained in step 1.
3. Use the new token obtained in step 2 to refresh the token again, successfully this time.
4. Now, try to refresh the token using the original token obtained in step 1.
5. At this point, a 500 Internal Server Error is encountered, with the error message oauth2_provider.models.AccessToken.DoesNotExist: AccessToken matching query does not exist.

**Expected behavior**
Upon attempting to refresh using an older or non-existent token, the server should return a 401 Unauthorized response. The system's current behavior of querying the database to find an AccessToken entry at this stage seems unnecessary and unclear.

**Version**
Django OAuth Toolkit Version: 2.2.0

- [x] I have tested with the latest published release and it's still a problem.
- [x] I have tested with the master branch and it's still a problem.

**Additional context**
When attempting to refresh an non-existing token the response should be Unauthorized

dopry:
@discobeta, Do you have a stack trace so we can see what is raising the error?

discobeta:
hi @dopry thanks for following up here.

The error is 500, see the complete traceback below.

```Traceback (most recent call last):

  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner

    response = get_response(request)

               ^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response

    response = wrapped_callback(request, *callback_args, **callback_kwargs)

               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/usr/local/lib/python3.11/site-packages/sentry_sdk/integrations/django/views.py", line 84, in sentry_wrapped_callback

    return callback(request, *args, **kwargs)

[...]

PR #1337, opened 2023-10-12
Fix access token 500

discobeta:
Fixes https://github.com/jazzband/django-oauth-toolkit/issues/1318

## Description of the Change
We are now try / except when attempting to find a vali token to avoid a 500 `oauth2_provider.models.AccessToken.DoesNotExist: AccessToken matching query does not exist`

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,6 +26,7 @@ Antoine Laurent
 Anvesh Agarwal
 Aristóbulo Meneses
 Aryan Iyappan
+Asaf Klibansky
 Ash Christopher
 Asif Saif Uddin
 Bart Merenda
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * #1273 Add caching of loading of OIDC private key.
 * #1285 Add post_logout_redirect_uris field in application views.
 * #1311 Add option to disable client_secret hashing to allow verifying JWTs' signatures.
+* #1337 Gracefully handle expired or deleted refresh tokens, in `validate_user`.
 
 - ### Fixed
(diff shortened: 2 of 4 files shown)


dopry:
@discobeta could you review your tests? It looks like the new change isn't covered.

discobeta:
added additional tests

dopry:
@discobeta, It looks like https://github.com/jazzband/django-oauth-toolkit/pull/1337/checks?check_run_id=17910869256 still isn't covered. You probably need to setup a scenario where you make a valid AccessToken, delete the token, then make a request.

discobeta:
Added a test for a deleted token

dopry:
The except branch added by your patch is still not covered, https://app.codecov.io/gh/jazzband/django-oauth-toolkit/pull/1337?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=jazzband

discobeta:
I pushed an update that covers this change. Can you please approve the
workflow to recheck coverage?

discobeta:
@dopry It looks like this is now covered. Thank you for your recommendation and support with this.

https://app.codecov.io/gh/jazzband/django-oauth-toolkit/pull/1337?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=jazzband#diff-b2F1dGgyX3Byb3ZpZGVyL29hdXRoMl92YWxpZGF0b3JzLnB5

dopry:
I think the last thing we need is a changelog entry and adding yourself to Authors if you're not in there already.

discobeta:
@dopry can you please take a look here and see if there are any further changes required?

dopry (approved):

PR #1347, opened 2023-10-21
Added oauth2_settings.RESOURCE_SERVER_INTROSPECTION_RESPONSE_FIELD setting to let po…

### LINKED ISSUE #1325: Add the possibility to map user info with user model field.
Hello,

I would like to know if it would interesting to let possibility to fill the UserModel with some user information just after the authentication server. This could be a custom map.

Indeed the keycloack întrospection (and all standard oauth2 introspection in general) return many informations about the user than It would be nice to map here.

More specificly, in my case the USERNAME_FIELD is map to the uuid field of my user model and this feature will permit to me to not add a pre-save traietment to the save of my model to prevent null username field.

The line concerned:
https://github.com/jazzband/django-oauth-toolkit/blob/9aa27c7528cdeda0b85bac5a8a00b39d696a43f9/oauth2_provider/oauth2_validators.py#L387C1-L387C21

fijemax:
Fixes #1325

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -383,7 +383,11 @@ class OAuth2Validator(RequestValidator):
         if "active" in content and content["active"] is True:
             if oauth2_settings.RESOURCE_SERVER_INTROSPECTION_RESPONSE_FIELD in content:
                 user, _created = UserModel.objects.get_or_create(
-                    **{UserModel.USERNAME_FIELD: content[oauth2_settings.RESOURCE_SERVER_INTROSPECTION_RESPONSE_FIELD]}
+                    **{
+                        UserModel.USERNAME_FIELD: content[
+                            oauth2_settings.RESOURCE_SERVER_INTROSPECTION_RESPONSE_FIELD
+                        ]
+                    }
                 )
             else:
                 user = None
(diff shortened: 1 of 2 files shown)


dopry:
@fijemax, I think I see your need here. The user is being fetched or created here. The current implementation uses username as the key field. Username may not be set as it's an optional in the response and frankly we don't know the required fields on someones usermodel so this approach is a bit naieve.  

I think this approach might work for your use case, but at the moment I don't think I have enough context to be confident this is a good fix that will work for the wider community. I'm gonna do some digging around this method so I better understand how it's used. Can you please provide more detail on your use case in your original issue to provide more context?

@n2ygk, do you have thoughts?

n2ygk:
@dopry I have not been following this issue. Isn't getting the sub or other added claims described [here](https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html#adding-claims-to-the-id-token)?

n2ygk:
Isn't this accomplished by #1325?

fijemax:
Hello,
What do you mean ?

n2ygk:
> Hello, What do you mean ?

@fijemax See #1325 which allows setting the username field value. This looks to me to address the same issue you are addressing here.

n2ygk:
@fijemax please edit the PR description. I've initialized it with the PR template.

n2ygk:
I believe this feature is resolved by #1325

PR #1350, opened 2023-10-27
Update supported versions, remove eol dj22 and py37, add dj5 and py12

adamchainz:
## Description of the Change

Drop EOL Python 3.7 and Django 2.2.

Add Python 3.12 and Django 5.0.

I looked back at #1264 to figure out the changes required.

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [x] unit-test added
- [x] documentation updated
- [x] `CHANGELOG.md` updated (only for user relevant changes)
- [x] author name in `AUTHORS`

PR #1219, opened 2022-10-27
Parallelize CI workflow by adding Django versions to job matrix

joshuadavidthomas:
## Description of the Change
Adds an additional job matrix that includes all the Django versions this package supports and excludes certain combinations of Python and Django from running. Also removes the `max-parallel` restriction for parallel CI jobs.

This should dramatically speed up CI tests, cutting almost 20 minutes from the total runtime.

Note: I have another open PR (#1218) that adds Python 3.11 to CI and tox. Either this PR or that one will need to be rebased once either one is accepted and merged.

### Before

**~25 minutes**

image

### After

**~8 minutes**

image

## Checklist

- [x] PR only contains one change (considered splitting up PR)
- [ ] unit-test added
- [ ] documentation updated
- [ ] `CHANGELOG.md` updated (only for user relevant changes)
- [ ] author name in `AUTHORS`

PR #1352, opened 2023-10-27
Speed up tests ~2x

adamchainz:
## Description of the Change

Speed up the test suite to be about twice as fast, using these techniques:

1. Move model creation into `setUpTestData()`, so it runs only once per test case. See [my post](https://adamj.eu/tech/2021/04/12/how-to-convert-a-testcase-from-setup-to-setuptestdata/).
2. Remove `tearDown()` methods that removed model instances. These were always unnecessary as Django resets the database for you.
3. Move immutable `RequestFactory`s to class-level variables.

Measuring on the Python 3.12 and Django 5.0 environment (added in #1350), I got 207s before these changes and 105s afterwards, approximately twice as fast.

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/tests/test_application_views.py
+++ b/tests/test_application_views.py
@@ -14,13 +14,10 @@ UserModel = get_user_model()
 
 
 class BaseTest(TestCase):
-    def setUp(self):
-        self.foo_user = UserModel.objects.create_user("foo_user", "test@example.com", "123456")
-        self.bar_user = UserModel.objects.create_user("bar_user", "dev@example.com", "123456")
-
-    def tearDown(self):
-        self.foo_user.delete()
-        self.bar_user.delete()
+    @classmethod
+    def setUpTestData(cls):
+        cls.foo_user = UserModel.objects.create_user("foo_user", "test@example.com", "123456")
+        cls.bar_user = UserModel.objects.create_user("bar_user", "dev@example.com", "123456")
 
 
 @pytest.mark.usefixtures("oauth2_settings")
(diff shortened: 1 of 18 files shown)


adamchainz:
`setUpTestData` copying was only added in Django 3.2, before that the [djagno-testdata package](https://github.com/charettes/django-testdata) is needed. I think we should merge #1350 first to drop Django 2.2 support, then I'll rebase this.

dopry (approved):
@adamchainz As a maintainer who updates a lot of branches and spends a lot of time waiting on CI spinners, I can't tell you how much I appreciate this!

n2ygk:
@adamchainz @dopry not sure why but I've noticed that all the tests run and then the last two `success` jobs seem to hang for a long time waiting for a runner. Maybe because `success` is a different job from `build`? Is there a way to add the `Success` step into the build job somehow?  I'm not sure how to skip the matrix part within the `build` job...

dopry:
@n2ygk, Success needs to wait on the matrix build to complete.  It can't be bundled into the same. can you open a new issue linking to an example?

n2ygk:
@dopry not point opening a new issue if we are stuck with the current situation. Still better than the old way.

dopry:
If you can't point to an example of the issue I can't troubleshoot it further and I suspect it is not directly related to reducing the overheat to setup test data, but is more likely related to capacity or organizational throttling by GH.  Correlation is not causation.

n2ygk:
@dopry https://github.com/jazzband/django-oauth-toolkit/issues/1376

n2ygk:
But I think I should have been commenting on #1219

PR #1357, opened 2023-11-04
Move signal import to django.core

AlanCoding:
## Description of the Change

Hi :wave: , I believe this is causing us a problem, but I lack technical rigor to be able to explain it fully. By applying this change under certain circumstances, I seem to avoid a core dump that happens as uWSGI recycles workers when it is running in the default prefork mode. The most relevant description of the series of events is this comment: https://github.com/unbit/uwsgi/issues/1969#issuecomment-462252471

Here, I just wish to present this as an improvement to follow best practice. In the `django.test.signals` module, you can see it imports from this modification with no modifications.

https://github.com/django/django/blob/main/django/test/signals.py

However, somewhat obviously... if you import this test _module_, you will also register all the test signal connections. Now, that _shouldn't_ do anything because the `setting_changed` signal is documented to only fire when running tests. However, it also pulls in a lot of _other_ imports which just aren't needed, and as this happens in settings, it triggers fairly early in app load order.

This argument also applies to 1 import in DRF, which I also hope to submit a patch for. I'm still working out final testing on my side, so consider this a draft as of opening.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Adam Johnson
 Adam Zahradník
 Adheeth P Praveen
 Alan Crosswell
+Alan Rominger
 Alejandro Mantecon Guillen
 Aleksander Vaskevich
 Alessandro De Angelis
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * #1284 Allow to logout with no id_token_hint even if the browser session already expired
 * #1296 Added reverse function in migration 0006_alter_application_client_secret
 * #1336 Fix encapsulation for Redirect URI scheme validation
+* #1357 Move import of setting_changed signal from test to django core modules
 
 ### Removed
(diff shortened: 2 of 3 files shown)


AlanCoding:
For some more background - here is an issue where they requested to move this signal out of tests https://code.djangoproject.com/ticket/20349

And in 2014 it was implemented in https://github.com/django/django/commit/5dddd79433ceb88ab67d9851b49a44ce5b8f509c

Nonetheless it seems plausible that libraries still have the old import, and that it has even proliferated.

AlanCoding:
It turns out that the change was already adopted into Django Rest Framework here https://github.com/encode/django-rest-framework/pull/8699

dopry:
@AlanCoding  this looks like a reasonable change. I don't think additional unit tests are needed here as long as tests continue passing. Can you update your branch to add a note to the CHANGELOG.md and add yourself to AUTHORS?

dopry (approved):
Thanks! Your contributIon is appreciated!

PR #1361, opened 2023-11-10
fix: prompt=none shows a login screen

### LINKED ISSUE #1268: prompt=none redirects to login screen.
**Describe the bug**
When trying to do silent login or prompt=none login /o/authorize redirects to the login screen. 

This is because it oauth2_provider.views.AuthorizationView extends LoginRequiredMixin via BaseAuthorizationView. 

**To Reproduce**

setup any OIDC silent auth enabled client, I used https://github.com/dopry/svelte-oidc, with the django oauth toolkit and enable the chrome debugger in your browser, login then reload the client. 

You will see a request to http://127.0.0.1:8000/o/authorize/... in an iframe with a 302 Location redirect to  /admin/login?next=/o/authorize/

you will also see in the console `fused to display 'http://127.0.0.1:8000/' in a frame because it set 'X-Frame-Options' to 'deny'.`

**Expected behavior**

per https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

> prompt - OPTIONAL. Space delimited, case sensitive list of ASCII string values that specifies whether the Authorization Server prompts the End-User for reauthentication and consent. The defined values are:
> - none

>   The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be login_required, interaction_required, or another code defined in [Section 3.1.2.6](https://openid.net/specs/openid-connect-core-1_0.html#AuthError). This can be used as a method to check for existing authentication and/or consent.

and there should be no xframe deny error. 
 
**Version**

25f6de50d4b7287a76c87e1cf511d8f9a61c6faa

- [X] I have tested with the master branch and it's still a problem.

**Additional context**

andyzickler:
Fixes #1268

## Description of the Change

Fix bug preventing support for Silent Authentication. If an unauthorized request to `AuthorizationView` with a query parameter that contains `prompt=none` happens, then we will redirect with an error code of `login_required` otherwise everything will proceed as before. 

See https://auth0.com/docs/authenticate/login/configure-silent-authentication#error-responses
and https://openid.net/specs/openid-connect-core-1_0.html#AuthError

fully supporting prompt=none will require implementing validate_silent_login in the validator. this doesn't implement that, but will allow people to implement it if they want until we can implement a good implementation for DOT. 

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/tests/test_authorization_code.py
+++ b/tests/test_authorization_code.py
@@ -572,8 +572,6 @@ class TestOIDCAuthorizationCodeView(BaseTest):
         self.assertIn("scope=openid", next)
         self.assertIn("redirect_uri=http%3A%2F%2Fexample.org", next)
 
-
-
     def test_id_token_skip_authorization_completely(self):
         """
         If application.skip_authorization = True, should skip the authorization page.
--- a/tests/test_authorization_code.py
+++ b/tests/test_authorization_code.py
@@ -545,6 +545,35 @@ class TestAuthorizationCodeView(BaseTest):
 
 @pytest.mark.oauth2_settings(presets.OIDC_SETTINGS_RW)
 class TestOIDCAuthorizationCodeView(BaseTest):
+    def test_login(self):
+        """
+        Test login page is rendered if user is not authenticated
+        """
+        self.oauth2_settings.PKCE_REQUIRED = False
+
+        query_data = {
+            "client_id": self.application.client_id,
+            "response_type": "code",
+            "state": "random_state_string",
+            "scope": "openid",
+            "redirect_uri": "http://example.org",
+        }
+        path = reverse("oauth2_provider:authorize")
+        response = self.client.get(path, data=query_data)
+        # The authorization view redirects to the login page with the
+        self.assertEqual(response.status_code, 302)
+        scheme, netloc, path, params, query, fragment = urlparse(response["Location"])
+        self.assertEqual(path, settings.LOGIN_URL)
+        parsed_query = parse_qs(query)
+        next = parsed_query["next"][0]
+        self.assertIn(f"client_id={self.application.client_id}", next)
+        self.assertIn("response_type=code", next)
+        self.assertIn("state=random_state_string", next)
+        self.assertIn("scope=openid", next)
+        self.assertIn("redirect_uri=http%3A%2F%2Fexample.org", next)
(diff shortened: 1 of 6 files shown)


andyzickler:
This PR was a bit optimistic and naïve. Tests written via spec passed, but not in real world testing with the ldp and rp apps. Future implementation likely needs to implement oauthlib's `validate_silent_login`

dopry:
I think it was on the right track. Let's isolate the dispatch fixes so someone could at least in theory implement validate_silent_login

dopry:
@n2ygk @tonial I'd love to get a review from you guys on this. I've been working on it with @andyzickler. This bug is blocking an SSO implementation for me I'd really like to complete.

tests/test_authorization_code.py
```diff
@@ -645,6 +672,33 @@ def test_prompt_login(self):
...
+        }
+
+        response = self.client.get(reverse("oauth2_provider:authorize"), data=query_data)
+
+        self.assertEqual(response.status_code, 302)
+
+        scheme, netloc, path, params, query, fragment = urlparse(response["Location"])
+        parsed_query = parse_qs(query)
+
+        self.assertIn("login_required", parsed_query["error"])
```
tonial, line 699: You could use `self.assertEqual(parsed_query, expected_dict)` to be sure we don't add anything else.

tests/test_authorization_code.py
```diff
@@ -645,6 +672,33 @@ def test_prompt_login(self):
...
+        Should redirect to redirect_uri with an error of login_required
+        """
+        self.oauth2_settings.PKCE_REQUIRED = False
+
+        query_data = {
+            "client_id": self.application.client_id,
+            "response_type": "code",
+            "state": "random_state_string",
+            "scope": "read write",
+            "redirect_uri": "http://example.org",
```
tonial, line 688: Maybe add a query parameter, or add another test with a query parameter in the redirect_url to test the `seperator` computed line 268 in `handle_no_permission()`

tonial (approved):
I made a few comments on the tests, but the implementation seems correct to me.

Great addition !

PR #1364, opened 2023-11-10
Fix/wellknown openid configuration no trailing slash

### LINKED ISSUE #1363: /.well-known/openid-configuration is redirected to /.well-known/openid-configuration/
**Describe the bug**
https://openid.net/specs/openid-connect-discovery-1_0.html#rfc.section.4 
specifies the path as /.well-known/openid-configuration

currently when one requests that path they are redirected to /.well-known/openid-configuration
which slows discovery on the client and can cause cors issues with the redirect. 

**To Reproduce**
start up tests/app/idp and open http://localhost:8000/o/.well-known/openid-configuration in your browsers, you will be redirected to 
http://localhost:8000/o/.well-known/openid-configuration/

**Expected behavior**
The JSON is returned and there is no redirect.

**Version**
2.3.0

dopry:
Fixes #1363 

## Description of the Change

Update endpoint to handle both with and without trailing slash, add comment noting the tailing slash is deprecated and should be removed in the future. 

## Checklist

- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/oidc.rst
+++ b/docs/oidc.rst
@@ -407,7 +407,7 @@ the URLs accordingly.
 ConnectDiscoveryInfoView
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Available at ``/o/.well-known/openid-configuration/``, this view provides auto
+Available at ``/o/.well-known/openid-configuration``, this view provides auto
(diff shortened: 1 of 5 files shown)


tonial (approved):
:+1:

PR #1375, opened 2023-12-18
Update index.rst requirements

suspiciousRaccoon:
I simply copy pasted the ones from the [README.rst](https://github.com/jazzband/django-oauth-toolkit/blob/master/README.rst)

## Checklist

- [ ] unit-test added
- [X] documentation updated


n2ygk (approved):
Thanks for this. It would be DRYer if we were able to have one file listing these requirements and include it in the markdown, RST, and setup.cfg but that appears to be not something that can easily be done.

PR #1382, opened 2024-01-02
Fixed Form UrlEncoded OAuth Lib Core

### LINKED ISSUE #296: JSONOAuthLibCore breaks multipart form-data request for django-rest-framework

ferrants:
Fixes #296 

## Description of the Change

Adds another Oauth Lib Core that isn't broken for base64 encoded form submissions.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/oauth2_backends.py
+++ b/oauth2_provider/oauth2_backends.py
@@ -1,5 +1,6 @@
 import json
-from urllib.parse import urlparse, urlunparse
+import base64
+from urllib.parse import urlparse, urlunparse, unquote
 
 from oauthlib import oauth2
 from oauthlib.common import Request as OauthlibRequest
@@ -236,6 +237,21 @@ class JSONOAuthLibCore(OAuthLibCore):
             body = ""
 
         return body
+
+class JSONAndFormUrlencodedOAuthLibCore(JSONOAuthLibCore):
+    def extract_body(self, request):
+        # fixes base64 encoded form-submission. you can't control what all oauth clients use.
+        # if request.content_type in ['multipart/form-data', 'application/x-www-form-urlencoded']:
+        if request.content_type in ["application/x-www-form-urlencoded"]:
+            query_string = base64.b64decode(request.body).decode("utf-8")
+            query_params = {p.split("=")[0]: p.split("=")[1] for p in query_string.split("&")}
+            if "redirect_uri" in query_params:
+                query_params["redirect_uri"] = unquote(query_params["redirect_uri"])
+            res = query_params.items()
+
+            return res
+
+        return super(OAuthLibCoreFixed, self).extract_body(request)
 
 
 def get_oauthlib_core():


oauth2_provider/oauth2_backends.py
```diff
@@ -238,6 +239,22 @@ def extract_body(self, request):
         return body
 
 
+class JSONAndFormUrlencodedOAuthLibCore(JSONOAuthLibCore):
+    def extract_body(self, request):
+        # fixes base64 encoded form-submission. you can't control what all oauth clients use.
+        # if request.content_type in ['multipart/form-data', 'application/x-www-form-urlencoded']:
+        if request.content_type in ["application/x-www-form-urlencoded"]:
+            query_string = base64.b64decode(request.body).decode("utf-8")
+            query_params = {p.split("=")[0]: p.split("=")[1] for p in query_string.split("&")}
```
ferrants, line 248: There's a safer way to do this

oauth2_provider/oauth2_backends.py
```diff
@@ -238,6 +239,22 @@ def extract_body(self, request):
...
+            if "redirect_uri" in query_params:
+                query_params["redirect_uri"] = unquote(query_params["redirect_uri"])
```
ferrants, line 250: can we assume all of them need to be unquoted?

oauth2_provider/oauth2_backends.py
```diff
@@ -238,6 +239,22 @@ def extract_body(self, request):
         return body
 
 
+class JSONAndFormUrlencodedOAuthLibCore(JSONOAuthLibCore):
+    def extract_body(self, request):
+        # fixes base64 encoded form-submission. you can't control what all oauth clients use.
+        # if request.content_type in ['multipart/form-data', 'application/x-www-form-urlencoded']:
```
ferrants, line 245: remove additional line or address the non-encoded one

PR #1394, opened 2024-01-29
Don't access oauth2_settings if OAuthLibCore server parameter is prov…

claytonsingh:
## Description of the Change
Dont access oauth2_settings if server parameter is provided

## Checklist
- [x] unit-test added
- [x] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Bart Merenda
 Bas van Oostveen
 Brian Helba
 Carl Schwan
+Clayton Singh
 Daniel Golding
 Daniel 'Vector' Kerr
 Darrel O'Pry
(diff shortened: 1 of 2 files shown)


n2ygk:
@claytonsingh what is the purpose of this change? How and why would one do this?

claytonsingh:
I have some specific requirements and am not using this library in `INSTALLED_APPS`. The methods and classes do provide extensive value though.

n2ygk (changes requested):
Please add a use case example to the documentation and add a test.

I can't merge something that doesn't have an clear (to others) purpose.
Thanks.

dopry:
@claytonsingh bump, do you have time to wrap up this PR? Id be good to get the use case documented and some tests around the feature.

PR #1396, opened 2024-01-31
Update oidc.rst: match example file name to import

elohmeier:
Just a small typo fix. If users copy the example module reference from below they can avoid the import error if the file is named correctly.

## Checklist

- [ ] unit-test added
- [x] documentation updated


n2ygk (approved):
Thanks!

PR #1366, opened 2023-11-11
feat: containerized apps

dopry:
## Description of the Change

- add Dockerfile to build the tests/app/idp
- add Dockerfile to build the tests/app/rp
- add a docker compose file to build and start instances of the idp and rp

You can test this setup with
```docker-compose up```

This could be nice for testing new branches with the following

```
docker-compose down -v # remove existing env
docker-compose build      # rebuild the images
docker-compose up          # spin up the docker  compose environment
```

You should be able to goto http://localhost:5173 and http://localhost:8000 to access the compose instances. 

This is a precursor to publishing the test IDP as a container via docker hub. I hope making a Django based IDP more accessible will help grow our contributor and user community. 

## Checklist

- [X] PR only contains one change (considered splitting up PR)
- [N/A ] unit-test added
- [ ] documentation updated

   I wasn't quite sure where or how to document this at this juncture. I'm open to ideas.

- [N/A] `CHANGELOG.md` updated (only for user relevant changes)
- [X] author name in `AUTHORS`

PR #1430, opened 2024-06-12
fix: npm install waa failing

dopry:
it seems we merged a major update from dependabot to some of the RP dependencies that were incompatible with some of the other package versions. This resolves that and updates the RP to Svelte4/SvelteKit 2.

Test by following the instructions in /test/app/README.md to launch the local environment.


dopry:
@n2ygk This should be ready to merge. If you can take a quick look. I updated the Deps for the RP and added some build jobs so we'll catch if dependabot breaks builds in the RP via CI. I'd like to get this in first then I can rebase https://github.com/jazzband/django-oauth-toolkit/pull/1366

n2ygk:
@dopry looks like this workflow failed:

[Invalid workflow file: .github/workflows/test.yml#L120](https://github.com/jazzband/django-oauth-toolkit/actions/runs/9485770246/workflow)
The workflow is not valid. .github/workflows/test.yml (Line: 120, Col: 9): Job 'success' depends on unknown job 'test-rp'.

.github/workflows/test.yml
```diff
@@ -84,8 +84,40 @@ jobs:
...
+
+      - name: Run build
+        run: npm run build
+        working-directory: tests/app/rp
+
   success:
-    needs: build
+    needs:
+      - test-package
+      - test-rp
```
n2ygk, line 120:
```suggestion
      - test-demo-rp
```
Is this what it should be?

n2ygk (changes requested):

dopry:
That's what I get for doing last minute rename changes

n2ygk (approved):

PR #1435, opened 2024-06-20
Drop re_path from installation guide

jaap3:
## Description of the Change

Removed the bit about using `re_path` to add urls. It's just confusing, as a user I don't know why I should pick this option over the first one.

## Checklist

- [ ] unit-test added
- [x] documentation updated

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -26,17 +26,6 @@ If you need an OAuth2 provider you'll want to add the following to your :file:`u
         path('o/', include('oauth2_provider.urls', namespace='oauth2_provider')),
     ]
 
-Or using ``re_path()``
-
-.. code-block:: python
-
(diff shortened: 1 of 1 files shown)


jaap3:
Just a note, including the urls like this:

```
from oauth2_provider import urls as oauth2_urls

urlpatterns = [
    path("o/", include(oauth2_urls))
]
```

works just as well, and removes the need to set a namespace (`app_name` in the urls module does the job).

n2ygk (changes requested):
Please make the change you mentioned in your comment since keeping things simple is best:

```
from oauth2_provider import urls as oauth2_urls

urlpatterns = [
    path("o/", include(oauth2_urls))
]
```

n2ygk (approved):
See #1436 for more

PR #1445, opened 2024-07-31
use user.pk instead of user.id

sahama:
Fixes :
 * use user.pk instead of user.id

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -102,6 +102,7 @@ Rustem Saiargaliev
 Sandro Rodrigues
 Shaheed Haque
 Shaun Stanworth
+Sayyid Hamid Mahdavi
 Silvano Cerza
 Sora Yanai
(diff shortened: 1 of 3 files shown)


n2ygk:
Please explain why this is needed. Is there an open issue that this fixes?

sahama:
> Please explain why this is needed. Is there an open issue that this fixes?

django allow's to use custom primary key. and in django oauth toolkit it is ok to use custom Application and other models.
so for example if i use uuid as primary key and name primary key as uid or gid, this accure bug in some cases. 

```
class Application(AbstractApplication):
    id=None
    uid = models.UUIDField(primary_key=True, editable=False, default=uuid.uuid4, unique=True, db_index=True)

```

PR #1446, opened 2024-07-31
models pk instead of models id

sahama:
Fixes #
* Using pk instead of id.
## Checklist

- [ ] unit-test added
- [] documentation updated


n2ygk:
Is this a repeat of #1445?

sahama:
> Is this a repeat of #1445?

No. #1445 refer to custom primary key in user model and this is about custom primary key in django-oauth-toolkit models.

n2ygk:
> > Is this a repeat of #1445?
> 
> No. #1445 refer to custom primary key in user model and this is about custom primary key in django-oauth-toolkit models.

Ah, I see. Since these changes are closely related, can you please put the #1445 changes in here as well? Thanks.

oauth2_provider/models.py
```diff
@@ -234,8 +234,9 @@ def clean(self):
             ):
                 raise ValidationError(_("You cannot use HS256 with public grants or clients"))
 
+    # TODO: I removed usage of this in templates. so it can label as deprecated.`
```
n2ygk, line 237: What's this new TODO comment about?

n2ygk (changes requested):
Thanks for this!  Please take a look at the one question I have and resolve that.

oauth2_provider/models.py
n2ygk: PS: Also please close #1445 since you've moved the change over to this PR.

sahama: Method `get_absolute_url` one time used in one template for getting url of a page. I used `{% url %}` tag bug i suggest to change it to deprecated instead of removing it. It is unusual to remove or change behavior of a method suddenly.

sahama:
> > > Is this a repeat of #1445?
> > 
> > 
> > No. #1445 refer to custom primary key in user model and this is about custom primary key in django-oauth-toolkit models.
> 
> Ah, I see. Since these changes are closely related, can you please put the #1445 changes in here as well? Thanks.

ok

oauth2_provider/models.py
n2ygk:
```suggestion
```
Suggest you remove the comment and just leave the `get_absolute_url` in case someone may be using it elsewhere (e.g. in their own template or other code based on the application model class). It's probably not worth deprecating it and then keeping track that it has to get removed in a future release.

n2ygk (changes requested):

n2ygk (approved):

PR #1447, opened 2024-08-02
change token to TextField in AbstractAccessToken model

### LINKED ISSUE #1412: Remove 255 Character Limit on Tokens to Support JWT with Additional Claims
### Problem Description
When using django-oauth-toolkit to issue JWT tokens, the current implementation imposes a 255 character limit on token size. This restriction becomes problematic when adding additional claims to the JWT, such as user roles, permissions, or other user-specific data. For example, including a longer username or additional claims exceeds the limit, causing the application to crash.

### Proposed Solution
I propose removing the 255 character limit on tokens. JWT tokens are designed to be extensible and should support a variable length to accommodate different use cases. By removing this limit, django-oauth-toolkit can offer more flexibility in issuing JWTs, making it a more robust solution for modern OAuth 2.0 applications that rely on JWT for extensive user claims.

### Example Scenario
Below is an example scenario where the current token size limit is problematic:

```python
from datetime import datetime, timedelta, timezone
import jwt

def generate_jwt_token(request, refresh_token=None):
    user = request.user
    exp_time = datetime.now(timezone.utc) + timedelta(seconds=settings.OAUTH2_PROVIDER.get('ACCESS_TOKEN_EXPIRE', 3600))

    claim = {
        'user_id': user.id.__str__(),
        'username': user.username,
        'exp': exp_time,
    }
    token = jwt.encode(claim, settings.SECRET_KEY, algorithm='HS256')
    return token
```

In this scenario, if we add more items to the claim or if the `username` is longer, the token size can easily exceed 255 characters, leading to application failures.

### Benefits
- **Flexibility:** Allows developers to include necessary information in the JWT without worrying about hitting the size limit.
- **Security:** Larger tokens can include more detailed claims, improving security by precisely defining access controls.
- **Compatibility:** Ensures compatibility with standards that do not impose such limits on token size, making django-oauth-toolkit more versatile.

### Conclusion
Removing the 255 character limit on tokens in django-oauth-toolkit will provide developers with the needed flexibility to use JWTs effectively in their applications. This change will make the toolkit a more adaptable and forward-looking solution for OAuth 2.0 implementations.

fazeelghafoor:
Fixes #1412 

## Description of the Change

- Changed token field to TextField and added TokenChecksumField in AbstractAccessToken model.
- Updated middleware, validators, and views to use token checksums.
- Modified test migrations to include token_checksum in sampleaccesstoken model.
- Added tests for token checksum field.

## Checklist

- [X] unit-test added
- [ ] documentation updated


fazeelghafoor:
@n2ygk please review and let me know if anything needs to change

n2ygk:
@dopry since I see you are actively looking at DOT right now, would you mind reviewing this PR please?

n2ygk:
oops my rebase messed up the migrations. will fix it.

n2ygk (approved):

n2ygk:
I'm embarrassed to say I've made yet another error. This time I'm running tox locally before pushing!

PR #1449, opened 2024-08-06
replace obsolete m2r with sphinx_mdinclude

a-detiste:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,7 +41,7 @@ extensions = [
     "sphinx.ext.todo",
     "sphinx.ext.coverage",
     "rfc",
-    "m2r",
+    "sphinx_mdinclude",
     "sphinx_rtd_theme",
 ]
 
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,6 @@
 Django
 oauthlib>=3.1.0
-m2r>=0.2.1
-mistune<2
+sphinx_mdinclude
 sphinx==7.2.6
 sphinx-rtd-theme==1.3.0
 -e .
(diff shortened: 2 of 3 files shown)


n2ygk (approved):
thanks

n2ygk (changes requested):
RTD build fails with this change. See https://readthedocs.org/projects/django-oauth-toolkit/builds/25221545/

a-detiste:
Can we try a no-change CI run as comparison basis ?

a-detiste:
It replaces the perfeclty fine docutils by an older one that is too old.

https://github.com/omnilib/sphinx-mdinclude/commit/db37d3f36d4d3ecaaaab4797f0139f761d7983d0

Downloading docutils-0.18.1-py2.py3-none-any.whl (570 kB)

a-detiste:
The pinned libs are too old.

sphinx==7.2.6
sphinx-rtd-theme==1.3.0

n2ygk:
> The pinned libs are too old.
> 
> sphinx==7.2.6 sphinx-rtd-theme==1.3.0

So please submit an update commit that pins the appropriate versions. the mdinclude version should also be pinned.

a-detiste:
I have absolutely no idea what an "appropriate version" shoud be. I would simply drop pinning altogether & see what happens.

n2ygk:
> I have absolutely no idea what an "appropriate version" shoud be. I would simply drop pinning altogether & see what happens.

The usual approach is to upgrade everything and make sure it works. Then pin those versions (pip freeze).

dopry:
@a-detiste thanks for your contribution! Do you have time to wrap up this PR? We'd really like to get it in.

dopry:
Closing. the branch is out of date and OP is non-responsives.

PR #1450, opened 2024-08-07
Honor database assignment from router

### LINKED ISSUE #868: Multi database setup w/ read only endpoints
We're attempting to migrate our app to use read only db endpoints. In our implementation we decided that the read only endpoint should be the default endpoint i.e. you must explicitly decide you want to write. Which to us seems to make sense. Also we implemented our own database router as follows. [Router docs](https://docs.djangoproject.com/en/3.1/topics/db/multi-db/#database-routers). Which handles the direction of individual models read/writes.

```python
class Router:
    def db_for_write(model, **hints):
        return "writer"

    def db_for_read(model, **hints):
        return "default"
```

This appears to be perfectly fine for areas where transactions are not used. However when transactions are used we get an error. 

```bash
raise TransactionManagementError('select_for_update cannot be used outside of a transaction.')
django.db.transaction.TransactionManagementError: select_for_update cannot be used outside of a transaction
```

This emanates from `Oauth2Validator.save_bearer_token` which opens a [transaction](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L550)  against`"default"` and at various points will call `select_for_update`. This can also happen when `AbstractRefreshToken.revoke` is called as it also opens a [transaction](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/models.py#L406).

Since we can't specify the `using` argument in the `transaction.atomic` we have resorted to implementing our own validator and refresh token model. Overriding the methods and copy pasting the original implementation with one that allows us to specify the `using` argument. For obvious reasons (we don't want to have to validate the implementation hasn't changed each oauth toolkit release) this is not ideal.

That all being said my initial questions are

What ways have other users solved this?

Is there a general Django convention for transaction connection management? e.g. assume you can always write against default

shaleh:
Fixes #868 

## Description of the Change
The token models might not be stored in the default database. There might not _be_ a default database.
Instead, the code now relies on Django's routers to determine the actual database to use when creating
transactions. This required moving from decorators to context managers for those transactions.

An addition to Django's system checks is included to detect when the Token models are not all in the same database together. This is important because a transaction is only made for the database where AccessToken is stored.

To test the multiple database scenario, a new settings file was added which derives from settings.py
and then defines different databases and the routers needed to access them. The commit is larger than
might be expected because when there are multiple databases the Django tests have to be told which
databases to work on. Rather than copying the various test cases or making multiple database specific
ones the decision was made to add wrappers around the standard Django TestCase classes and programmatically
define the databases for them. This enables all of the same test code to work for both the one database
and the multi database scenarios with minimal maintenance costs. A tox environment that uses the
multi db settings file has been added to ensure both scenarios are always tested.

## Checklist

- [x] unit-test added
- [x] documentation updated


shaleh:
The contributing doc says I should request a reviewer but I do not see a way to do that. Trying by using an @ reference.

@jazzband/django-oauth-toolkit

oauth2_provider/oauth2_validators.py
```diff
@@ -557,8 +557,12 @@ def rotate_refresh_token(self, request):
         """
         return oauth2_settings.ROTATE_REFRESH_TOKEN
 
-    @transaction.atomic
     def save_bearer_token(self, token, request, *args, **kwargs):
+        # Use the AccessToken's database instead of making the assumption it is in 'default'.
```
dopry, line 561: Should we add a comment here that users who want to override the `save_bearer_token` behavior should override `_save_bearer_token_internals`?  Also lets call _save_bearer_token_internals to just _save_bearer_token... I find the _internals suffix confusing the _ prefer should be sufficient to indicate it is internal.

tests/common_testing.py
```diff
@@ -0,0 +1,35 @@
...
+
+class OAuth2ProviderTestCase(DjangoTestCase):
+    """Place holder to allow overriding behaviors."""
+
+
+class OAuth2ProviderTransactionTestCase(DjangoTransactionTestCase):
+    """Place holder to allow overriding behaviors."""
+
+
+if len(settings.DATABASES) > 1:
```
dopry, line 14: I'm not a fan of import side effects. Can you encapsulated the db setup in  the  OAuth2ProviderTestCase and OAuth2ProviderTransactionTestCase or in a Mixin both can inherit?

dopry (changes requested):
@shaleh this is an exceptional PR! Thank you for this contribution.  I've only got a few nitpicks, but otherwise everything looks great.  If you can address the comments inline, I'd be satisfied. I want to wait on @n2ygk for final approval.

oauth2_provider/models.py
```diff
@@ -501,16 +502,27 @@ def revoke(self):
         """
         access_token_model = get_access_token_model()
         refresh_token_model = get_refresh_token_model()
-        with transaction.atomic():
+
+        access_token_database = router.db_for_write(access_token_model)
+        refresh_token_database = router.db_for_write(refresh_token_model)
+
+        # This is highly unlikely, but let's warn people just in case it does.
```
n2ygk, line 509: Does it make sense to make it a requirement that access and refresh token databases must be the same?

oauth2_provider/models.py
```diff
@@ -501,16 +502,27 @@ def revoke(self):
...
+
+        # Use the access_token_database instead of making the assumption it is in 'default'.
+        with transaction.atomic(using=access_token_database):
             token = refresh_token_model.objects.select_for_update().filter(pk=self.pk, revoked__isnull=True)
             if not token:
                 return
             self = list(token)[0]
 
-            try:
+            with suppress(access_token_model.DoesNotExist):
```
n2ygk, line 523: I learned something new today!

n2ygk (changes requested):
See my comment about requiring a single database rather than warning about only one being in a transaction.

Also, please update the documentation in `advanced_topics` to talk about multi-database, especially the constraint that the DOT models have to be in the same database as `auth_user` due to the FK.

Once you resolve @dopry's requests I'm good to go with this.

Really nice work!

oauth2_provider/oauth2_validators.py
shaleh: I can do both. As for the name, yes, I was also uncertain. I renamed it to be very clear and catch people's attention. Looks like it worked.

tests/common_testing.py
shaleh: Let me explore that. Should be possible.

oauth2_provider/models.py
shaleh: Yes, it likely does make sense. This could be extracted into a runtime check with a clear failure along with a documentation update.

oauth2_provider/models.py
shaleh: yeah, I learned about it like maybe a year ago? It was adding in Python 3.3 or 3.4 . It has become something I recommend in PR reviews. The explicit clarity of "I don't care" is really nice.

tests/common_testing.py
shaleh: The problem is some of the tests use pytest style fixtures and decorators instead of Django TestCases. This is why `test_database_names` and `database_for_oauth2_provider` are not defined within the TestCase classes. The pytest defined test functions need this information at import time.

The cleverness employed here is what allows the same set of tests to be single and multiple database tests. It is a trade off. I could get rid of the import time checks but I would need to duplicate the test cases which seems worse.
Definitely open to alternatives or other cleverness I could use.

shaleh:
Pushed a commit which attempts to address the PR comments. As the commit message says, the tox addition is not something which can be merged. I am uncertain how to proceed there. Not include it? Mark it as one not to run requiring occasional manual checking? Move the check from ready() to some other part of the code?

n2ygk (approved):

n2ygk:
> Pushed a commit which attempts to address the PR comments. As the commit message says, the tox addition is not something which can be merged. I am uncertain how to proceed there. Not include it? Mark it as one not to run requiring occasional manual checking? Move the check from ready() to some other part of the code?

Since it raises `ImproperlyConfigured` I think your test case has to expect that exception and return a successful test if the exception happens.  You may want to make a single test just for this rather than running over the entire suite of tests. One test should be sufficient to prove that the invalid configuration exception has happened.

shaleh:
I agree with you but I was struggling with how to implement it and have been in a time crunch so I have not devoted tons of energy into this. The problem with me putting the check in `ready()` is it fails really really early. Which was my goal rather than having things look ok and then fail later when users actually do things.

n2ygk:
> I agree with you but I was struggling with how to implement it and have been in a time crunch so I have not devoted tons of energy into this. The problem with me putting the check in `ready()` is it fails really really early. Which was my goal rather than having things look ok and then fail later when users actually do things.
> 
> What I can try to do is use Django TestCase's ability to override a setting and see if I can capture this instead of using a separate settings file.

I think you can leave the check in `ready()` but remove the added `tox.ini` line and special-case test settings and instead code up a single test case that does a `@pytest.mark.oauth2_settings()` that sets the multiple databases instead of `multi_db_settings_invalid_token_configuration.py`

So, yes, what you said:-)

n2ygk:
@shaleh FYI - I'm in the midst of merging several other PRs so hold off for a bit but then you'll need to rebase and probably rename/resequence your migration(s).

n2ygk:
> @shaleh FYI - I'm in the midst of merging several other PRs so hold off for a bit but then you'll need to rebase and probably rename/resequence your migration(s).

@shaleh OK I've merged some PRs that cause merge conflicts as well as undetected (by git) conflicts with the migrations graph. Please go ahead and rebase or let me know if you want me to try.

tests/common_testing.py
dopry: I think it complicates our testing infrastructure in a way that might not be obvious to new contributors and might not play nice with other decorators on our tests in the future.  I feel like there has to be a way to achieve the same functionality without having side effects.  Even most monkey patches come in a function flavor these days so they need to be explicitly applied and don't just happen on import. 

How difficult would it be to set self.databases and self.database_for_oauth2_provider in the __init__ of OAuth2ProviderTestCase and OAuth2ProviderTestCase, then use self.* where necessary rather than doing it as a side effect? 

Would making test_database_names a settings property be an option that solves the problem and allows it to be overrode on each test? 

or export a function that returns a value rather than conditionally declaring it on import?

shaleh: I can likely fix it for the TestCase based tests. It is the pytest based ones I am struggling with.

shaleh:
No worries, I can update things.

shaleh:
After re-reading some Django docs, I realized the checks system was a better way to implement the configuration validation. It is marked as a database check so it will run on migrations but also general system checks should catch it.

shaleh:
I still need to 1) add validation for those new checks and 2) explore ways to avoid magic on import.

shaleh:
I just removed the magic. This breaks all of the pytest based testing.

Here is where I am in my understanding. Ideas welcomed.

When there are multiple databases, the Django TestCase needs to know which databases to use. It will not guess. A pytest functional test behind the scenes generates a Django TestCase in _django_db_helper in pytest-django. This is a fixture that is instantiated when any of the pytest-django 'db' fixtures are loaded which happens whenever the mark `django_db` is used. To define the databases they have to be passed into that mark. Unlike fixtures I am not aware of a way to wrap a mark in code.

What I think this means is the only way to get this working would be code that looks like this:

```
@pytest.mark.django_db(**make_database_names())
def test_mytest(....):
    # Test logic
```

which is also kinda ugly.

oauth2_provider/checks.py
```diff
@@ -0,0 +1,29 @@
+from django.apps import apps
+from django.core import checks
+from django.db import router
+
+from .settings import oauth2_settings
+
+
+@checks.register(checks.Tags.database)
+def validate_token_configuration(app_configs, **kwargs):
+    breakpoint()
```
9128305, line 10: It should be removed

shaleh: yep, code is not 100% ready for merge. Thanks for checking this.

shaleh:
There is a flake8 package that checks for debug lines. But ruff does not support external lint packages. All lints have to be ported into ruff.

9128305:
Ruff also has similar rules https://docs.astral.sh/ruff/rules/debugger/#debugger-t100

dopry:
> What I think this means is the only way to get this working would be code that looks like this:
> 
> ```
> @pytest.mark.django_db(**make_database_names())
> def test_mytest(....):
>     # Test logic
> ```
> 
> which is also kinda ugly.

This is kinda what I expected... alternatively you could be explicit and not use a function. I strongly prefer not being too clever in testing. In my experience when I make my tests too clever I spend more time debugging my test harnesses than writing production code.

Personally I'd lean toward something along the lines of, 
```
@override_settings(DATABASE=...)
@pytest.mark.django_db(...)
class MyMultiDBTestCase(...):
   ...
```
but mind you it's coming from a more general frame of mind not specific to DOT's existing test setup so might not work for some reason I'm not aware of at the moment..

I should have some free time later in the afternoon to look at this in more detail.

shaleh:
ok, django_db mark now uses the retrieve_current_databases function.

What remains is:
- test and validate the new checks work
- assuming they do, update the documentation to mention the checks are available and what their failure means.

shaleh:
@dopry this issue is not the definition of the DATABASES themselves. The issue is in Django's own TestCase. It has to have a definition of the databases allowed in testing defined on the class itself. I made the change in setUpClass because Django does database work at the class level before the per instance `__init__` is ever called.

When a functional test from pytest uses pytest-django's `django_db` this internally generates a one off, local TestCase class and loads it essentially like a context manager. https://github.com/pytest-dev/pytest-django/blob/5598264ce8b2b333632efd4f06b7d740276953de/pytest_django/fixtures.py#L212
I did not see a good way to interfere with this. Open to ideas. Instead I added the function call to the marks.

shaleh:
ok, I think at this point I have resolved open concerns. Let me know if there is anything I can improve. Thank you for the interest and attention. It is super appreciated.

docs/contributing.rst
```diff
@@ -252,6 +252,26 @@ Open :file:`mycoverage/index.html` in your browser and you can see a coverage su
 
 There's no need to wait for Codecov to complain after you submit your PR.
 
+The tests are generic and written to work with both single database and multiple database configurations. tox will run
+tests both ways. You can see the configurations used in tests/settings.py and tests/multi_db_settins.py.
```
n2ygk, line 256:
```suggestion
tests both ways. You can see the configurations used in tests/settings.py and tests/multi_db_settings.py.
```
minor typo

n2ygk (changes requested):

n2ygk (approved):

n2ygk:
@dopry Please provide final approval on this.

@shaleh Sorry for the delay in reviewing; I was on vacation last week.

shaleh:
All good. These things move at their own pace. Thank you for the typo fixes.

n2ygk:
@dopry sorry to be a pain but are you able to do a final review of this. If not I'll just go ahead and merged it. I'm pretty sure it's all good at this point.

dopry:
No worries. I appreciate the reminder, been distracted by life

dopry:
@shaleh Thank you for all your work on this and addressing my issues. I don't have the time for functional tests right now, but all my concerns with code style and developer accessibility are addressed! I got into the DB stuff with it after my last comments and I see why I never used it. And again, Thank you so much for your hard work on this!

dopry (approved):

dopry:
to future maintainers, I'm reverting the tests which use `retrieve_current_databases()` with default filtered out. In retrospect, Django has too many assumptions about it being set to effectively run without. I think asserting requests are hitting the correct db are sufficient. also it is causing major headaches for primary/replica support testing across the suite. If we want to recreate an adversarial test like that it should be more tightly isolated. see: https://github.com/django-oauth/django-oauth-toolkit/pull/1709 for the primary/replica testing where that bit is being reverted. I think we should guide users toward using default in a way that plays nice with django and not go out of our way to support non-standard configurations.  The db router support and tests verifying it are remaining in place.

PR #1452, opened 2024-08-09
Refresh Token Reuse Protection

### LINKED ISSUE #1404: Refresh token reuse detection
I want to increase security by detecting reuse of old refresh tokens. Scenario:

- A malicious actor steals the refresh token of a user
- The malicious actor uses the refresh token to get a new access & refresh token
- The user tries to use the refresh token to get a new access & refresh token
- The user's request is denied because they used an outdated refresh token
- **The malicious actor is still logged in**

In case an old refresh token is used (taking `REFRESH_TOKEN_GRACE_PERIOD_SECONDS` into consideration), not only should it be rejected, but also the whole token family should be invalidated.

The implementation should be according to the RFC: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-25#section-4.14.2-5.2.2
> **Refresh token rotation:** the authorization server issues a new refresh token with every access token refresh response. The previous refresh token is invalidated but information about the relationship is retained by the authorization server. If a refresh token is compromised and subsequently used by both the attacker and the legitimate client, one of them will present an invalidated refresh token, which will inform the authorization server of the breach. The authorization server cannot determine which party submitted the invalid refresh token, but it will revoke the active refresh token. This stops the attack at the cost of forcing the legitimate client to obtain a fresh authorization grant.
>
> Implementation note: The grant to which a refresh token belongs may be encoded into the refresh token itself. This can enable an authorization server to efficiently determine the grant to which a refresh token belongs, and by extension, all refresh tokens that need to be revoked. Authorization servers MUST ensure the integrity of the refresh token value in this case, for example, using signatures.

**Additional context**
- Description and diagram from auth0 explains the rational pretty good: https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation#automatic-reuse-detection

soerface:
Fixes #1404

## Description of the Change

This PR adds a new setting / feature `REFRESH_TOKEN_REUSE_PROTECTION`. It allows django-oauth-toolkit to revoke related refresh and access tokens whenever an already revoked refresh token is used again. This indicates a breach - for details see #1404.

The implementation does it by adding a new column `token_family` to the refresh token table. When a new refresh token is created with an existing refresh token, the new refresh token get's assigned the same `token_family` value as the original token. This ensures that all tokens from the same session share the same family. Otherwise (first token of a session, gathered by e.g. username & password), the `token_family` is a random value.

Whenever a revoked refresh token is presented, the server will query for all tokens with the same `token_family` value, and call the `.revoke()` method on it.

To discuss: The "select all tokens of a family and revoke" code is always being run when a revoked token is presented. Maybe this could provide a vector for a denial of service attack. Therefore, I was wondering if we should set the `token_family` value to NULL after the whole family has been revoked. Another thought to ease database load was furthermore to use `objects.filter(...).delete()`, however I was afraid that the logic inside the `.revoke()` method was relevant and shouldn't be skipped lightheartedly. Furthermore, do you think an index on that new column would be required?

## Checklist

- [x] unit-test added
- [x] documentation updated


tests/migrations/0006_basetestapplication_token_family.py
soerface, line 1: I think I did something wrong with the migration. According to the guidelines, I was supposed to run `django-admin makemigrations --settings tests.mig_settings`, which didn't work at all because of some missing OIDC settings. I've instead run `django-admin makemigrations --settings tests.settings`, which does work, but also added a lot of unrelated migrations in here. Should this file be touched at all? Should I remove the `AlterField` operations for the unrelated changes?

oauth2_provider/oauth2_validators.py
```diff
@@ -752,22 +761,25 @@ def validate_refresh_token(self, refresh_token, client, request, *args, **kwargs
...
 
-        null_or_recent = Q(revoked__isnull=True) | Q(
-            revoked__gt=timezone.now() - timedelta(seconds=oauth2_settings.REFRESH_TOKEN_GRACE_PERIOD_SECONDS)
-        )
-        rt = (
-            RefreshToken.objects.filter(null_or_recent, token=refresh_token)
-            .select_related("access_token")
-            .first()
-        )
+        rt = RefreshToken.objects.filter(token=refresh_token).select_related("access_token").first()
```
soerface, line 764: I needed to remove the filter, so that I'm able to get the refresh token and it's family for the revocation process. Therefore, those checks live further below.

oauth2_provider/settings.py
```diff
@@ -54,6 +54,7 @@
     "ID_TOKEN_EXPIRE_SECONDS": 36000,
     "REFRESH_TOKEN_EXPIRE_SECONDS": None,
     "REFRESH_TOKEN_GRACE_PERIOD_SECONDS": 0,
+    "REFRESH_TOKEN_REUSE_PROTECTION": False,
```
soerface, line 57: I've set the default to False for compatibility reasons (if someone has a faulty client that due to a bug reuses the token, it might cause accidental logouts), but at the same time you could also argue it is better to have the most secure option as default.

oauth2_provider/migrations/0011_refreshtoken_token_family.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 5.2 on 2024-08-09 16:40
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('oauth2_provider', '0010_application_allowed_origins'),
```
n2ygk, line 9:
```suggestion
        ('oauth2_provider', '0010_application_allowed_origins'),
        migrations.swappable_dependency(settings.REFRESH_TOKEN_MODEL)
```
I think this needs to get added. Same question I have outstanding with @dopry on another pending PR.

Because of swappable models I believe the automatically generated migrations have to have this dependency manually added.

tests/migrations/0006_basetestapplication_token_family.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 5.2 on 2024-08-09 16:40
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tests', '0005_basetestapplication_allowed_origins_and_more'),
```
n2ygk, line 9: needs `migrations.swappable_dependency`?

n2ygk (changes requested):
@soerface This is a really nice security improvement by following the newer BCPs.  I agree that the default for now has to be False to avoid immediate breakage for existing users. We can perhaps raise a DeprecationWarning and eventually make the default be True for a future major release.

I have some questions about the migrations and whether they need `migrations.swappable_dependency` added to them. I'm stumped by this swappable stuff and you can see there are a number of issues raised over the years by people trying to use the functionality....

oauth2_provider/migrations/0011_refreshtoken_token_family.py
soerface: `0001_inital.py` is using `oauth2_settings.REFRESH_TOKEN_MODEL` for a relation, so `from oauth2_provider.settings import oauth2_settings` and then using `oauth2_settings.REFRESH_TOKEN_MODEL` here?

tests/migrations/0006_basetestapplication_token_family.py
soerface: Added it. But I'm not quite sure if it is needed. In contrast to `0001_initial.py`, the migration itself is not directly referring to the model, just to the model name with a string. In `0001_initial.py`, the setting `oauth2_settings.REFRESH_TOKEN_MODEL` and `settings.AUTH_USER_MODEL` are used directly (although only for AUTH_USER_MODEL the dependency is declared, but maybe it's a mistake from the past that the REFRESH_TOKEN_MODEL wasn't declared?).

So I'm wondering what the option would actually do.

tests/migrations/0006_basetestapplication_token_family.py
n2ygk: hmm, we'll need to fix that... #1453

tests/migrations/0006_basetestapplication_token_family.py
n2ygk: I asked a possible liar (ChatGPT): 

> Yes, when working with swappable models in Django, you should add migrations.swappable_dependency() to the auto-generated migration code. This ensures that the migration system correctly handles dependencies between the swappable models and other models.
> 
> What is migrations.swappable_dependency()?
> The migrations.swappable_dependency() function is used to specify a dependency on a swappable model. This ensures that migrations are applied in the correct order, especially when custom models are used to replace the default ones.

oauth2_provider/migrations/0011_refreshtoken_token_family.py
n2ygk: Let's leave it as is for now and then think about a separate PR to understand and update all the migrations to deal with swappable models appropriately. The main issue I've seen is when someone has been using DOT for a while already and then tries to switch to a swappable model after having already populated a "standard" model. When I've done it, I've tended to start over from scratch with an empty database and manually deal with adding applications, etc.

n2ygk (approved):

PR #1461, opened 2024-08-15
modernize packaging using pyproject.toml

dulmandakh:
## Description of the Change

This PR modernizes packaging using pyproject.toml instead of setup.py. See https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

You can test the change using below command
```sh
tox -e build
```

## Checklist

- [ ] unit-test added
- [ ] documentation updated


n2ygk:
We need to think hard about this since the package builds and twine uploads to pypi are somehow automated by jazzband magic.

dulmandakh:
```python -m build``` output artifact is same with ```python setup.py sdist --format=gztar bdist_wheel```, thus `twine check` works as expected. IMHO, there won't be any change in package publishing, please see https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/.

PR #1469, opened 2024-08-27
Work around double parsing of ui_locales

### LINKED ISSUE #1468: `ui_locales` request parameter triggers `AttributeError` under certain circumstances
**Describe the bug**

A valid authorization request, for a client that doesn't require consent, that includes the `ui_locales` parameter triggers `AttributeError: 'list' object has no attribute 'split'`.

**To Reproduce**

* Configure DOT, with OIDC enabled.
* Create an Application and set `skip_authorization`
* While logged in trigger an authorization request that includes the `scope` and `ui_locales` parameter, i.e.:
  `/o/authorize/?response_type=code&client_id=test&scope=openid&ui_locales=de`

**Expected behavior**

I expect the authorization request to succeed and the user to be redirected to the redirect_uri with a code response.

**Version**

django==5.1
django-oauth-toolkit==2.4.0
oauthlib==3.2.2

**Additional context**

I've written a testcase that triggers the issue, `TestUILocalesParam.test_trusted_application_ui_locales_param` is the one that fails:

```python
from django.contrib.auth import get_user_model
from django.test import TestCase, override_settings
from django.urls import reverse

from oauth2_provider.models import get_application_model

UserModel = get_user_model()


@override_settings(OAUTH2_PROVIDER={
    "OIDC_ENABLED": True,
    "PKCE_REQUIRED": False,
    "SCOPES": {
        "openid": "OpenID connect",
    },
})
class TestUILocalesParam(TestCase):
    @classmethod
    def setUpTestData(cls):
        cls.application = Application.objects.create(
            name="Test Application",
            client_id="test",
            redirect_uris="https://www.example.com",
            client_type=Application.CLIENT_PUBLIC,
            authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
        )
        cls.trusted_application = Application.objects.create(
            name="Trusted Application",
            client_id="trusted",
            redirect_uris="https://www.example.com",
            client_type=Application.CLIENT_PUBLIC,
            authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
            skip_authorization=True,
        )
        cls.user = UserModel.objects.create_user("test_user")
        cls.url = reverse("oauth2_provider:authorize")

    def setUp(self):
        self.client.force_login(self.user)

    def test_application_ui_locales_param(self):
        response = self.client.get(
            f"{self.url}?response_type=code&client_id=test&scope=openid&ui_locales=de",
        )
        self.assertEqual(response.status_code, 200)

    def test_trusted_application_ui_locales_param(self):
        response = self.client.get(
            f"{self.url}?response_type=code&client_id=trusted&scope=openid&ui_locales=de",
        )
        self.assertEqual(response.status_code, 302)
```

My guess is that the call to `self.create_authorization_response` in https://github.com/jazzband/django-oauth-toolkit/blob/9fceef11c59a200711d1e7023495131e42dfae0e/oauth2_provider/views/base.py#L194-L198 triggers this issue because it tries to parse the request parameters for the second time, the first time is here: 
https://github.com/jazzband/django-oauth-toolkit/blob/9fceef11c59a200711d1e7023495131e42dfae0e/oauth2_provider/views/base.py#L149

jaap3:
Fixes #1468

## Description of the Change

Makes sure `credential["ui_locales"] is a string before passing it to `oauthlib`.

`oauthlib` always calls `split` on the values of `ui_locales`:

https://github.com/oauthlib/oauthlib/blob/d319c54ae0342d9a2596ef7afa1e17984c303550/oauthlib/openid/connect/core/grant_types/base.py#L317

The way django-oauth-toolkit is structured the `credentials` dictionary contains the value of `ui_locales` as processed by `oauthlib`, which makes it a list. It is then passed back into `oauthlib` which tries to parse it again, resulting in an `AttributeError`.

`oauthlib` is a bit more careful with the way it handles the `prompt` parameter:

https://github.com/oauthlib/oauthlib/blob/d319c54ae0342d9a2596ef7afa1e17984c303550/oauthlib/openid/connect/core/grant_types/base.py#L289-L292

So this might be a bug in the way `oauthlib` handles the incoming `ui_locales` parameter.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #1443 Query strings with invalid hex values now raise a SuspiciousOperation exception (in DRF extension) instead of raising a 500 ValueError: Invalid hex encoding in query string.
+* #1468 `ui_locales` request parameter triggers `AttributeError` under certain circumstances
 ### Security
 
 ## [2.4.0] - 2024-05-13
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -186,6 +186,10 @@ class AuthorizationView(BaseAuthorizationView, FormView):
         # a successful response depending on "approval_prompt" url parameter
         require_approval = request.GET.get("approval_prompt", oauth2_settings.REQUEST_APPROVAL_PROMPT)
 
+        if "ui_locales" in credentials and isinstance(credentials["ui_locales"], list):
+            # Make sure ui_locales a space separated string for oauthlib to handle it correctly.
+            credentials["ui_locales"] = " ".join(credentials["ui_locales"])
+
         try:
             # If skip_authorization field is True, skip the authorization screen even
             # if this is the first use of the application and there was no previous authorization.
--- a/tests/test_ui_locales.py
+++ b/tests/test_ui_locales.py
@@ -4,17 +4,20 @@ from django.urls import reverse
 
 from oauth2_provider.models import get_application_model
 
+
 UserModel = get_user_model()
 Application = get_application_model()
 
 
-@override_settings(OAUTH2_PROVIDER={
-    "OIDC_ENABLED": True,
-    "PKCE_REQUIRED": False,
-    "SCOPES": {
-        "openid": "OpenID connect",
-    },
-})
+@override_settings(
+    OAUTH2_PROVIDER={
+        "OIDC_ENABLED": True,
+        "PKCE_REQUIRED": False,
+        "SCOPES": {
+            "openid": "OpenID connect",
+        },
+    }
+)
 class TestUILocalesParam(TestCase):
     @classmethod
     def setUpTestData(cls):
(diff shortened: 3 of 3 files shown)


jaap3:
I didn't really know where to put the tests, so it is in it's own module.

n2ygk (commented):
I'm wondering if it would be more appropriate to push this upstream to oauthlib?

jaap3:
As mentioned in the upstream PR, it's unclear whether this issue stems from a bug in `oauthlib` or a potential misuse of the `oauthlib` API by `django-oauth-toolkit` (DOT). The problem occurs when DOT creates the `credentials` dictionary from what `oauthlib` refers to as `request_info`, and then passes that back into `oauthlib`.

The multiple layers in DOT and `oauthlib` make the process quite confusing and challenging to follow, not to mention debug. Therefore, I opted to work around the issue as close to the point of failure as possible.

I’m waiting for feedback on the `oauthlib` side. In the meantime, we wrote a middleware that removes `ui_locales` from the querystring using a redirect. It's not ideal, but at least OIDC works.

n2ygk (approved):
Let's go ahead and merge this fix here and keep an eye on whether oauthlib implements the fix there in the future at which point this change can potentially be reverted.
Thanks for this!

jaap3:
Thanks!

PR #1476, opened 2024-09-03
Raise InvalidGrantError if no associated grant exists when invalidating an authorization code

mforner13:
## Description of the Change
This change raises an `InvalidGrantError` from the `OAuth2Validator.invalidate_authorization_code` method if the Grant object intended to be deleted does not exist.

Currently, when invalidating an authorization code after it has been used, if for whatever reason the associated grant object no longer exists, an uncaught `Grant.DoesNotExist` exception is raised. This leads to 500 responses being returned to clients. This could, for example, be caused by concurrent requests being made using the same authorization code.

The change in this PR handles this scenario gracefully by catching `Grant.DoesNotExist` and raising an `InvalidGrantError` which will return a 400 'invalid_grant' response to the client.

## Checklist

- [x] unit-test added
- [ ] documentation updated (No existing documentation for this validator class)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -119,3 +119,4 @@ pySilver
 Wouter Klein Heerenbrink
 Yaroslav Halchenko
 Yuri Savin
+Miriam Forner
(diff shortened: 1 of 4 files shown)


n2ygk (approved):
Hey, thanks for this! Nice improvement.

PR #1486, opened 2024-09-05
Fix test for changed error message from newer Django (djmain)

### LINKED ISSUE #1482: py{310,311,312}-djmain tests.test_commands.CreateApplicationTest failing
**Describe the bug**

djmain has changed the wording of a user not found error message, causing the test for a failed validation to fail:

```
______________________________________________ CreateApplicationTest.test_validation_failed_message _______________________________________________

self = <tests.test_commands.CreateApplicationTest testMethod=test_validation_failed_message>

    def test_validation_failed_message(self):
        output = StringIO()
        call_command(
            "createapplication",
            "confidential",
            "authorization-code",
            "--redirect-uris=http://example.com http://example2.com",
            "--user=783",
            stdout=output,
        )
    
        self.assertIn("user", output.getvalue())
        self.assertIn("783", output.getvalue())
>       self.assertIn("does not exist", output.getvalue())
E       AssertionError: 'does not exist' not found in "\x1b[31;1mPlease correct the following errors:\n - user: ['user instance with id 783 is not a valid choice.']\x1b[0m\n"

tests/test_commands.py:145: AssertionError
```

**To Reproduce**

```
tox -e py310-djmain
```

**Expected behavior**

All tests should have passed

**Version**

**Additional context**

n2ygk:
Fixes #1482

## Description of the Change
Older Django has one error message that the failure test was looking for. Newer Django (main branch) has a different error message, so check for both.
 
## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -145,6 +145,5 @@ class CreateApplicationTest(TestCase):
         self.assertIn("783", output_str)
         # newer Django (>5.1) changes the error message from "does not exist" to "is not a valid choice"
         self.assertTrue(
-            any(substring in output_str for substring in ["does not exist", "is not a valid choice"]),
-            f"Output did not contain 'does not exist' or 'is not a valid choice'. Actual output: {output_str}",
+            any(substring in output_str for substring in ["does not exist", "is not a valid choice"])
         )
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -140,6 +140,11 @@ class CreateApplicationTest(TestCase):
             stdout=output,
         )
 
-        self.assertIn("user", output.getvalue())
-        self.assertIn("783", output.getvalue())
-        self.assertIn("does not exist", output.getvalue())
+        output_str = output.getvalue()
+        self.assertIn("user", output_str)
+        self.assertIn("783", output_str)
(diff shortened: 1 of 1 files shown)


tests/test_commands.py
```diff
@@ -140,6 +140,10 @@ def test_validation_failed_message(self):
...
         )
 
-        self.assertIn("user", output.getvalue())
-        self.assertIn("783", output.getvalue())
-        self.assertIn("does not exist", output.getvalue())
+        output_str = output.getvalue()
+        self.assertIn("user", output_str)
+        self.assertIn("783", output_str)
+        # newer Django (>5.1) changes the error message from "does not exist" to "is not a valid choice"
+        self.assertTrue(
```
dopry, line 147: would something like the following be more self documenting?

```python
import django 

if django.VERSION <= (5,1):
   self.assertIn("does not exist", output.getvalue())
else:
   self.assertIn("is not a valid choice", output.getvalue())
```

n2ygk: I had to make it a test for `< (5,2)` because `(5,1)` is not `<= (5, 1, 1, 'final', 0)` .  I suppose the alternative would be to take the first 2 elements of the tuple, but I think this is readable enough....

dopry (approved):

PR #1491, opened 2024-09-06
add a forward migration function that computes checksum for existing tokens

### LINKED ISSUE #1490: 3.0.0 access token migration does not work
**Describe the bug**

The AccessToken has a new checksum field, that is both `unique` and `blank`, which means that existing tokens will be assigned empty checksum by default. However, empty string is not unique.

```
django.db.utils.IntegrityError: could not create unique index  oauth2_provider_accesstoken_token_checksum_key"
DETAIL:  Key (token_checksum)=() is duplicated.          
```

**To Reproduce**
Run migrate on an existing database.

**Expected behavior**

The migration should run successfully.

**Version**

3.0.0

**Additional context**

n2ygk:
Fixes #1490

## Description of the Change
Add a forward migration that causes a `token_checksum` to be added to existing migrated tokens.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/migrations/0012_add_token_checksum.py
+++ b/oauth2_provider/migrations/0012_add_token_checksum.py
@@ -4,6 +4,16 @@ import oauth2_provider.models
 from django.db import migrations, models
 from oauth2_provider.settings import oauth2_settings
 
+def forwards_func(apps, schema_editor):
+    """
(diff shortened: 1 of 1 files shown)


dopry (approved):
I didn't so any functional testing, but it looks good from a code review perspective.

n2ygk:
I can't think of a good way to test migrations for cases like this.

PR #1498, opened 2024-09-16
Support for specifying client secret hasher

matejsp:
With every Django version PBKDF2 password hasher gets iterations increased and the next one will be set to 1 mio.
This poses a problem in our case because it affects the performance of /token endpoint. We whitelist to the IPs from which endpoint is accessible so we want to use weaker algorithm. For normal UI login we still want to retain default one.

## Description of the Change
Added a setting for specifying client server hasher using CLIENT_SECRET_HASHER.

## Checklist

- [X] unit-test added
- [X] documentation updated

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -89,6 +89,8 @@ INSTALLED_APPS = (
     "tests",
 )
 
+PASSWORD_HASHERS = django.conf.settings.PASSWORD_HASHERS + ["tests.custom_hasher.MyPBKDF2PasswordHasher"]
+
 LOGGING = {
     "version": 1,
     "disable_existing_loggers": False,
--- a/AUTHORS
+++ b/AUTHORS
@@ -86,6 +86,7 @@ Ludwig Hähne
 Łukasz Skarżyński
 Madison Swain-Bowden
 Marcus Sonestedt
+Matej Spiller Muys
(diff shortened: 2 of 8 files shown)


tests/settings.py
```diff
@@ -89,6 +89,15 @@
...
 
+PASSWORD_HASHERS = [
+    "django.contrib.auth.hashers.PBKDF2PasswordHasher",
+    "django.contrib.auth.hashers.PBKDF2SHA1PasswordHasher",
+    "django.contrib.auth.hashers.Argon2PasswordHasher",
+    "django.contrib.auth.hashers.BCryptSHA256PasswordHasher",
+    "django.contrib.auth.hashers.ScryptPasswordHasher",
+    "tests.custom_hasher.MyPBKDF2PasswordHasher",
+]
+
```
n2ygk, line 100: How about something like this rather than repeating the list of "standard" hashers:
```suggestion
+PASSWORD_HASHERS = django.conf.settings.PASSWORD_HASHERS + ["tests.custom_hasher.MyPBKDF2PasswordHasher"]
```

n2ygk (changes requested):

n2ygk (approved):
Thanks for this. Nice job documenting and adding test cases!

PR #1499, opened 2024-09-16
deal with 404 or 405 validator error

n2ygk:
Fixes #1497 unrelated tox failures

## Description of the Change
The test mocked validation endpoint sometimes returns a 404 or a 405. It's not clear where this changed
but it first appeared in #1497 which was unrelated.

It turns out the test neglected to mock the POST endpoint so the POST was being sent to the live example.com website. Apparently example.com changed their response from a 404 to a 405.

The fix mocks the endpoint to always return a 404 Not Found.
 
## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/tests/test_oauth2_validators.py
+++ b/tests/test_oauth2_validators.py
@@ -3,6 +3,7 @@ import datetime
 import json
 
 import pytest
+import requests
 from django.contrib.auth import get_user_model
 from django.contrib.auth.hashers import make_password
 from django.utils import timezone
@@ -505,16 +506,22 @@ class TestOAuth2ValidatorErrorResourceToken(TestCase):
         """
         Ensure we log the error when the authentication server returns a non-200 response.
         """
-        with self.assertLogs(logger="oauth2_provider") as mock_log:
-            self.validator._get_token_from_authentication_server(
-                self.token, self.introspection_url, self.introspection_token, None
-            )
-            self.assertTrue(
-                any(
-                    "Failed to get a valid response from authentication server" in message
-                    for message in mock_log.output
+        mock_response = requests.Response()
+        mock_response.status_code = 404
+        mock_response.reason = "Not Found"
+        with mock.patch("requests.post") as mock_post:
+            mock_post.return_value = mock_response
+            with self.assertLogs(logger="oauth2_provider") as mock_log:
+                self.validator._get_token_from_authentication_server(
+                    self.token, self.introspection_url, self.introspection_token, None
                 )
-            )
+                self.assertIn(
+                    "ERROR:oauth2_provider:Introspection: Failed to "
+                    "get a valid response from authentication server. "
+                    "Status code: 404, Reason: "
+                    "Not Found.\nNoneType: None",
+                    mock_log.output,
+                )
 
 
 @pytest.mark.oauth2_settings(presets.OIDC_SETTINGS_RW)
--- a/tests/test_oauth2_validators.py
+++ b/tests/test_oauth2_validators.py
@@ -501,9 +501,9 @@ class TestOAuth2ValidatorErrorResourceToken(TestCase):
         cls.introspection_token = "test_introspection_token"
         cls.validator = OAuth2Validator()
 
-    def test_response_when_auth_server_response_return_404_405(self):
+    def test_response_when_auth_server_response_not_200(self):
         """
-        Deal with either 404 or 405 response
+        Ensure we log the error when the authentication server returns a non-200 response.
         """
         with self.assertLogs(logger="oauth2_provider") as mock_log:
             self.validator._get_token_from_authentication_server(
--- a/tests/test_oauth2_validators.py
+++ b/tests/test_oauth2_validators.py
@@ -501,17 +501,19 @@ class TestOAuth2ValidatorErrorResourceToken(TestCase):
         cls.introspection_token = "test_introspection_token"
         cls.validator = OAuth2Validator()
 
-    def test_response_when_auth_server_response_return_404(self):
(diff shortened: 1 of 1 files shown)


dopry:
I would be more confident this is an appropriate fix if we knew why we were getting the 405.  If this varies with Django Version it would like to see a check against the django version to determine the correct assertion, or a comment explaining why we get differing response. Is this solely the result of the django version or is pip resolving a different version of OAuthLib for different django versions?

n2ygk:
> I would be more confident this is an appropriate fix if we knew why we were getting the 405. If this varies with Django Version it would like to see a check against the django version to determine the correct assertion, or a comment explaining why we get differing response. Is this solely the result of the django version or is pip resolving a different version of OAuthLib for different django versions?

I've been trying to find the source. It "suddenly" broke for `< djmain` which is unexpected.

dopry:
Maybe this filtered in from a dependency. Can the stack trace lead us back to the raise to figure out which module the 405 is originating in?

n2ygk:
> Maybe this filtered in from a dependency. Can the stack trace lead us back to the raise to figure out which module the 405 is originating in?

I've been trying to find it. It's a response from a mocked HTTP server so the stack trace ends at the request. I tried drilling down for quite a while and was unable to find where it's coming from. Must be in Django somewhere since it varies based on django version

dopry:
Where and how is the mocked http server setup? Maybe we can get into it? I looked into OAuthLib no new releases there. Doesn't look like anything changed in Django so maybe the mocking library changed it's default response.

n2ygk:
@dopry I'll do some more digging...

n2ygk:
I'm not sure where something changed, but now I am unable to reproduce. All flavors of py*-dj* are consistently returning a 405.

n2ygk:
@dopry I think I found it. https://www.python.org/downloads/release/python-31015/ was released 9/7/24. However the Python 10 I have locally is Python 3.10.13.  Mock is part of cpython. Investigating further....

dopry:
I wonder  if it could have something to do with https://github.com/jazzband/django-oauth-toolkit/blob/90d7300c06be80e565c6557f0b8f260968748634/oauth2_provider/views/introspect.py#L49 not explicitly setting a status?

n2ygk:
> I wonder if it could have something to do with
> 
> https://github.com/jazzband/django-oauth-toolkit/blob/90d7300c06be80e565c6557f0b8f260968748634/oauth2_provider/views/introspect.py#L49
> 
> not explicitly setting a status?

I tried to set a breakpoint there and the function is never reached. I think this is happening way earlier as the 405 is a method not allowed (POST I assume). I will try digging further but honestly I don't really understand what the test case is supposed to be testing for.

dopry:
I've been poking a but more and getting a 405 Method Not Allowed is a pretty big change from getting a 404 Not Found.  For a hot second I thought it was a CORS related issue, but I sent an option request in `_get_token_from_authentication_server` and it returned ```
'Allow': 'OPTIONS, GET, HEAD, POST', 
```
so that looks right, but the 405 response on the POST request doesn't return the `Allow` header which it should... 

so I'm starting to share the suspicion that the mock server is misbehaving. I still don't quite understand how the mock server get stood up. I'm also not having any luck stepping into the requests.post with my debugger to figure our how the mocked server is working..  I'm used to the server being mocked the the TestCase and using self.client to call the django instance under test... I'm not quite sure how that works in pytest. I get the sense this mocked server may be setup by pytest...  and maybe it's mocking has changes somehow... or a library it depends on has changed.

**Comment by n2ygk** [comment]
Yeah I wonder if this test case was always just wrong because there is no mock server. If you look at other test cases they use MagicMock(wraps=Request) (which is oauthlib's Request).
https://github.com/jazzband/django-oauth-toolkit/blob/90d7300c06be80e565c6557f0b8f260968748634/tests/test_oauth2_validators.py#L59-L65

This might have been an attempt to show what happens when a bad url is used or something. I don't really understand the logic since a 404 is also not useful. It's not testing if a bad Bearer token was provided....

**Comment by dopry** [comment]
looking at the test itself it's just ensuring we're logging an error when we get any response other than 200...  from that perspective your approach to resolving the issue seems 'Good Enough'. While I'm still concerned about what changed and what it might mean for our test infrastructure, I'm gonna go ahead an approve this.

**Comment by n2ygk** [comment]
HAHA! found it:
```
django-oauth-toolkit$ curl  -v -X POST http://example.com/token/introspection
* Host example.com:80 was resolved.
* IPv6: (none)
* IPv4: 93.184.215.14
*   Trying 93.184.215.14:80...
* Connected to example.com (93.184.215.14) port 80
> POST /token/introspection HTTP/1.1
> Host: example.com
> User-Agent: curl/8.7.1
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 405 Method Not Allowed
< Content-Type: text/html; charset=UTF-8
< Date: Thu, 19 Sep 2024 18:51:30 GMT
< Server: ECAcc (nyd/D12F)
< Content-Length: 0
< 
* Connection #0 to host example.com left intact
* 
```

Example.com changed their web server to return a 405. I guess. Still not sure why 404 sometimes. Maybe to do with headers?

Saw these headers in the debugger for the response:
```
{'Content-Type': 'text/html; charset=UTF-8', 'Date': 'Thu, 19 Sep 2024 18:46:54 GMT', 'Server': 'ECAcc (nyd/D136)', 'Content-Length': '0'}
```

n2ygk:
This test is just incorrectly connecting to example.com. We need to fix that.

dopry:
oh that sounds like we need to mock the request...  Are you up to doing the fix?

dopry (approved):

PR #1502, opened 2024-09-21
Fixed Handled error in OAuth2ExtraTokenMiddleware when authheader has `Bearer` with no token-string following up

### LINKED ISSUE #1496: Empty Bearer token results in unhandled exception
**Describe the bug**
If you do not pass the "token" string after Bearer in `Authorization` header, it will result in an unhandled exception.

**To Reproduce**
Use `django-oauth-toolkit` and setup as a Django middleware.
Define any route and register in `urls.py`

**Expected behavior**

**Version**
3.0.1

Tested with 

**Additional context**
My request header looks like this:
image

I think problem is in [here](https://github.com/jazzband/django-oauth-toolkit/blob/90d7300c06be80e565c6557f0b8f260968748634/oauth2_provider/middleware.py#L56)
It should check if there is space before splitting `authheader` by space.

Wondering, if you are expecting a PR.

Tuhin-thinks:
Fixed the crash in application while using OAuth2ExtraTokenMiddleware. When Bearer token passed is empty.
`Authorization: Bearer` would result in this crash.

Fixes #1496

## Checklist

- [ ] unit-test added
- [ ] documentation updated


oauth2_provider/middleware.py
```diff
@@ -52,8 +52,8 @@ def __init__(self, get_response):
 
     def __call__(self, request):
         authheader = request.META.get("HTTP_AUTHORIZATION", "")
-        if authheader.startswith("Bearer"):
-            tokenstring = authheader.split()[1]
+        if authheader.startswith("Bearer") and len(authheader.split(maxsplit=1)) == 2:
+            tokenstring = authheader.split(maxsplit=1)[1]
```
n2ygk, line 56: I suggest not calling split() twice:
```suggestion
        splits = autheader.split(maxsplit=1)
        if authheader.startswith("Bearer") and len(splits) == 2:
            tokenstring = splits[1]
```

n2ygk (changes requested):
Thanks for this fix. Please also add a test case in `tests/test_auth_backends.py` to assure full code coverage.

oauth2_provider/middleware.py
Tuhin-thinks: Thank you for your suggestion, I will make the changes by today!

dopry:
@Tuhin-thinks  this looks good. I think the last task to get this merge ready is the test.

Tuhin-thinks:
@dopry Thanks, I am getting them ready.

PR #1508, opened 2024-09-28
feat: allowed_origin and redirect_uri wildcards

### LINKED ISSUE #1506: feat: Wildcard support for Allowed Origins and Redirect URIs
## Use cases
As an app developer I want to be able to login from netlify, vercel, or cloudflare feature branch preview deployments without needing to specify every domain variation in the origins and redirect_uris.

## Current approaches

The wildcard support can be done by implementing a custom application class that overrides the redirect_uri_allowed, origin_allowed, and post_logout_redirect_uri_allowed methods.  I feel these use cases are common enough that we should make a better effort to support them out of the box.

## Security considerations 

https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2 states "The redirection endpoint URI MUST be an absolute URI as defined by [[RFC3986] Section 4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3)."  This is based on https://datatracker.ietf.org/doc/html/rfc6749#section-10.6. 10.6 focuses on who controls the URI. 

In my opinion the specification takes 10.6's concerns about control over a URI to an extreme with the absolute-URI requirement.  I feel well informed users can craft wildcard domain specifications that will only allow resources under their control. I feel that by ensuring wildcards are only allowed at subdomains and deeper we satisfy the spirit of 10.6 if not the letter of 3.1.2.

## Proposed implementation

- allow a * prefix domain on the host portion starting at the third level domain

   valid: https://\*.example.com/path
   valid: https://\*.netlify.example.com/path
   valid: https://*-project.vercel.app
   invalid: https://\*ample.com/path
  
@n2ygk I'd love your thoughts on this

dopry:
Fixes #1506 

## Description of the Change

Implement Wildcard support in allowed origins and redirect uris

## Checklist

- [x] unit-test added
- [X] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-<!--
+
 ## [unreleased]
 ### Added
-* Support for specifying client secret hasher via CLIENT_SECRET_HASHER setting.
+* Support for Wildcard Origin and Redirect URIs, https://github.com/jazzband/django-oauth-toolkit/issues/1506
+<!--
 ### Changed
 ### Deprecated
 ### Removed
 ### Fixed
 ### Security
-  -->
+-->
+
 
 ## [3.0.1] - 2024-09-07
 ### Fixed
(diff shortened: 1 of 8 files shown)


tonial (approved):
Maybe you could add a test to ensure that `"https://*.example.com/*"` and `*` are not valid `redirect_uri`.

Apart from that, It's all good for me :+1:

dopry:
> Maybe you could add a test to ensure that `"https://*.example.com/*"` and `*` are not valid `redirect_uri`.

`*` is a valid character in a uri path so `https://*.example.com/*` is valid, but * is not treated as a wild card. 
I did add a test for just `*` which raises invalid scheme

n2ygk (approved):
Looks good. Thanks.

PR #1512, opened 2024-10-03
Add client_secret to sensitive_post_parameters

jaap3:
The client_secret is posted to the token endpoint when using the client_credentials grant.

(sorry just a basic drive-by PR because I noticed this while debugging something else)


n2ygk (approved):
LGTM. Thanks.

matejsp:
Any chance to have this fix released? 
This looks like security issue when used with error capture like sentry or logging and would need release ASAP.

PR #1517, opened 2024-10-08
fix: OP prompts for logout when no OP session

dopry:
The OAuth provider is prompting users who no longer have an user session with the OAuth Provider to logout of the OP. This happens in scenarios given the user has logged out of the OP directly or via another client. In cases where the user does not have a session on the OP we should not prompt them to log out of the OP as there is no session, but we should still clear out their tokens to terminate the session for the Application.

I also renamed a test that used expired_session in the name to without_op_session. I believe that was their original intent, @tonial can you confirm? 

## Checklist

- [X] unit-test added
- [X] documentation updated

### DIFF
--- a/tests/test_oidc_views.py
+++ b/tests/test_oidc_views.py
@@ -311,6 +311,10 @@ def test_must_prompt(oidc_tokens, other_user, rp_settings, ALWAYS_PROMPT):
         == ALWAYS_PROMPT
     )
     assert RPInitiatedLogoutView(request=mock_request_for(other_user)).must_prompt(oidc_tokens.user) is True
+    assert (
+        RPInitiatedLogoutView(request=mock_request_for(AnonymousUser())).must_prompt(oidc_tokens.user)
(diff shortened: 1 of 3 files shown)


tests/test_oidc_views.py
```diff
@@ -577,29 +581,46 @@ def test_token_deletion_on_logout(oidc_tokens, logged_in_client, rp_settings):
...
+
     rsp = client.get(
         reverse("oauth2_provider:rp-initiated-logout"),
         data={
             "id_token_hint": oidc_tokens.id_token,
             "client_id": oidc_tokens.application.client_id,
         },
     )
-    assert rsp.status_code == 200
+    assert rsp.status_code == 302
```
n2ygk, line 599: why is this changed from a 200 to a 302 redirect?

n2ygk (approved):

tonial:
@dopry yes, that was the case i was covering

PR #1543, opened 2025-01-27
Session management

### LINKED ISSUE #1130: OpenID Connect Session Management 1.0
Implement [OpenID Connect Session Management](https://openid.net/specs/openid-connect-session-1_0.html)

* [ ] add chcek_session end point per [Section 3.2](https://openid.net/specs/openid-connect-session-1_0.html#rfc.section.3.2)
* [ ] add check_session_iframe to discovery metadata per [Section 3.3](https://openid.net/specs/openid-connect-session-1_0.html#rfc.section.3.2)

lullis:
Fixes #1130 

## Description of the Change

Implementation of OIDC Session Management. This PR:

 - Adds some settings variables to enable OIDC Session Management
 - Adds checks to validate OIDC Session Management settings - if it is enabled, then Django will require a `OIDC_SESSION_MANAGEMENT_DEFAULT_KEY` to be present.
 - Adds the session_state parameter to the parameters given to the end user client upon succcessful authorization. The session state is calculate by a salted hash combination of client id, origin and "OP user agent state", which for the intents of session management is itself just a hashed value of the session key for authenticated users, and `OIDC_SESSION_MANAGEMENT_DEFAULT_KEY` if the user is not authenticated. This is enough for the OP to indicate whether the end user session has changed (logged in, logged out) 
 - Add a OIDCSessionManagementMiddleware that sets a cookie with the value of "OP user agent state" 
 - Adds a SessionIFrameView endpoint for the "[OP IFrame](https://openid.net/specs/openid-connect-session-1_0.html#OPiframe)"

## Checklist

- [x] unit-test added
- [X] documentation updated


dopry:
@lullis Thank you for working on this.  I've only given it a cursory review, but it looks like a great start. It'll probably be late next week before I can look at this closely.  Ping me again if I don't manage to get you a review by next Friday.

Qup42:
@lullis I'm quite busy at the moment. I think that I should be able to have a look at this in the next ~3 months. Do please remind me if I forget it.

lullis:
@Qup42, just pinging to remind you about this PR.

dopry:
@lullis  any guidance on testing procedures for this PR?

lullis:
@dopry I got a small app going to test these features at https://codeberg.org/raphael/oidc-client-testbed.

I won't make any strong claims about its overall quality (I relied on Cursor to generate a lot of the boilerplate and the vue components), but I did check the actual functionality and it works well enough to check the session frame part.

dopry:
@lullis   I was wondering more about test plans I could follow for testing manually. I can try to use your OIDC test bed, we also have the test/app/rp which we can use, although it may need to be updated. Not sure if I have session handling in that client.

lullis:
@dopry, I am reasonably confident that there is nothing about session handling in that client - at least not related to the session iframe. If started my vue app precisely because I couldn't find any demo application that exercised this functionality.

Anyway, I was testing this with my app by running both the idp and the vue app. You can then login through the vue app, which will authenticate you in the idp. Once you login there will be a "session info" tab that will show you as logged in. You can open a console and you will see that the rp has an iframe which will be pinging the idp every few seconds. If you open another tab and log *out* directly on the IDP, the session will be finished and the rp will then show you as logged out.

docs/oidc.rst
```diff
@@ -381,6 +381,39 @@ token, so you will probably want to reuse that::
...
+<https://openid.net/specs/openid-connect-session-1_0.html>`_
+specification defines how to monitor the End-User's login status at
+the OpenID Provider on an ongoing basis so that the Relying Party can
+log out an End-User who has logged out of the OpenID Provider.
+
+To enable it, you will need to add
+``oauth2_provider.middleware.OIDCSessionManagementMiddleware`` to MIDDLEWARES and set
+``OIDC_SESSION_MANAGEMENT_ENABLED`` to ``True`` on
+``OAUTH2_PROVIDER``. You will also need to provide a string on
+``OIDC_SESSION_MANAGEMENT_DEFAULT_SESSION_KEY``. This setting is
```
dopry, line 444: `OIDC_SESSION_MANAGEMENT_DEFAULT_SESSION_KEY` needs to be documented in the settings docs. What is it? Why is it needed to ensure consistent browser state?  Is this a arbitrary string?  Are there any security considerations for it?

docs/settings.rst
```diff
@@ -353,6 +359,18 @@ this you must also provide the service at that endpoint.
 If unset, the default location is used, eg if ``django-oauth-toolkit`` is
 mounted at ``/o/``, it will be ``<server-address>/o/userinfo/``.
 
+OIDC_SESSION_IFRAME_ENDPOINT
```
dopry: Why allow an iframe endpoint setting? It seems like this could become a support headache. Couldn't someone just override the url view if they wanted?

dopry (changes requested):
Looking good... just two items stand out to me at the moment upon a quick review.

docs/settings.rst
lullis: Good question. To be honest I was just following the pattern established by the UserInfoView, which seems to do something similar?

docs/oidc.rst
lullis: [As I understood it](https://openid.net/specs/openid-connect-session-1_0.html#OPiframe), the default key is meant to provide an unique value to the client_id, client_origin and session_identifier for *users who are not yet authenticated* and therefore have no request.session_key.

Without it, any user without a session would have different starting values for the session state, so the "get user agent state" would have different values for unauthenticated users.

I will add some more information in the docs for it.

lullis:
@dopry added missing tests, looks like codecov is happy now...

dopry:
A way you can help streamline the review process for me is lay out an easy end to end tests for me using the tests/rp and tests/idp or some other setup using docker compose so I can verify the functionality more easily without needing to implement a client to test each of these features myself.

lullis:
> or some other setup using docker compose

Can you please check if https://codeberg.org/raphael/oidc-client-testbed would work for you? It does have a way to test the session iframe. I can make a docker image if you want, as well.

dopry:
I only have a few hours a week for work on this project... Having to spin up and configure a 3rd party example will eat up all my review time for a week. 3rd Party examples do not create an easy way for other developers to test the features in the future, nor does it provide example implementations for consumers. I'd really like to keep all the functionality demonstration in tests/app/[idp,rp]. That said I'll try to get your example setup to test... just don't expect any additional review by me this week.

lullis:
@dopry , I've made changes to the rp client. Let me know if this works for you.

dopry:
@lullis  I'll try to get to a review next week. Of the open ones, which would you prefer to see merged first?

lullis:
Backchannel logout, then session management, then RP-initiated registration. But any order is fine, if you think the PR is good to merge.

oauth2_provider/middleware.py
```diff
@@ -64,3 +65,24 @@ def __call__(self, request):
...
+
+        cookie_name = oauth2_settings.OIDC_SESSION_MANAGEMENT_COOKIE_NAME
+        is_authenticated = request.user.is_authenticated
+        session_key = request.session.session_key if is_authenticated else None
+        if is_authenticated and session_key is not None:
+            session_key_bytes = session_key.encode("utf-8")
+            hashed_key = hashlib.sha256(session_key_bytes).hexdigest()
+            response.set_cookie(cookie_name, hashed_key, samesite="None", secure=True)
+        else:
+            response.delete_cookie(cookie_name, samesite="None")
```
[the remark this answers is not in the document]
lullis: fixed by using the same function to derive the state key, but will keep deleting the cookie for anonymous users.

dopry:
@lullis  I've rebased and addressed some of the copilot issues.. tell me a bit more about this docker-compose.yml you've added. What is it for?

lullis:
This docker compose file is just something I use to get a public endpoint for the test app. It does require a valid domain that is managed on cloudflare, so I'd guess it doesn't make sense for everyone. I've updated it with instructions in case someone else wants to use it, but if you are not interested in it you can remove it.

dopry:
@lullis I wanted to reach out... I've been busy the last few days trying to get a bug fix release out. I also did an architecture review and found some changes I'm thinking to land before this. https://github.com/django-oauth/django-oauth-toolkit/issues/1723,  I would appreciate your thoughts since it was reviewing your work that got be started on the deeper investigations in DOT architecture that led to that work.  My current thought is to land phase 1 of #1723 as 4.x and move this work to a 4.x series with a more stable foundation and resolve some of the underlying bugs/workarounds for session handling. I would probably merge phase 2 of the ADR with this particular PR.  thoughts?

dopry:
@lullis  I'm strongly considering closing this PR, and I want to be upfront that this is not a reflection of the quality of the work. This was a complete, well-tested, well-documented implementation of the spec,

I'm thinking DOT shouldn't ship Session Management 1.0 becuase the check_session_iframe channel depends on the OP iframe reading its cookie in a third-party context. Safari blocks all third-party cookies outright and Firefox partitions them, so the mechanism silently fails or fires false logouts for a large number of users. While chrome backed off the block, it seems like Apple and Mozilla are committed.  So it will create an inconsistent experience for a large number of users and support overhead for anyone deploying DOT. 

I took a pass at rectifying this PR with the updates proposed in #1723, and the middleware session thing would need to be reworked on the  proposed Session model, but I'm questioning whether the additional effort is worth it given the third party cookie issues.  

thoughts? Is this something you **need**?

Natureshadow:
>thoughts? Is this something you **need**? 

Yes. We don't normally invest a year of work and thousand of Euros in things we don't need.

If it gets rejected, we will fork DOT and maintain a version that includes Session Management.

dopry:
Sunk cost isn't a need.  Let's stick to technical requirements and use cases. What is your current use case? Do you have an electron client or similar that can reliably leverage 3rd party cookies? Do you have mitigation strategies for dealing with browsers that don't support 3rd party cookies? Is it impossible to leverage front and back channel logout in your use case, instead of session management?

Natureshadow:
I did not use sunk cost as an argument. I explained to you that I wouldn't have invested time and money if this weren't needed.

I have no idea why you question whether this is needed or not, when obviously it **is** needed, as you can conclude from the PR being here and the fact that we invested tons of time and money in it.

dopry:
> I did not use sunk cost as an argument. I explained to you that I wouldn't have invested time and money if this weren't needed.
> 
> I have no idea why you question whether this is needed or not, when obviously it **is** needed, as you can conclude from the PR being here and the fact that we invested tons of time and money in it.

I was asking you to clarify your specific use case and how sessions will will still work in your use case with the changes to 3rd party cookie support across browsers.  The only thing you seems to have mentioned so far, is that you've invested time an money in the feature.  Frankly, your investment has no bearing on my decisions as a maintainer. Neither do emotional arguments. I strongly encourage you to make a technical case for the feature.

PR #1547, opened 2025-01-29
Fix range of allowed django versions

lullis:
Fixes #

## Description of the Change
Running `docker-compose build` fails at the moment due to a package resolution conflict. Allowing to use Django 4.2 solves the issue.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/tests/app/idp/requirements.txt
+++ b/tests/app/idp/requirements.txt
@@ -1,4 +1,4 @@
-Django>=3.2,<4.2
+Django>=3.2,<=4.2
 django-cors-headers==3.14.0
 django-environ==0.11.2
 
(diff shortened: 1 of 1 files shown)


tests/app/idp/requirements.txt
```diff
@@ -1,5 +1,5 @@
-Django>=3.2,<4.2
+Django>=3.2,<=4.2
```
n2ygk, line 1:
```suggestion
Django>=4.2,<=5.1
```

Shouldn't the versions match those [here](Django 4.2, 5.0 or 5.1)?

n2ygk (changes requested):
It looks like updates to the version requirements for the test app got missed. Please try with the up to date version requirements.

n2ygk (changes requested):
Thanks. I didn't notice on prior review. Please update the intial comment to fill out the pull request template checklist and take all the listed actions.  I've edited the comment to include the template.

n2ygk (approved):
On second thought, this is such a tiny dependency-only change that there's not much to the checklist. Thanks for this.

PR #1549, opened 2025-01-30
Easier docker setup for development environment

### LINKED ISSUE #1548: Dockerfile for idp test app not using correct paths
**Describe the bug**

When building the idp/rdp test apps, the idp server fails when initiating login, as it can not find the `registration/login.html` template.

**To Reproduce**
 - Run `docker-compose up -d --build` to build the docker image and start the services.
 - Go to the rdp app (localhost:5173)
 - Click "login"
 - Results in Server Error

**Expected behavior**
 - The server should be able to find the login template and render the page properly.

**Version**

**Additional context**
Is this "tests project" meant *only* for testing, or could/should it be used for development as well? It looks like no one is actually using during development (otherwise other people would spot these glaring errors), so perhaps I shouldn't be spending so much time in order to get this going.

lullis:
I couldn't find anything in the documentation indicating how to set up the docker service for development. The  Dockerfile is outdated (the second target is setup to run the idp project, but the env variables defined are not matching) and not really reflecting the current tooling used elsewhere. The docker-compose also does not have any mechanism to set an environment file which makes it hard to make changes.

Fixes #1548 

## Description of the Change

The Dockerfile now is used only to give a container with all the installed dependencies, and the docker-compose is focused on a development workflow, the application code is loaded as a volume, so changes in the code repository are immediately available inside the container. Environment variables that are specific to the development environment are defined in the dockerc-compose file, and environment variables that might be controlled by the developer (e.g, whether to enable or disable OIDC), can be configured through a .env file placed at the root of the project.

## Checklist

- [ ] unit-test added
- [X] documentation updated

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -12,10 +12,30 @@ This is a `Jazzband <https://jazzband.co>`_ project. By contributing you agree t
 Setup
 =====
 
-Fork ``django-oauth-toolkit`` repository on `GitHub <https://github.com/jazzband/django-oauth-toolkit>`_ and follow these steps:
 
- * Create a virtualenv and activate it
- * Clone your repository locally
+Fork ``django-oauth-toolkit`` repository on `GitHub
+<https://github.com/jazzband/django-oauth-toolkit>`_ and clone to your
+development machine. The recommended method to manage your development
+environment is to use `uv <https://docs.astral.sh/uv/>`_, but you can
+also use pip directly to manage the virtual environment.
+
+
+Running the test IDP and RP applications (Docker)
+=================================================
+
+During development, it might be helpful to run an application that
+actually exercises the many authentication flows. We provide a django
+project that can be set up to run as the Identity Provider (IDP) and a
+separate Javascript application that works as the Relying Party (RP).
+If you have docker and docker-compose configured, you can get them
+both the services running with a few simple steps:
+
+ - make a copy of the ``.env.sample.docker`` to ``.env`` and make changes as desired
+ - Run ``docker-compose up -d --build``
+
+The IDP service will be available on `http://localhost:8000` and the
+RP service will be available on `http://localhost:5173`
+
 
 Issues
 ======
--- a/.env
+++ b/.env
@@ -1,2 +1,0 @@
-# required for vscode testing activity to discover tests
-DJANGO_SETTINGS_MODULE=tests.settings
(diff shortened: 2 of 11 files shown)


lullis:
Looks like the RP javascript app also needs to be changed.

dopry:
I appreciate the contribution, but some of your changes are counter to our intent for the docker compose environment.  The /tests/app/idp is a proof of concept we're working on as a potential app that represents a best practice DOT IDP that people can run in their own containerized infrastructure. It's not meant to be a 'development' container. The changed you've made here would make it impossible to distribute in the future.

The compose environment itself is mostly for maintainers and testers to stand up an environment quickly for QA and Pull Request code review. 

Can you clarify what issues you had starting the environment initially and what you mean by 
> the env variables defined are not matching

docs/contributing.rst
```diff
@@ -12,10 +12,30 @@ This is a `Jazzband <https://jazzband.co>`_ project. By contributing you agree t
 Setup
 =====
 
-Fork ``django-oauth-toolkit`` repository on `GitHub <https://github.com/jazzband/django-oauth-toolkit>`_ and follow these steps:
 
- * Create a virtualenv and activate it
- * Clone your repository locally
+Fork ``django-oauth-toolkit`` repository on `GitHub
+<https://github.com/jazzband/django-oauth-toolkit>`_ and clone to your
+development machine. The recommended method to manage your development
```
dopry, line 18: `uv` looks interesting but this is separate from the bug you're trying to address. I'd say open an issue for discussion of uv if you think it is something the project could benefit from and why you think it is better than other project tools like pip, poety, pipenv, hatch, pdm, rye, etc.

dopry:
I'm closing since #1550 was merged and should resolve #1548. I appreciate the work on this. You've presented some interesting ideas that are outside the scope of the issue, that I think warrant discussion.  In particular a more mature package manager and a development container.  Feel free to open issues for each of them independently as DX improvements and demonstrate how they're make our maintainers' and contributors' lives easier.

docs/contributing.rst
lullis: Apologies, I wrongly assumed that uv was the detault tool being used by package management. I **know** that it isn't poetry or pipenv because these tools have their own format requirements, but I should have considered that standard setuptools should be enough.

lullis:
> The compose environment itself is mostly for maintainers and testers to stand up an environment quickly for QA and Pull Request code review.

That is reasonable, but the question I have now is "how are people developing against this in the first place?"

Should I just create a separate project and install DOT in editable mode next to it?

dopry:
By running the idp locally, not in the container

PR #1562, opened 2025-04-09
Fix: Handle AttributeError in IntrospectTokenView

dawidwolski-identt:
## Description of the Change

Handle `AttributeError` in `get_token_response`. If there is no `token` parameter in introspect endpoints they raise `AttributeError`.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/views/introspect.py
+++ b/oauth2_provider/views/introspect.py
@@ -2,7 +2,7 @@ import calendar
 import hashlib
 
 from django.core.exceptions import ObjectDoesNotExist
-from django.http import HttpResponseBadRequest, JsonResponse
+from django.http import JsonResponse
 from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
 
--- a/oauth2_provider/views/introspect.py
+++ b/oauth2_provider/views/introspect.py
@@ -2,7 +2,7 @@ import calendar
 import hashlib
 
 from django.core.exceptions import ObjectDoesNotExist
-from django.http import JsonResponse, HttpResponseBadRequest
+from django.http import HttpResponseBadRequest, JsonResponse
 from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
 
--- a/oauth2_provider/views/introspect.py
+++ b/oauth2_provider/views/introspect.py
@@ -33,7 +33,7 @@ class IntrospectTokenView(ClientProtectedScopedResourceView):
                 .objects.select_related("user", "application")
                 .get(token_checksum=token_checksum)
             )
-        except ObjectDoesNotExist:
+        except (AttributeError, ObjectDoesNotExist):
             return JsonResponse({"active": False}, status=200)
         else:
             if token.is_valid():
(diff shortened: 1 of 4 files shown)


jackwebidentt (approved):

dawidwolski-identt:
Contribution guidelines says
> Make sure to request a review by assigning Reviewer jazzband/django-oauth-toolkit.

but I do not have permission to set assignee.

dopry:
I'm not 100% sure this is the correct handling. Per https://datatracker.ietf.org/doc/html/rfc7662#section-2.3,
> Note that a **properly formed** and authorized query for an inactive or
   otherwise invalid token (or a token the protected resource is not
   allowed to know about) is not considered an error response by this
   specification.  In these cases, the authorization server MUST instead
   respond with an introspection response with the "active" field set to
   "false" as described in [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).

I'm not sure that a missing token would qualify as properly formed. I feel that a 400 might be the proper response in this case since the request is missing the required 'token' parameter. My instinct says a present, but invalid or blank token, i.e. a body  of ```
token=
```
would return an `active: false`, but missing `token=` altogether would be malformed and we should let the caller know they're sending a malformed request, since there is a bug in their implementation. I feel returning a 200 `active: false` in the case of a malformed request may give consumers/developers confidence in a broken client implementation.

File: oauth2_provider/views/introspect.py [hunk 1]
```diff
@@ -35,6 +35,11 @@ def get_token_response(token_value=None):
             )
         except ObjectDoesNotExist:
             return JsonResponse({"active": False}, status=200)
+        except AttributeError:
```
dopry, line 38: What if something other than the token raised the attribute error? Is there a way to proactively test the token so we avoid the overhead of exception handling.

dawidwolski-identt: I think it is a proper place to check for the presence of a token. It's the first place where the token is accessed. To move this check one level up, we can use `request.GET.get("token", None)` without a default, but then we need to handle the missing token twice.

dopry: I see both get and post `return self.get_token_response(request.[GET|POST].get("token", None))`

Why not check `if token_value is None:` before calculating a checksum, rather than waiting on `token_value.encode` to raise an `AttributeError` and avoid the overhead of handling an exception unnecessarily.

dawidwolski-identt: Okay, I've changed that. Take a look

merito: @dopry I don't have access to the company account I used earlier. You can reach me through this account from now on.

dopry: You can probably check out the branch into your new account and open a new PR...

merito: Do I need to do this? It tooks ridiculously long to merge such a simple fix. What else is needed here?

dopry: oh sorry I misunderstood. I thought you were saying you no longer had access to the branch and needed to make changes. Now I realize you're just letting me know your account changed. Sorry for the confusion. I just rebased, hopefully the test coverage will be good and I can merge it.

merito: The test workflow needs approval to run

merito: The coverage looks good

PR #1567, opened 2025-04-18
Improve settings documentation

brianhelba:
## Description of the Change
Previously, it was unclear that the swappable model settings should always be set without a namespace, as the sub-section titles didn't include the required `OAUTH2_PROVIDER_` prefix. The warning at the top may not be noticed by people looking for a specific settings, and it was still unclear given the sub-section titles.

Additionally, this documents the `OAUTH2_PROVIDER_ID_TOKEN_MODEL` and `OAUTH2_PROVIDER_GRANT_MODEL` settings, which were previously undocumented.

Also, this corrects a mistake where `ACCESS_TOKEN_GENERATOR` was mentioned as the setting which enables the use of `SettingsScopes`; the actual setting is `SCOPES_BACKEND_CLASS`.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -1,10 +1,8 @@
 Settings
 ========
 
-Our configurations are all namespaced under the ``OAUTH2_PROVIDER`` settings with the exception of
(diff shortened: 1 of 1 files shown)


n2ygk (approved):
thanks!

PR #1568, opened 2025-04-21
Remove unnecessary select for removing an auth code grant

q0w:
Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/oauth2_validators.py
+++ b/oauth2_provider/oauth2_validators.py
@@ -322,10 +322,10 @@ class OAuth2Validator(RequestValidator):
 
         :raises: InvalidGrantError if the grant does not exist.
         """
-        try:
-            grant = Grant.objects.get(code=code, application=request.client)
-            grant.delete()
-        except Grant.DoesNotExist:
+        deleted_grant_count = Grant.objects.filter(
+            code=code, application=request.client
+        ).delete()
+        if not deleted_grant_count:
             raise errors.InvalidGrantError(request=request)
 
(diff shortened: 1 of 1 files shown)


dopry:
What's the upside to this change?

oauth2_provider/oauth2_validators.py
```diff
@@ -322,7 +322,7 @@ def invalidate_authorization_code(self, client_id, code, request, *args, **kwarg
 
         :raises: InvalidGrantError if the grant does not exist.
         """
-        deleted_grant_count = Grant.objects.filter(code=code, application=request.client).delete()
+        deleted_grant_count, _ = Grant.objects.filter(code=code, application=request.client).delete()
```
dopry, line 325: if we're not going to use `_`, let's not assign it into scope.

dopry: is deleted_grant_count the number of deleted records or the tuple when you do not specify the , _?

q0w: https://docs.djangoproject.com/en/5.2/ref/models/querysets/#delete

>returns the number of objects deleted and a dictionary with the number of deletions per object type.

dopry: okay, yeah, put the _ back. I've been working with TS a lot lately  and having trouble switching my brain back to python. mode. Sorry for the run around.

93578237: Done

dopry (approved):

PR #1573, opened 2025-05-06
Backchannel Logout

### LINKED ISSUE #1545: Support for "Back-Channel Logout"
Currently DOT does not provide a way for the OP to indicate that the user has signed out, this causes RP sessions to be kept open in end 
 users agents even when they user believes that they signed out already.  

[Back-Channel Logout](https://openid.net/specs/openid-connect-backchannel-1_0.html) allows for more reliable communication between RP and OP to indicate that the end user wishes to terminate its login session. 

**Additional context**
This is part of the improvements ongoing for the [Aleksis Project](https://nlnet.nl/project/AlekSIS/), which is extensively using DOT.

lullis:
- Add migration to add backchannel logout support
 - Add parameters related to backchannel logout on OIDC Discovery View
 - Change application creation and update form
 - Change template that renders information about the application
 - Add "handlers" module for signals handlers definition.

Fixes #1545 

## Checklist

- [x] unit-test added
- [x] documentation updated


dopry:
@lullis 
1. for now the back channel can live in the logout request, we can move it later.  Maybe start with a fire and forget approach to minimize blocking.  Rather than logging out as a part of the OP logout, maybe we have a list of clients to explicitly logout without necessarily terminating the OP session?
2. You should be using the sid from the tokens I believe. Do we issue the sid claim currently?  If not this may not be applicable at this juncture.
3. not sure off hand. will look more closely, but try to following existing patterns where possible.

lullis:
@dopry 

1. Maybe I am missing something, but how would you set up any fire-and-forget function handler in the response-request cycle? 

2. Yeah, I don't think there is anything about the sid claim. So would it be okay if we had a first version where the server supports backchannel logout, but not the session identifier?

oauth2_provider/models.py
```diff
@@ -650,6 +657,71 @@ class Meta(AbstractIDToken.Meta):
         swappable = "OAUTH2_PROVIDER_ID_TOKEN_MODEL"
 
 
+class AbstractLogoutToken(models.Model):
```
dopry: Why store the logout token in the database? Can't we just generate as needed to sent to the RP's?  Is there a reason to look at it later? It seems like an ephemeral resource to me.

lullis: My idea was to have the token in the database to have some retry mechanism. But I agree that as-is, it is not needed. Should I remove it?

dopry: Yes. Please remove it. I don't want to add additional things we need to support, especially if it'll just aggregate in a DB leading to storage costs and unecessary database overhead for the users of DOT.

lullis: Done.

dopry:
@lullis it looks like the next step is expanding test coverage right now the patch is just shy of 70%, our target is 100% coverage on patches going forward. At the very least we aim to match the current total coverage.

lullis:
I will take a look to see if I can run the report of the diff locally. Running coverage for the whole report showed that all files touched were at least at 99% coverage.

dopry:
@lullis  I hear you. Right now I'm in a bit of a holding pattern waiting on @jezdez to transfer the project out of jazzband so we can fix our codecov setup and get back to merging code. He was supposed to do it the last two weeks, but flaked each time.

hansegucker:
Hi @dopry, is there any update?

dopry:
No @jezdez  still hasn't transferred the repo.

lullis:
@dopry , I've increased the test coverage as much as I could, but it codecov is hung up on one line on application not being covered. Any tips on the best approach to test that exceptions are being caught in the middle of a function, or is it okay as it is?

oauth2_provider/models.py
```diff
@@ -629,6 +636,53 @@ def revoke(self):
         """
         self.delete()
 
+    def send_backchannel_logout_request(self, ttl=timedelta(minutes=10)):
```
dopry: Why put the `send_backchannel_logout_request`  method on AbstractIdToken? It doesn't operate on the token and the token is not an entity I think of as being responsible for sending requests. I feel like it's responsibility it carrying the identity data. A standalone function in handlers.py that receives an id_token as an argument would be more straight forward approach and would keep the handler bundled with where it is used for easier discoverability if someone were to want to extend it's scope by wrapping it in a proxy function and overriding it.

oauth2_provider/checks.py
```diff
@@ -26,3 +26,16 @@ def validate_token_configuration(app_configs, **kwargs):
...
 
     return []
+
+
+@checks.register()
+def validate_backchannel_logout(app_configs, **kwargs):
+    errors = []
+
+    if oauth2_settings.OIDC_BACKCHANNEL_LOGOUT_ENABLED:
+        if not callable(oauth2_settings.OIDC_BACKCHANNEL_LOGOUT_HANDLER):
```
dopry: OIDC_BACKCHANNEL_LOGOUT_HANDLER is not documented. What are the use cases for making a setting to override the logout handler? Is the specification flexible enough that it makes sense to give people the option to override this?

oauth2_provider/models.py
```diff
@@ -859,3 +913,15 @@ def is_origin_allowed(origin, allowed_origins):
             return True
 
     return False
+
+
+def send_backchannel_logout_requests(user):
```
dopry: per the comment on id_token.send_backchannel_logout_request, let move this to the handlers.py...

dopry (changes requested):
We're pretty close to an initial implementation. There are some organizational issues I'd like to see addressed and I'm on the fence about whether it is wise to include OIDC_BACKCHANNEL_LOGOUT_HANDLER  or drop it for now until we get some firm use cases for it.

oauth2_provider/models.py
lullis: The original implementation had a LogoutToken model for this, and I ended up moving to IDToken because most of the data needed is there. I am not a fan of putting too much business logic in a handlers module, but I can move it there if you prefer it that way.

oauth2_provider/checks.py
lullis: The main use-case for a configurable handler is simple: I would like to give consumers of this library a hook to let them execute this function outside of the request cycle. For example, if someone wants to run this as a celery task or implement a function that can keep track of errors.  I will add it to the documentation.

oauth2_provider/models.py
dopry: I am also not a fan of putting too much business logic in a handler module. I recognize much of the data we need it on the AbstractIdToken, but the method doesn't really depend (like a dependency injected at construction time) on it as much as requires the token as parameter at call time.  I'm also not a fan of handler functions floating around in the models.py such that importing them into a handler module required pulling in the entirety of models and I've have bad experiences with circular dependencies when I do that and need to defer the import to inside the function. Maybe there is another more appropriate place if we want to make it common and shared... the first that comes to mind is oauth2_backends.OAuthLibCore, which can already be overridden via oauth2_settings.OAUTH2_BACKEND_CLASS... or maybe backends.OAuth2Backend... Give it some thought and I'll try to do the same over the next couple days.

oauth2_provider/checks.py
dopry: Delegating to celery sounds reasonable, especially if people want to implement retries etc. Do we not have logging for error tracking or are you thinking of some particular 3rd party instrumentation? Can you expand on your thoughts in regard to error tracking?

lullis: I pushed some changes, you will see that now the handler is made to work on id_tokens, not on the user directly.

My idea was to leave it completely open to the developer using DOT. The way I see it, someone that wants to have the requests on celery could do something like this:

1. On myapp/tasks.py

```python

from celery import shared_task
from oauth2_provider.models import get_id_token_model
from oauth2_provider.exceptions import BackchannelLogoutRequestError
from oauth2_provider.handlers import send_backchannel_logout_request as dot_request

IDToken = get_id_token_model()

@shared_task
def send_backchannel_logout_request(id_token_id):
    try:
        id_token = IDToken.objects.get(id=id_token_id)
        dot_request(id_token=id_token)
    except IDToken.DoesNotExist:
        handle_missing_id_token()
    except BackchannelLogoutRequestError:
        handle_error()
        
def schedule_logout_request(id_token):
     send_backchannel_logout_request.delay(id_token.id)
```

2. on myproject/settings.py

```python

OAUTH2_PROVIDER = {
   "OIDC_BACKCHANNEL_LOGOUT_ENABLED": True,
   "OIDC_BACKCHANNEL_LOGOUT_HANDLER": "myapp.tasks.schedule_logout_request"
}
    
```

oauth2_provider/models.py
dopry: @lullis any thoughts on better locations for the send_backchannel_logout_request function?

lullis: Yeah, I was taking a look at it seems that maybe it would make sense to have a OidcBackend which could be the class for implementing backchannel logout. What do you think?

dopry:
@lullis  I've been thinking about this a bit.  We need to be more selective about what clients we notify. We definitely shouldn't send a logout request for every id token.  Reviewing the spec

> After receiving a valid Logout Token from the OpenID Provider, the RP locates the session(s) identified by the iss and sub Claims and/or the sid Claim. The RP then clears any state associated with the identified session(s). The mechanism by which the RP achieves this is implementation specific. If the identified End-User is already logged out at the RP when the logout request is received, the logout is considered to have succeeded.

Since DOT doesn't have any session tracking and we can only send  a token with a sub, we should only send 1 request per client with a back channel logout url to which the user has a refresh token without offline_access or access token  (not tied to a refresh token with offline?)

Natureshadow:
Ping @lullis !

thisislawatts:
Thanks @lullis for all the time and energy getting the PR to the current state. I am sketching out a solution for my local Makerspace and this functionality would be a great option!

I believe the test failure maybe tied to: https://github.com/django-oauth/django-oauth-toolkit/pull/1631

Natureshadow:
@dopry Does @lullis ' change in e229dac address your requested changes?

oauth2_provider/handlers.py
```diff
@@ -0,0 +1,106 @@
...
+        .order_by("application", "-expires")
+    )
+
+    # Group by application and send one request per application
+    applications_notified = set()
+    for id_token in id_tokens:
+        if id_token.application not in applications_notified:
+            applications_notified.add(id_token.application)
+            try:
+                handler(id_token=id_token)
```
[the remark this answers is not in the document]
lullis: I updated the docs to reflect that it needs to be a keyword argument.

dopry:
@lullis thank you for attending to the requests of our pedenatic overlords.  There is another round.  I wish I could push to your branch here becuase I'd be happy to address these and put a bow on it... unfortunately I don't seem to have permissions to do that with org forks as I do with individual contributors.

oauth2_provider/models.py
```diff
@@ -152,6 +153,9 @@ class AbstractApplication(models.Model):
         help_text=_("Allowed origins list to enable CORS, space separated"),
         default="",
     )
+    backchannel_logout_uri = models.URLField(
+        blank=True, null=True, help_text=_("Backchannel Logout URI where logout tokens will be sent")
```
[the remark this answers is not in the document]
dopry: We should validate to the spec, but not with oauth2_settings.ALLOWED_REDIRECT_URI_SCHEMES as they're for redirect uris. Also keep in mind that the spec  says
```
   backchannel_logout_uri
      OPTIONAL.  RP URL that will cause the RP to log itself out when
      sent a Logout Token by the OP.  This URL SHOULD use the "https"
      scheme and MAY contain port, path, and query parameter components;
      however, it MAY use the "http" scheme, provided that the Client
      Type is "confidential", as defined in Section 2.1 of OAuth 2.0
      [RFC6749], and provided the OP allows the use of "http" RP URIs.
``` 
and we are permissive with `http` for non-confidential clients on localhost for development and testing scenarios where certs aren't always easy to use.  see: docs/settings.rst ALLOWED_SCHEMES for some examples of the language we've used when allowing this in the past and the implementations for how we warn and communicate about these risks at run time.

lullis:
Hi @dopry. I understand it might help you to deal with some of the maintenance load, but to be 100% honest dealing with Copilot is quite demotivating, At least for me, it makes me less interested in becoming a more active contributor to the project.

Anyway, I gave you write permission to my repo. Let me know if this helps us make quicker progress.

dopry:
@lullis thanks for the permissions. I've confirmed they're working with a rebase. I'll support where I can and notify you when I make/push changes. 

I hear you regarding copilot. I'd love to hear more about what you find demotivating. I started a ["Copilot, thoughts?" discussion topic](https://github.com/django-oauth/django-oauth-toolkit/discussions/1680) for the conversation.

lullis:
@dopry , please provide some guidance about how to fix these multi-db tests that are failing. They have nothing to do with my PR and I'm honestly frustrated seeing a PR open for over an year.

dopry:
the multidb issue isn't  your problem.

dopry:
@lullis  I've rebased and reviewed the code. We still need to address validation of backchannel_logout_uri per the [comment](https://github.com/django-oauth/django-oauth-toolkit/pull/1573#discussion_r3019421573).

oauth2_provider/models.py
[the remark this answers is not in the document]
lullis: The reason I didn't implement this is that I am not sure how to implement it according to the spec *while* being permissive with http during development. I'd likely let `validators=[AllowedURIValidator(["https", "http"])]` always, and just check in the `clean` method for the cases where the application is *public* and the scheme is http. Then, if the check fails, check whether DEBUG is true to find out if we should issue an warning or raise ValidatorError.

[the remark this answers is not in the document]
dopry: Implemented in `e3cd372`, following dopry's reading of the spec rather than the suggestion above.

`backchannel_logout_uri` is validated against its own scheme list, `OIDC_LOGOUT_URI_ALLOWED_SCHEMES` (default `["https"]`), not `ALLOWED_REDIRECT_URI_SCHEMES` — a logout URI and a redirect URI are different roles and a deployment should be able to loosen one without the other. The list is shared with the front-channel logout URI, which §3.1 of Front-Channel Logout 1.0 constrains the same way.

Validation runs in `AbstractApplication.clean()` through `AllowedURIValidator(..., allow_path=True, allow_query=True)`, matching §2.2's "MAY contain port, path, and query parameter components". `http` is then accepted only when the client type is confidential, with the loopback carve-out this project already makes for development: `127.0.0.1` and `[::1]` always, `localhost` when `ALLOW_LOCALHOST_LOOPBACK` is enabled — the same trade-off `ALLOWED_SCHEMES` documents for redirect URIs.

It is deliberately in `clean()` rather than a `validators=` kwarg: field validators are serialized into migrations, which would freeze the scheme list at the time the migration was written instead of leaving it a runtime setting.

---

tests/app/rp/src/routes/api/logout-events/+server.js
```diff
@@ -0,0 +1,43 @@
...
+				},
+				close: () => controller.close()
+			};
+			addClient(sid, res);
+			// Remove client on stream close
+			controller.signal?.addEventListener('abort', () => {
+				removeClient(sid, res);
+			});
+		}
+	});
```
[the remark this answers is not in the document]
dopry: Fixed in `d127faf`. `ReadableStreamDefaultController` has no `signal`, so `removeClient()` never ran and the `clients` map grew without bound. Cleanup now happens in the stream's `cancel()` callback, which is what the platform invokes when the client disconnects.

---

tests/app/rp/src/routes/api/validateLogoutToken.js
```diff
@@ -0,0 +1,30 @@
...
+		}
+		// sub or sid must be present
+		if (!payload.sub && !payload.sid) {
+			throw new Error('Logout token missing sub and sid');
+		}
+		return payload;
+	} catch (e) {
+		throw new Error('Logout token validation failed: ' + e.message);
+	}
+}
```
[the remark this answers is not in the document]
dopry: Fixed in `d127faf`. `jwtVerify` is now given `issuer: ISSUER`, `audience: AUDIENCE` and `typ: 'logout+jwt'`, and the demo RP additionally rejects a token carrying a `nonce` (§2.6 step 9) or missing the `http://schemas.openid.net/event/backchannel-logout` event claim, so the sample walks through the §2.6 validation steps rather than just checking the signature.

---

docs/settings.rst
```diff
@@ -689,6 +689,28 @@ This setting is required when ``OIDC_RP_INITIATED_REGISTRATION_ENABLED`` is
...
+When is set to ``False`` (default) the `OpenID Connect Backchannel Logout <https://openid.net/specs/openid-connect-backchannel-1_0.html>`_
+extension is not enabled. OpenID Connect Backchannel Logout enables the :term:`Authorization Server` (OpenID Provider) to submit a JWT token to an endpoint controlled by the :term:`Client` (Relying Party)
+indicating that a session from the :term:`Resource Owner` (End User) has ended.
+
+OIDC_BACKCHANNEL_LOGOUT_HANDLER
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Default: ``oauth2_provider.authorization_server.oidc.handlers.send_backchannel_logout_request``
+
+Upon logout, the :term:`Authorization Server` (OpenID Provider)  will look for all ID Tokens associated with the user on applications that support Backchannel Logout. For every id token that is found, the function defined here will be called. The default function can be used as-is, but if you need to override or customize it somehow (e.g, if you do not want to execute these requests on the same HTTP request-response from the user logout view), you can change this setting to any function that takes ``id_token`` as a keyword argument.
+
```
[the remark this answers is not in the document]
dopry: Fixed in `40bd1cc`. `OIDC_BACKCHANNEL_LOGOUT_HANDLER` is now documented as being called **once per application**, with the most recently issued of that application's live ID Tokens, which is what §2.5 describes. The same section now also states the three things a custom handler has to know: it may run on a worker thread, the `id_token` it receives may be an instance whose row is already deleted (so a task queue handler must serialize claims rather than store a primary key), and DOT sends each request once.

---

oauth2_provider/authorization_server/oidc/handlers.py
```diff
@@ -0,0 +1,130 @@
...
+    ttl = kwargs.get("ttl") or timedelta(minutes=10)
+
+    if not oauth2_settings.OIDC_BACKCHANNEL_LOGOUT_ENABLED:
+        raise BackchannelLogoutRequestError("Backchannel logout not enabled")
+
+    if id_token.application.algorithm == AbstractApplication.NO_ALGORITHM:
+        raise BackchannelLogoutRequestError("Application must provide signing algorithm")
+
+    if not id_token.application.backchannel_logout_uri:
+        raise BackchannelLogoutRequestError("URL for backchannel logout not provided by client")
```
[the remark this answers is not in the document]
dopry: Confirmed and fixed in `b079473` — on `claude/django-oauth-toolkit-rebase-6n5vvu`, which is two commits ahead of this PR's head, so the fix is not in `40bd1cc` yet.

Reproduced before fixing: with `id_token.application = None`, `AttributeError: 'NoneType' object has no attribute 'algorithm'` did escape the documented single-exception contract. The body moved into a private `_send_backchannel_logout_request()`; the public `send_backchannel_logout_request()` is now a thin wrapper that re-raises `BackchannelLogoutRequestError` unchanged and wraps anything else with `raise BackchannelLogoutRequestError(str(exc)) from exc`. That covers the nullable `user` as well, plus key-material errors from `jwk_key` and jwcrypto signing failures. Regression test added.

---

docs/settings.rst
```diff
@@ -689,6 +689,28 @@ This setting is required when ``OIDC_RP_INITIATED_REGISTRATION_ENABLED`` is
 ``True``: if it is unset or cannot be resolved, ``ImproperlyConfigured`` is
 raised when a ``prompt=create`` request is received.
 
+OIDC_BACKCHANNEL_LOGOUT_ENABLED
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Default: ``False``
+
+When is set to ``False`` (default) the `OpenID Connect Backchannel Logout <https://openid.net/specs/openid-connect-backchannel-1_0.html>`_
+extension is not enabled. OpenID Connect Backchannel Logout enables the :term:`Authorization Server` (OpenID Provider) to submit a JWT token to an endpoint controlled by the :term:`Client` (Relying Party)
+indicating that a session from the :term:`Resource Owner` (End User) has ended.
```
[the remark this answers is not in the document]
dopry: Fixed in `b079473` — on `claude/django-oauth-toolkit-rebase-6n5vvu`, not yet in this PR's head. It now reads "When it is set to ``False`` (default)".

---

docs/settings.rst
```diff
@@ -837,6 +940,17 @@ The default is to delete the tokens of all applications if this flag is enabled.
...
+An RP-initiated logout revokes the user's access tokens and every refresh token derived
+from them, with no exemption for grants carrying the ``offline_access`` scope.
+`OpenID Connect Back-Channel Logout 1.0
+<https://openid.net/specs/openid-connect-backchannel-1_0.html>`__ section 2.7 says those
+"normally SHOULD NOT be revoked", though it says so in a section addressed to the
+:term:`Client` (Relying Party); `RFC 9700 <https://www.rfc-editor.org/rfc/rfc9700>`__
+section 4.14 separately permits an :term:`Authorization Server` to "revoke refresh tokens
+automatically in case of a security event, such as ... logout at the authorization
+server". Set this to ``False`` if the offline grants your relying parties hold must
+survive an interactive logout.
```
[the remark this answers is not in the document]
dopry: Correct — my earlier insert landed the paragraph after the `Resource Server settings` header. Fixed in `7b4b621` (on `claude/django-oauth-toolkit-rebase-6n5vvu`, not yet in this PR's head): the paragraph now sits directly under `OIDC_RP_INITIATED_LOGOUT_DELETE_TOKENS`, so "Set this to ``False``" resolves to that setting, and the resource-server section is back to starting with `RESOURCE_SERVER_INTROSPECTION_URL`.

---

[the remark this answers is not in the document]
dopry: Follow-up: the suppressed comment on the review of `7b4b621` flagged a second defect a few lines up, in the `OIDC_BACKCHANNEL_LOGOUT_HANDLER` section — an anonymous hyperlink target followed with no blank line by a paragraph indented one space. It's real, and worse than described. Built the docs both ways to check:

Before, docutils folded the paragraph into the target's URL — the rendered href was `...openid-connect-backchannel-1_0.htmlThedefaultfunctioncanbeusedas-is,butifyou…keywordargument.`, so the "Backchannel Logout" link was broken and the paragraph never appeared on the page. Contrary to the comment, **no warning was emitted** on sphinx 7.2.6 / docutils 0.20.1, so even a `-W` docs build would have passed.

Fixed in `9ca2dbe`, along with the `e.g,` → `e.g.` typo on the same line. Docs now build clean and the paragraph renders.

---

oauth2_provider/authorization_server/oidc/views.py
```diff
@@ -540,6 +550,17 @@ def must_prompt(self, token_user):
...
+        # below, and take ownership of the dispatch so the user_logged_out receiver does
+        # not also fire: it runs before django.contrib.auth.logout() flushes the session,
+        # and no relying party should be able to delay -- or, if the worker dies mid-
+        # request, prevent -- either the flush or the token deletion. RP-Initiated Logout
+        # 1.0 section 2 only requires the notifications to precede the post-logout
+        # redirect, which they still do.
+        logout_targets = collect_backchannel_logout_targets(
+            user if not isinstance(user, AnonymousUser) else None
+        )
+        setattr(self.request, DISPATCHED_ATTR, True)
```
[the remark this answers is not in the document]
dopry: Valid, and worth taking rather than leaving optional — RP-initiated logout is a common configuration on its own, and this made every one of those logouts pay for an ID Token query that could never produce a target.

Verified before fixing: with `OIDC_BACKCHANNEL_LOGOUT_ENABLED = False`, `collect_backchannel_logout_targets()` still ran its query and still returned a live ID Token.

Fixed in `6411d80`. The guard went inside `collect_backchannel_logout_targets()` rather than at the call site, so both callers get it and the function's contract stays "the targets to notify" — nothing to notify when the feature is off. A regression test asserts `assertNumQueries(0)` for that path.

I left `DISPATCHED_ATTR` set unconditionally. It costs nothing, and keeping it independent of the setting preserves the invariant that `RPInitiatedLogoutView` always owns dispatch for its request, rather than making ownership depend on when the setting happened to be read.

---

oauth2_provider/core/checks.py
```diff
@@ -361,3 +361,38 @@ def validate_access_token_expiry_configuration(app_configs, **kwargs):
         ]
 
     return []
+
+
+@checks.register()
```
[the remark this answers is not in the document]
dopry: Confirmed and fixed in `a0e0f4e`.

Proved the consequence before fixing rather than taking it on the inconsistency alone: with `OIDC_BACKCHANNEL_LOGOUT_ENABLED = True` and `OIDC_ENABLED = False` — a configuration where back-channel logout cannot work — `manage.py check --tag security` reported "System check identified no issues (0 silenced)". With the tag added it raises `SystemCheckError` on `OIDC_ENABLED must be True`.

Registered under `checks.Tags.security`, matching `validate_request_body_configuration` and `validate_access_token_expiry_configuration`. Not `deploy=True`: this guards a configuration that is broken in any environment, not just a deployed one, so it should run on a plain `manage.py check` too.

Added `test_backchannel_logout_check_is_tagged`, following the existing `test_check_is_tagged` in that module.

---

lullis:
@dopry , it looks like master has gone through a big refactor. I've rebased my branch to get aligned with it. Let me know if there is anything else you need from me or if you'd prefer to have done with automated assistance.

dopry:
Thanks for updating to match the reorg.. It looks like there is a test coverage gap.

PR #1592, opened 2025-08-12
chore: update from jazzband to django-oauth

dopry:
Fixes #

## Checklist

- [N/A] unit-test added
- [X] documentation updated
- [??] `CHANGELOG.md` updated (only for user relevant changes)


dopry:
@n2ygk wondering if you think this change is work mention in the changelog. I'm angling everything to be django-commons compatible for now...

PR #1593, opened 2025-09-02
use pk instead of id in clear expired tokens

sahama:
Fixes #

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -537,7 +537,7 @@ class AbstractRefreshToken(models.Model):
             self = list(token)[0]
 
             with suppress(access_token_model.DoesNotExist):
-                access_token_model.objects.get(id=self.access_token_id).revoke()
+                access_token_model.objects.get(pk=self.access_token_id).revoke()
 
             self.access_token = None
             self.revoked = timezone.now()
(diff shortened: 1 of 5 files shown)


oauth2_provider/models.py
```diff
@@ -532,7 +532,7 @@ def revoke(self):
             self = list(token)[0]
 
             with suppress(access_token_model.DoesNotExist):
-                access_token_model.objects.get(id=self.access_token_id).revoke()
+                access_token_model.objects.get(pk=self.access_token_id).revoke()
```
merito, line 540: I'm not sure if it will work

dopry:
@sahama  are you willing to complete the work on this PR or should we close it?

oauth2_provider/models.py
dopry: @merito  why not?

dopry:
@n2ygk  I implemented tests for this branch. I'd appreciate it if you could give it a double check. I feel adding the pk support is correct... however, I am concerned about adding one more thing we need to support and test against and one more edge case that can go wrong.

n2ygk (approved):
Using pk sounds like the right thing to do

PR #1597, opened 2025-09-16
Fix TypeError at /o/token/

aibaq:
Fixes #

## Description of the Change

Django User's last_login field is nullable, so `/o/token/` endpoint raises TypeError when last_login is null when using oidc

Снимок экрана 2025-09-16 в 13 05 42

This fix sets timezone.now() as a default value for auth_time

## Checklist

- [x] unit-test added
- [o] documentation updated (not needed)


dopry:
@aibaq this is a great start. to help me understand the issue a bit more, in what scenario did you encounter this issue?  I would expect that a user would have logged in at least once by the time get_id_token_dictionary was called. I'd like to understand how this was occurring to both correct my expectations and ensure there isn't a bug somewhere else that is allow a token to be generated for a user that isn't authenticated.  It also looks like there is a code path that is uncovered that will need to be tested.

dopry:
@aibaq  can you address the copilot comments and rebase this branch?

dopry:
@aibaq  thanks for shining a light on this. I'm going with another PR as the base for the solution since it had a better documented case for how this scenario happens.

superceded by https://github.com/django-oauth/django-oauth-toolkit/pull/1696#pullrequestreview-4394233800

PR #1601, opened 2025-10-12
Fix refresh token max length (occurs with Microsoft OAuth)

richardARPANET:
Fix for:

value too long for type character varying(255)

```
refresh_token = get_refresh_token_model().objects.create(
        user=org_user,
        token=token_data['refresh_token'],
        application=application,
        access_token=access_token,
        token_family=token_family,
    )

```

255 is far too short, as per OAuth guidelines. changed to 8000. Now Microsoft OAuth works.


dopry:
@richardARPANET  can you describe how this is setup with MS OAuth for me to better understand the setup for testing. I'd like to do a live test. We'll need a test case to cover this and we'll need to consider the performance implications of the larger token. Generally this happens when you're getting JWT tokens instead of opaque tokens...

dopry:
@richardARPANET hey do you have time to look into the copilot feedback? I like the idea of moving to a larger text field, but we need to confirm lookups are using checksums first to keep the indexes peformant. We'll also need the tests to get this in.

dopry:
I dug into this a bit today. We're using checksums on Access to resolve the token length issues. I'm going to put together a hash checking implementation for refresh tokens to address this issue without requiring us to index on a massive field that would break indexing in maria and mysql.

dopry:
@richardARPANET I appreciate you time and effort working on this.  I'm going with the implementation in #1720, I'd love your feedback on that PR if you have time time.

PR #1631, opened 2025-12-16
fix python version to allow 3.14.* (anything <3.15)

n2ygk:
testing for <= 3.14 breaks on minor releases like 3.14.2

## Description of the Change
Fixed pyproject.toml to test for <3.15

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "django-oauth-toolkit"
 dynamic = ["version"]
-requires-python = ">=3.8,<3.15"
+requires-python = ">=3.8"
 authors = [
   {name = "Federico Frenguelli"},
   {name = "Massimiliano Pippi"},


n2ygk:
@dopry not clear to me how uv.lock gets updated. I tried `uv lock` but it didn't change. Is this lock file even supposed to be checked in to the project?

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "django-oauth-toolkit"
 dynamic = ["version"]
-requires-python = ">=3.8,<=3.14"
+requires-python = ">=3.8,<3.15"
```
hugovk:
```suggestion
requires-python = ">=3.8"
```

hugovk (changes requested):
Hello, Python 3.14 and 3.15 release manager here! 👋 

I strongly recommend against an upper cap.

First, by disallowing installation on 3.15 (currently in alpha), you're preventing people testing their own code on 3.15. I would really like people to be able to test their code on 3.15 to find and report bugs in Python so we can fix them before the big release in October. Also to help prepare their own code for 3.15.

Not only those who depend on django-oauth-toolkit, but those who depend on dependencies of django-oauth-toolkit. And so on down the chain.

But more importantly, it's too late to add any top version here, you would have needed to do it from day one (or else need to yank or delete _all_ past releases!).

For example, if you release django-oauth-toolkit 3.2.0 with `<=3.15` and I try and install with Python 3.15, I'll get the last version available without the top pin, 3.1.0.

Fast forward a year or two. Maybe you have `<=3.17`, and have released django-oauth-toolkit 5.1.0 or something. I try and install with 3.17, and end up with 3.1.0 from way back in 2025, which is definitely not intended.

See https://iscinumpy.dev/post/bound-version-constraints/ for the long version.

pyproject.toml
n2ygk: Sounds good!

n2ygk:
@dopry over to you for your review of the suggestion not to set an upper bound on requirements. Testing versions will continue to need to be update as new python releases are publshed.

PR #1632, opened 2026-01-07
test: add failing test for redirect URI case sensitivity

webjunkie:
PR #1508 (wildcard support) changed redirect URI comparison to use `parsed_uri.hostname` instead of `netloc`. Since `.hostname` auto-lowercases, `https://EXAMPLE.COM` now matches `https://example.com`.

OAuth 2.0 Security BCP suggests exact string matching for redirect URIs - is this change intentional?

Adding a test that passes on 3.0.1 but fails on master to clarify expected behavior.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -721,10 +721,9 @@ def test_redirect_to_uri_allowed_expects_allowed_uri_list():
 def test_redirect_to_uri_allowed_case_sensitive():
     """
     OAuth 2.0 Security Best Current Practice requires exact string matching for redirect URIs.
-    See: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics
+    See: https://datatracker.ietf.org/doc/html/rfc9700#section-4.1.3
 
-    "When comparing client redirect URIs against pre-registered URIs, authorization
-    servers MUST utilize exact string matching."
+    "the authorization server MUST ensure that the two URIs are equal"
 
     This means https://EXAMPLE.COM should NOT match https://example.com
     """


webjunkie:
Thanks for the clarification. I may be missing an intended interpretation here, but I’m concerned the current behavior is no longer “exact string matching” for `redirect_uri`.

Multiple specs/guidance documents explicitly call for *exact* matching / *simple string comparison* for redirect URIs:

- RFC 9700 (OAuth 2.0 Security BCP): requires “exact string matching” for redirect URIs (except the loopback port case) and *intentionally* points to RFC 3986 §6.2.1 (Simple String Comparison), not to the normalization or equivalence rules (e.g., case-insensitive host comparison) in §6.2.2 or in RFC 7230.
  - https://www.rfc-editor.org/rfc/rfc9700.html
- OAuth 2.1 draft: requires rejecting requests where `redirect_uri` “doesn’t exactly match” a registered one (again with loopback port exception).  
  - https://datatracker.ietf.org/doc/draft-ietf-oauth-v2-1/
- OpenID Connect Core uses the same “MUST exactly match … matching performed as described in RFC 3986 §6.2.1” wording.  
  - https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

Separately, major providers document strictness in practice. For example, Microsoft Entra states redirect URIs are case-sensitive (at least for the URL path) and must match what’s configured:  
- https://learn.microsoft.com/en-us/entra/identity-platform/reply-url

If the maintainers prefer "semantic equivalence" for hostnames (case-insensitive host), could we at least:
1) document that redirect URI host matching is intentionally case-insensitive, and  
2) consider making this behavior opt-in (since it diverges from the strict "exact string matching" guidance above)?

PR #1659, opened 2026-02-27
Fix duplicate unique constraint for device_code (#1656)

### LINKED ISSUE #1656: DeviceGrant table definition breaks in Oracle 19c env
**Describe the bug**

migration 0013_alter_application_authorization_grant_type_device.py fails with error: 

`oracledb.exceptions.DatabaseError: ORA-02261: such unique or primary key already exists in the table
`
The cause seems to be that in the DeviceGrant, device_code is defined as:
 `               ('device_code', models.CharField(max_length=100, unique=True)),
`
and then there is a separate constraint defined on the same column:
`                'constraints': [models.UniqueConstraint(fields=('device_code',), name='oauth2_provider_devicegrant_unique_device_code')],
`
which causes the Oracle db to throw the above error.

( with PostgreSQL, the code works as expected, this seems to be specific to Oracle )

**To Reproduce**

env:
Python: 3.12
Django: 5.2
python-oracledb: 3.4.2
Oracle db version: 19c

run manage.py migrate

--> gives stack trace with above error

**Expected behavior**

the migration should be applied successfully ;)

**Version**

3.2.0

**Additional context**

febin-micheal:
Fixes #1656

## Description of the Change
This PR fixes issue #1656 by removing the second `UniqueConstraint` definition from `DeviceGrant` that causes `ORA-02261` on Oracle 19c deployments. Verified locally that this doesn't break PostgreSQL logic and added tests passing.
Also added febin-micheal to authors and #1656 to changelog.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Eduardo Oliveira
 Egor Poderiagin
 Emanuele Palazzetti
 Fazeel Ghafoor
+Febin Micheal Antony
 Federico Dolce
 Florian Demmer
 Frederico Vieira
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #1628 Fix inaccurate help_text on client_secret field of Application model
+* #1656 Fix duplicate unique constraint for device_code on Oracle 19c
 * #1674 Add `list_select_related` to `RefreshTokenAdmin` to avoid unbounded `JOIN` queries on the changelist
 * #1621 Fix device code tokens getting the wrong scope.
 * #1683 Fix swapped `DeviceGrant` model usage across the device authorization flow
--- a/oauth2_provider/migrations/0013_alter_application_authorization_grant_type_device.py
+++ b/oauth2_provider/migrations/0013_alter_application_authorization_grant_type_device.py
@@ -35,7 +35,6 @@ class Migration(migrations.Migration):
             options={
                 'abstract': False,
                 'swappable': 'OAUTH2_PROVIDER_DEVICE_GRANT_MODEL',
-                'constraints': [models.UniqueConstraint(fields=('device_code',), name='oauth2_provider_devicegrant_unique_device_code')],
             },
         ),
     ]
--- a/oauth2_provider/migrations/0015_remove_duplicate_devicegrant_constraint.py
+++ b/oauth2_provider/migrations/0015_remove_duplicate_devicegrant_constraint.py
@@ -1,0 +1,41 @@
+from django.db import migrations, models
+
+
+DUPLICATE_DEVICE_CODE_CONSTRAINT = "oauth2_provider_devicegrant_unique_device_code"
+
+
+def remove_duplicate_device_code_constraint(apps, schema_editor):
+    try:
+        DeviceGrant = apps.get_model("oauth2_provider", "DeviceGrant")
+    except LookupError:
+        return
+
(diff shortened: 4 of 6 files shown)


dopry:
@febin-micheal  this needs to be rebased. I'm noting a few issues

1. oauth2_provider/models.py:660 — The fix doesn't cover swapped/custom DeviceGrant models, including the repo's own test app. tests/models.py SampleDeviceGrant inherits Meta(AbstractDeviceGrant.Meta), but tests/migrations/0008_sampledevicegrant.py:53-58 still declares tests_sampledevicegrant_unique_device_code — model state and migration state now diverge, and makemigrations tests emits an unintended RemoveConstraint. No CI env catches this (the migrations-* tox envs' mig_settings.py doesn't install the tests app). Downstream projects with OAUTH2_PROVIDER_DEVICE_GRANT_MODEL swapped are exactly the same shape: their generated migrations still create the duplicate constraint (still ORA-02261 on fresh Oracle), and 0015 deliberately early-returns on swapped models. The PR needs a tests-app migration, and at minimum a CHANGELOG/docs note telling swapped-model users to regenerate their own migrations.

3. tests/test_device.py:1172 — test_device_code_unique_at_db_level will fail in the required py312-dj42-multi-db CI env . tests/db_router.py routes oauth2_provider writes to the alpha database, but transaction.atomic() opens its savepoint on default. The IntegrityError poisons the alpha connection (needs_rollback=True), and DeviceFlowBaseTestCase.tearDown's DeviceModel.objects.all().delete() then raises TransactionManagementError. Fix: with transaction.atomic(using=DeviceModel.objects.db):.

4. #1718 (https://github.com/django-oauth/django-oauth-toolkit/pull/1718) fixes the same issue the opposite way (keep the Meta constraint, drop the field-level unique=True). I'm not sure which approach  I prefer at the moment. 

5. oauth2_provider/migrations/0015:40 — RunPython isn't marked elidable=True. After the edited 0013, this operation is pure DB cleanup with no state effect — exactly what elidable is for. Without it, every fresh database (every new install, every CI test-DB creation) runs constraint introspection forever, and a future squashmigrations must carry the code permanently.

9. oauth2_provider/migrations/0015:8 — the try/except LookupError guard is dead code. The migration depends on 0014, so historical state at this point always contains DeviceGrant (created in 0013); the silent return would also mask a typo'd lookup as a successful no-op. Drop the guard.

10. tests/test_device.py:1181 — assert constraints == [] is over-broad. The regression is "duplicate uniqueness on device_code," but this forbids any future Meta constraint (e.g. a legitimate CheckConstraint on status), and its failure message would instruct the author to remove a correct change. Assert instead that no UniqueConstraint targets ("device_code",).

11. tests/test_device.py:1189, 1208 — the migration tests don't test the migration. test_migration_defines_only_field_level_uniqueness_for_device_code imports 0013 and asserts its source structure (a change-detector that breaks on any future squash), and the two TestDeviceGrantConstraintCleanupMigration tests are pure mock choreography restating the implementation — they'd stay green even if get_constraints/remove_constraint misbehaved on a real backend. One real-schema test (add the legacy constraint via schema_editor.add_constraint, run the function, introspect that it's gone; run it twice for idempotency) would replace all of this with actual coverage.

dopry:
@febin-micheal after some consideration I'm going to go with the newer PR that preserves the constraint. The contraint has a fixed name which is easier to grep and look for vs the field level which gets hashed. I suspect this would be easier for people who have implemented swappables.  I appreciate the PR and your work on the issue. I would appreciate it if you could test https://github.com/django-oauth/django-oauth-toolkit/pull/1718 and confirm that it fixes up your oracle situation.

PR #1664, opened 2026-03-02
fix: preserve '+' in client_secret during authentication.

awais786:
https://github.com/django-oauth/django-oauth-toolkit/issues/1639

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+### Fixed
+#1639 preserve '+' in client_secret during authentication
+
 ### Added
 * #1637 Support for Django 6.0
 


oauth2_provider/oauth2_validators.py
```diff
@@ -188,10 +192,24 @@ def _authenticate_request_body(self, request):
         clients unable to directly utilize the HTTP Basic authentication scheme.
         See rfc:`2.3.1` for more details.
         """
-        # TODO: check if oauthlib has already unquoted client_id and client_secret
         try:
             client_id = request.client_id
-            client_secret = getattr(request, "client_secret", "") or ""
+            # oauthlib uses parse_qsl to decode the request body, which converts
```
dopry, line 197: per RFC6749 2.3.1 
>    Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.  The client identifier is **encoded using the
   "application/x-www-form-urlencoded" encoding algorithm** per
   Appendix B, and the encoded value is used as the username; the **client
   password is encoded using the same algorithm** and used as the
   password.  The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.

In application/x-www-form-urlencoded data, the plus sign (+) is used to represent a space character. If you need to include a literal plus sign in your data, it must be encoded as %2B. It seems like the client isn't properly encoding the body of the request and unquote_plus is the proper behavior here.

tests/test_oauth2_validators.py
```diff
@@ -110,6 +110,72 @@ def test_authenticate_request_body(self):
...
+
+        # RFC-compliant: client encodes '+' as '%2B' in the body
+        self.request.body = "client_id=client_id&client_secret=secret%2Bwith%2Bplus"
+        self.assertTrue(
+            self.validator._authenticate_request_body(self.request),
+            "Authentication must succeed when '+' is sent as '%2B' in body",
+        )
+
+        # Lenient: client sends '+' literally in the body (common real-world case)
+        self.request.body = "client_id=client_id&client_secret=secret+with+plus"
```
dopry, line 135: This is malformed per the spec. The input needs to be encoded using the
   "application/x-www-form-urlencoded" encoding algorithm. This indicates a broken client.

dopry:
@awais786  I appreciate the effort. Unfortunately this is a case of a misbehaved client and not an DOT issue.

PR #1675, opened 2026-03-25
Avoid crawling the entire model graph when viewing `RefreshTokenAdmin`

### LINKED ISSUE #1674: `RefreshTokenAdmin` changelist triggers unbounded `select_related`, causing `Too many columns` on MySQL
**Describe the bug**

When viewing the `RefreshToken` page in the Django admin, Django generates a SQL query with an unbounded `SELECT` that `JOIN`s every related table reachable by following foreign keys from `RefreshToken`. On MySQL this eventually exceeds the server's maximum column limit and raises a `Too many columns` error.

The root cause is that `RefreshTokenAdmin` does not set `list_select_related`. Django admin's default behavior (`list_select_related = False`) detects FK fields in `list_display` (`user`, `application`) and falls back to calling `qs.select_related()` with no arguments, which recursively traverses the entire model graph.

**To Reproduce**

1. Create a Django project with a sufficiently complex model graph (e.g. multiple apps with FK relationships reachable from User or Application).
2. Log into the Django admin and navigate to the `RefreshToken` changelist (`/admin/oauth2_provider/refreshtoken/`).
3. Observe a `django.db.utils.OperationalError: (1117, 'Too many columns')` error from MySQL.

**Expected behavior**

- The Django admin page should load normally, with related fields and filters just like any other Django admin page.
- The changelist query should only `JOIN` the `user` and `application` tables (the two relations actually displayed in the list) rather than crawling the entire model graph.

**Version**
3.2.0

**Additional context**

In my case, this happened when `qs.select_related()` generated a query with over 4,096 columns. This may prove difficult or tedious to truly reproduce simply due to the scale required to trigger the problem (it only happens in my production environment). Additionally, I don't know enough about other databases to know if they have the same limit, but MySQL 8.0 fails there.

Despite reproduction issues, the solution is simply using the same `list_select_related` values as `AccessTokenAdmin` and `IDTokenAdmin`.

I'm putting together a PR from my fork.

IrishPrime:
Fixes #1674 

## Description of the Change

Explicitly set `list_select_related` to only look for related applications and users when viewing the `RefreshTokenAdmin` page. When this value is not set, highly connected model graphs (i.e. those where the `user` foreign-key is related to a large number of other models) can cause runaway recursion causing the admin page to fail to load.

Add a regression test that verifies the `RefreshTokenAdmin` changelist queryset uses bounded `select_related`. The test replicates the logic of `ChangeList.apply_select_related` and asserts that `qs.query.select_related` is a `dict` scoped to the declared fields rather than `True` (unbounded), which would cause Django to crawl the entire model graph.

## Checklist

- [x] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * #1628 Fix inaccurate help_text on client_secret field of Application model
+* #1674 Add `list_select_related` to `RefreshTokenAdmin` to avoid unbounded `JOIN` queries on the changelist
 
 ## [3.2.0] - 2025-11-13
 ### Added
--- a/oauth2_provider/admin.py
+++ b/oauth2_provider/admin.py
@@ -53,6 +53,7 @@ class IDTokenAdmin(admin.ModelAdmin):
 
 class RefreshTokenAdmin(admin.ModelAdmin):
     list_display = ("token", "user", "application")
+    list_select_related = ("application", "user")
     raw_id_fields = ("user", "access_token")
     search_fields = ("token",) + (("user__email",) if has_email else ())
     list_filter = ("application",)


IrishPrime:
You guys have saved me so much time and effort with this project that the least I could do was be thorough when trying to contribute back. :saluting_face:

PR #1682, opened 2026-04-01
Fix multi-db tests

### LINKED ISSUE #1681: Fix multi-db tests
Currently muiltidb tests are failing, we need to fix them and add them back as a success blocker.

joaquinhuigomez:
Fixes #1681

## Description of the Change

The multi-db test settings used a dummy `default` database backend (`"default": {}`), which caused Django's test infrastructure (TestCase flush/rollback) to crash. This gives `default` a real SQLite backend so the test runner can manage transactions on it, while the routers still direct all application queries to `alpha`/`beta`. Three test classes also needed `databases = "__all__"` so Django allows queries routed to non-default databases.

Adds a `test-multi-db` CI job as a required check in the success gate.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -229,9 +229,7 @@ jobs:
       - sphinxlint
       - migrations
       - migrate_swapped
-      # nees to fix multidb tests, not blocking success for now 
-      # so we can get the lint and migrations checks in place. 
-      # - multi-db
+      - multi-db
     runs-on: ubuntu-latest
     name: Test successful
     steps:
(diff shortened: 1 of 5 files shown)


dopry:
@joaquinhuigomez  Thank for this! I rebased it to incorporate some other CI changes I merged.  Do you have experience working with Django instances that use multiple databases?  Does an empty default make sense? Is it a typical setup? I was thinking multidb support would usually look something like default: primary, second: read-only replica  or reversed...  

I'm also looking at CrossDatabaseRouter.allow_relation and it seems like it should be 
```
return True  if obj1._state.db == obj2._state.db else None
```
could the routers becausing an issue here?

joaquinhuigomez:
Closing this — I don't have enough multi-db Django experience to confidently answer your question, and it's gone stale on my side. Happy to revisit if you'd like.

PR #1686, opened 2026-04-09
Use get_device_grant_model() on device-code runtime paths

### LINKED ISSUE #1683: Device-code flow still uses concrete DeviceGrant when OAUTH2_PROVIDER_DEVICE_GRANT_MODEL is swapped
**Describe the bug**
When `OAUTH2_PROVIDER_DEVICE_GRANT_MODEL` points to a swapped `DeviceGrant` model, django-oauth-toolkit 3.2 device-code support is only partially compatible with that swap. Some code paths correctly use `get_device_grant_model()`, but important runtime paths still use the concrete `oauth2_provider.DeviceGrant` class directly.

This breaks the device flow because creation, lookup, and exception handling can target the wrong model/table or the wrong `DoesNotExist` exception class.

**To Reproduce**
1. Define a custom model subclassing `oauth2_provider.models.AbstractDeviceGrant`.
2. Set `OAUTH2_PROVIDER_DEVICE_GRANT_MODEL` to that swapped model.
3. Enable and exercise the device authorization flow.
4. Hit the device authorization endpoint so DOT creates a device grant.
5. Poll the token endpoint with `grant_type=urn:ietf:params:oauth:grant-type:device_code`.
6. Optionally test the browser confirmation/status views using the generated user code.

Concrete problematic paths observed in DOT 3.2:
- `oauth2_provider.models.create_device_grant()` uses `DeviceGrant.objects.create(...)`
- `oauth2_provider.views.base.TokenView.device_flow_token_response()` uses `DeviceGrant.objects.get(...)`
- `oauth2_provider.views.device.DeviceGrantForm.clean_user_code()` catches `DeviceGrant.DoesNotExist` instead of the swapped model's exception
- `oauth2_provider.views.device.DeviceConfirmView` and `DeviceGrantStatusView` use the concrete `DeviceGrant` model in lookups/views

**Expected behavior**
All device-grant runtime paths should consistently use `get_device_grant_model()` so swapped models are fully supported anywhere DOT creates, loads, or catches exceptions for `DeviceGrant` instances.

**Version**
3.2.x

**Additional context**
I did a repo pass focused on `DeviceGrant` usages and found that some DOT code already handles swaps correctly, for example `oauth2_provider.utils.set_oauthlib_user_to_device_request_user()` and the `objects.get(...)` call in `DeviceGrantForm`. The bug appears to be incomplete migration to `get_device_grant_model()` rather than lack of swap support overall.

The likely fix is to replace concrete `DeviceGrant` usages in the affected runtime paths with `get_device_grant_model()` and add regression tests under swapped settings with a custom `SampleDeviceGrant`.

joaquinhuigomez:
`OAUTH2_PROVIDER_DEVICE_GRANT_MODEL` is documented as swappable, but several device-code runtime paths still referenced the concrete `oauth2_provider.DeviceGrant` class and its `DoesNotExist` attribute directly. That broke the device flow for projects pointing the setting at a swapped model, because creation, lookup, and exception handling targeted the wrong model/table.

This patch routes `create_device_grant`, `DeviceGrantForm.clean_user_code`, `DeviceConfirmView.get_object`, `DeviceGrantStatusView`, and the `device_flow_token_response` lookup through `get_device_grant_model()` so they honor the swap.

Fixes #1683

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -739,7 +739,7 @@ class DeviceCodeResponse:
 def create_device_grant(device_request: DeviceRequest, device_response: DeviceCodeResponse) -> DeviceGrant:
     now = datetime.now(tz=dt_timezone.utc)
 
-    return DeviceGrant.objects.create(
+    return get_device_grant_model().objects.create(
         client_id=device_request.client_id,
         device_code=device_response.device_code,
         user_code=device_response.user_code,
--- a/oauth2_provider/views/base.py
+++ b/oauth2_provider/views/base.py
@@ -15,7 +15,7 @@ from django.views.decorators.debug import sensitive_post_parameters
 from django.views.generic import FormView, View
 from oauthlib.oauth2.rfc8628 import errors as rfc8628_errors
 
-from oauth2_provider.models import DeviceGrant
+from oauth2_provider.models import get_device_grant_model
 
 from ..compat import login_not_required
 from ..exceptions import OAuthToolkitError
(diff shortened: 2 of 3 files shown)


dopry:
I merged https://github.com/django-oauth/django-oauth-toolkit/pull/1684 which addressed the same issue.

PR #1694, opened 2026-05-07
Add localhost as a valid loopback domain

leiserfg:
## Description of the Change

This adds localhost as a valid loopback domain, as
https://datatracker.ietf.org/doc/html/rfc8252#section-8.3
it should not be used but needs to be handled the same as 127.0.0.1
Some broken clients (aka Claude Code) use it :facepalm: 

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/oauth2_provider/models.py
+++ b/oauth2_provider/models.py
@@ -951,6 +951,7 @@ def redirect_to_uri_allowed(uri, allowed_uris):
         allowed_uri_is_loopback = parsed_allowed_uri.scheme == "http" and parsed_allowed_uri.hostname in [
             "127.0.0.1",
             "::1",
+            "localhost"
         ]
         """ check port """
         if not allowed_uri_is_loopback and parsed_allowed_uri.port != parsed_uri.port:


leiserfg:
Guys, sorry for rushing the PR, let me know if you really need the CHANGELOG and AUTHORS entries, I considered it too small of a thing to bother.

dopry:
We are not adding loopback as it is *NOT RECOMMENDED*

> While redirect URIs using localhost (i.e.,
   "http://localhost:{port}/{path}") function similarly to loopback IP
   redirects described in [Section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3), the use of localhost is NOT
   RECOMMENDED.  Specifying a redirect URI with the loopback IP literal
   rather than localhost avoids inadvertently listening on network

The client should be updated to address the non-conforming and potentially insecure behavior.

pencil:
I ran into this as well. I agree `localhost` is NOT RECOMMENDED per the RFC… But unfortunately "NOT RECOMMENDED" is advisory guidance to clients and it doesn't forbid them. Several widely-deployed native clients register `http://localhost/callback` and then receive the callback on an OS-assigned ephemeral port:
* Claude Code (`claude mcp add --transport http …`)
* MCP Inspector
* `mcp-remote`

"the client should be updated" isn't actionable for operators running DOT as their AS.

@dopry Would an opt-in setting (default `False`) be acceptable, similar to `ALLOW_URI_WILDCARDS`?

dopry:
I would be open to that. I have a pile of PRs lined up meant to help support mcp . If you could open a PR and cycle reviews with an agent till it gets stable I'd be happy to review it.

pencil:
I created #1739 with an `ALLOW_LOCALHOST_LOOPBACK` opt-in

dopry:
@leiserfg  I'm sorry about dismissing this out of hand earlier, I didn't really understand the context and impact. You did initially contribute this, so I want to ensure we keep you in authors... So you just want your GH handle there or is there a name we should add? closing in favor of #1739

PR #1696, opened 2026-05-20
Fix OIDC claims when never logged 

davidebianchi03:
## Description of the Change

This PR fixes an issue in the OpenID Connect ID token generation flow when `auth_time` is built for a user whose `last_login` value is `None`.

In the current implementation, `get_id_token_dictionary()` uses `request.user.last_login` to populate the `auth_time` claim. This works for users who have previously logged in, but it fails for users that exist in the system without a recorded login timestamp.

One example is a Django application using Django OAuth Toolkit as an OIDC provider together with an impersonation flow. An administrator may create a user from the Django admin interface and then impersonate that user, for example with `django-impersonate`, in order to verify permissions inside the client application. Since the impersonated user may never have performed a normal login, `last_login` can be `None`, causing ID token creation to fail.

This change adds a fallback for `auth_time` when `last_login` is not available. If the user has no recorded login timestamp, the current time is used instead. Existing behavior is preserved for users with a valid `last_login`.

## Why this is needed

OIDC login should not fail only because the user was created administratively or authenticated through an impersonation/debugging workflow and therefore has no `last_login` value.

The change keeps the ID token generation robust while preserving the existing claim structure.

## Checklist

- [ ] unit-test added
- [ ] documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 * #1594 Fix introspection token expiry handling to consistently use UTC and avoid the deprecated
   `datetime.utcfromtimestamp`.
+* #1696 Fix `auth_time` in oauth2 validator when user has never logged in.
+
 
 ## [3.3.0] - 2025-05-21
 
@@ -41,7 +43,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * #1683 Fix swapped `DeviceGrant` model usage across the device authorization flow
 * #1689 Fix invalid `Cache-Control` header value on the OIDC JWKS endpoint
 * #1692 Fix consent violation and scope escalation.
-* Fix `auth_time` in oauth2 validator when user has never logged in.
 
 ## [3.2.0] - 2025-11-13
 ### Added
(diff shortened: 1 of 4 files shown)


oauth2_provider/oauth2_validators.py
```diff
@@ -897,12 +897,17 @@ def get_id_token_dictionary(self, token, token_handler, request):
         claims = self.get_oidc_claims(token, token_handler, request)
 
         expiration_time = timezone.now() + timedelta(seconds=oauth2_settings.ID_TOKEN_EXPIRE_SECONDS)
+
+        last_login = request.user.last_login
```
dopry: Let's call this var auth_time instead of last_login, and it should be a utc time stamp per https://openid.net/specs/openid-connect-core-1_0.html#IDToken

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * #1683 Fix swapped `DeviceGrant` model usage across the device authorization flow
 * #1689 Fix invalid `Cache-Control` header value on the OIDC JWKS endpoint
 * #1692 Fix consent violation and scope escalation.
+* Fix `auth_time` in oauth2 validator when user has never logged in.
```
dopry: Prefix with the #PR reference

dopry (changes requested):

dopry (approved):

PR #1718, opened 2026-06-25
Remove redundant unique=True on DeviceGrant.device_code (duplicate unique index)

moscowmule2240:
## Problem

`AbstractDeviceGrant.device_code` declares **both** `unique=True` (field level) and a
`unique_device_code` `UniqueConstraint` in `Meta.constraints`. Each creates a unique index
on the same column, so `DeviceGrant` ends up with two identical unique indexes.

On MySQL, creating the `DeviceGrant` table (migration `0013`) emits `ER_DUP_INDEX` (1831):

> Duplicate index '...' defined on the table '...'. This is deprecated and will be disallowed in a future release.

Where database warnings are raised as errors this becomes a hard error and aborts `migrate`.
This happens, for example, with `mysql.connector`'s Django backend, which sets
`raise_on_warnings = settings.DEBUG`, so test-database creation under `DEBUG=True` fails.

## Fix

Remove the redundant field-level `unique=True` from `AbstractDeviceGrant.device_code` **and
from the `0013` `CreateModel`**, so the duplicate index is never created. Uniqueness is still
enforced by the explicit `unique_device_code` `UniqueConstraint`, which also gives the
swappable subclasses a stable, per-class name via `%(app_label)s_%(class)s_...`.

I edited migration `0013` in place rather than adding a follow-up migration on purpose: a new
`AlterField`/`RemoveIndex` migration would still let `0013` create the duplicate first and hit
`ER_DUP_INDEX` before the cleanup could run, so it would not fix fresh installs. The device
flow migration is recent (3.x); installs that already applied the old `0013` keep one harmless
extra index. Happy to switch to a separate migration or a squash if you prefer.

No behaviour change -- `device_code` remains unique.

## Verification

- `django-admin makemigrations --check --settings tests.mig_settings` -> *No changes detected*
- `pytest -k device` -> **27 passed**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   add nullable checksum → drop the old `("token", "revoked")` unique constraint → widen `token` to
   `TextField` → backfill → make checksum non-nullable → add the `("token_checksum", "revoked")`
   unique constraint.
+* If you use a swapped device grant model (`OAUTH2_PROVIDER_DEVICE_GRANT_MODEL`), run
+  `manage.py makemigrations` after upgrading: the redundant field-level `unique=True` was removed
+  from `AbstractDeviceGrant.device_code` (#1656), so your app needs a one-operation `AlterField`
+  migration that drops the extra unique index. Uniqueness remains enforced by the
+  `<app_label>_<class>_unique_device_code` constraint. If you are doing a *fresh* install on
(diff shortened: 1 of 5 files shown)


dopry:
@moscowmule2240 hey I'm planning on merging this in a bit. I'm going to work on a precursor PR to help catch migration drift. I should finish that up in the next day or so.  I'll reconcile the changelog if I get a chance, but if you have time I'd appreciate it.

dopry (approved):

Issue #1416, opened 2024-04-12
Add `localhost` to Allowed Loopback Addresses for Redirect URIs


Kanellaman:
### Problem Description
The library currently supports loopback redirect URIs with the `http` scheme for the addresses `127.0.0.1` and `::1` without specifying a port, aligning with the best practices for OAuth2 applications on native devices as per RFC8252. However, the hostname `localhost`, which is commonly used and interchangeable with `127.0.0.1` in most development environments, is not explicitly supported. This can lead to slight confusion as developers are required to use IP addresses directly, which may not always be ideal, particularly in environments where `localhost` is prevalently used.

### Proposed Change
I propose that we extend the validation logic to include `localhost` as a recognized hostname for loopback addresses. This change would allow developers to use `localhost` in redirect URIs without having to resort to numeric IP addresses, thus enhancing usability and consistency across different development setups.
[Current Validation Line](https://github.com/jazzband/django-oauth-toolkit/blob/560f84d9e20a499d10c29bce94efda7898c2939d/oauth2_provider/models.py#L781)

Change to:
```python
and parsed_allowed_uri.hostname in ["127.0.0.1", "::1", "localhost"]
```

### Rationale
Including `localhost` in the list of valid loopback addresses would enhance the developer experience and align the library with common development practices where `localhost` is preferred over direct IP addresses. It's a minor change that could make the library more intuitive and user-friendly without compromising security or functionality, as `localhost` is effectively synonymous with `127.0.0.1` and `::1` within the context of loopback addresses.

### Impact
This change would be backward-compatible, broadening the validation scope to include an additional, commonly used hostname. It will benefit developers by allowing more flexibility in redirect URI configurations without negatively impacting existing implementations.

dopry:
Implemented on `main` via the new **`ALLOW_LOCALHOST_LOOPBACK`** setting (#1739). When enabled, it extends the [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3) any-port loopback exemption to `http://localhost` redirect URIs, in addition to the existing `127.0.0.1` / `::1` support.

Note the implementation is **opt-in** (the setting defaults to `False`) rather than always treating `localhost` as loopback — `localhost` can resolve to a non-loopback address in some environments, so enabling it is left to the deployer. This will ship in the next release. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1739, opened 2026-07-07
feat: add ALLOW_LOCALHOST_LOOPBACK for RFC 8252 loopback redirect URIs

pencil:
## What this does

Adds an opt-in `ALLOW_LOCALHOST_LOOPBACK` setting (default `False`) that extends the RFC 8252 §7.3 any-port loopback exemption to `http://localhost`. From the discussion in #1416.

§7.3 wants the auth server to accept any port on a loopback `redirect_uri` so a native app can bind an OS-assigned port. DOT applies that to the IP literals `127.0.0.1` / `[::1]` only. §8.3 says `localhost` "functions similarly" but is NOT RECOMMENDED, so `localhost` gets strict port matching today, which is a fine default. This just makes it opt-in-able.

## Where we hit it

You added DCR in #1728. We run the same DCR, and on top of it CIMD (Client ID Metadata Documents), which DOT doesn't have yet.

CIMD fixes the problem of abandoned DCR clients piling up: instead of a stored client record per registration that nobody ever cleans up, the `client_id` is just an HTTPS URL and the server reads the client metadata (including `redirect_uris`) out of the document it points at, on demand. Every install of a client shares the one URL, so there's nothing to accumulate.

That's also where this bites. The doc is published ahead of time, so it can't know the ephemeral port the client will bind at runtime. It just lists `http://localhost/callback`, no port. Then the client (Claude Code, VS Code) grabs an OS-assigned port and calls back on `http://localhost:49152/callback`. The ports don't match, `localhost` isn't one of the exempted literals, so consent gets rejected and the flow dies.

Plain DCR doesn't hit this, the client registers at request time so it can hand over whatever redirect URI it's actually using. CIMD can't, the doc is static.

## Notes

- Default `False`, so nothing changes unless you turn it on. The §8.3 "NOT RECOMMENDED" default stays for everyone else.
- Mirrors `ALLOW_URI_WILDCARDS`: same shape, a permissive relaxation of redirect-URI matching, read inline in the matcher, documented with a security note.
- No cross-spelling. `localhost` still has to match exactly, so it never gets conflated with `127.0.0.1` / `[::1]` (tested both ways). `http`-only, path and query matching unchanged.

## tests/app/idp and tests/app/rp

Didn't touch these. They register fixed redirect URIs and run full flows, and this only shows up with an ephemeral loopback port those harnesses don't bind. It's covered by unit tests on `redirect_to_uri_allowed`, same place `ALLOW_URI_WILDCARDS` is tested. Happy to add a demo if you'd rather.

## Checklist

- [x] PR only contains one change
- [x] unit-test added
- [x] documentation updated
- [x] `CHANGELOG.md` updated
- [ ] tests/app/idp updated to demonstrate new features (N/A, see above)
- [ ] tests/app/rp updated to demonstrate new features (N/A, see above)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -101,6 +101,7 @@ Matej Spiller Muys
 Matias Seniquiel
 Michael Howitz
 Michael "Irish" O'Neill
+Nils Caspar
 Owen Gong
 Patrick Palacin
 Paul Dekkers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `DynamicClientRegistrationManagementView` with configurable permission classes and registration access
   tokens. Dynamically registered applications are flagged with a new `AbstractApplication.dcr_created`
   field and can be filtered in the Django admin.
+* #1739 `ALLOW_LOCALHOST_LOOPBACK` setting to extend the RFC 8252 §7.3 any-port loopback exemption to `http://localhost` redirect URIs (opt-in, default `False`)
 
 ### Security
 * Generate device-flow `user_code` values with the cryptographically secure `secrets` module
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -83,6 +83,29 @@ For example,
 This feature is useful for working with CI service such as cloudflare, netlify, and vercel that offer branch
 deployments for development previews and user acceptance testing.
 
+ALLOW_LOCALHOST_LOOPBACK
+~~~~~~~~~~~~~~~~~~~~~~~~~
+Default: ``False``
+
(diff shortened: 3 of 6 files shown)


dopry:
@pencil can you also share step by step how to reproduce the error with the misbehaving client that requires this?

pencil:
@dopry I updated the description. Initially I thought this happens with DCR in general but doing more testing I think this primarily affects CIMD (newer approach for how MCP clients register). DOT does not officially support CIMD yet though it can be built with the primitives exposed by it.

dopry:
@pencil cool. I'll fast track this one in. If you want to put together a CIMD PR that would be great.

dopry (approved):

pencil:
@dopry Yeah, I filed #1742 with a rough architecture proposal. If that looks good to you, I can send a PR.

PR #1799, opened 2026-07-27
fix(dcr): honour hashed token storage for registration access tokens

pencil:
## What

RFC 7592 registration access tokens are issued outside the validator's normal token path, so they do not honour `COMPLIANT_BCP_RFC9700_TOKEN_STORAGE`. `_issue_registration_token()` assigns the value straight onto the model:

```python
token = AccessToken.objects.create(
    ...
    token=generate_client_secret(),
)
```

Every other token DOT issues goes through `OAuth2Validator._set_token_value()`, which blanks the `token` column and keeps only the lookup checksum when the setting is enabled. This one does not, so a deployment that has explicitly opted into hashed-at-rest storage still has its registration access tokens persisted in cleartext — and nothing surfaces the divergence.

This is a latent gap rather than a live vulnerability: the setting defaults to `False`, and with it off `_set_token_value()` does exactly what the current code does. It only bites deployments that turned hashing on and reasonably assumed it applied to every token in the table.

## The fix

The redaction logic moves out of the validator and into `oauth2_provider.models.set_token_value()`, sitting next to `TokenChecksumField` where the `_raw_token` protocol it drives already lives. `OAuth2Validator._set_token_value()` stays put as a thin delegating wrapper, so a custom validator that overrides it keeps working for every token the validator issues; the DCR view calls the model helper directly rather than reaching into validator internals.

That alone would break the feature, which is why the diff is larger than the one-line cause suggests. Once the column is redacted, `registration_token.token` is `""`, and three response paths read it:

- **POST** `/register/` — returns the token it just minted.
- **GET** `/register/{client_id}/` — RFC 7592 §3 includes the registration access token in the read response. The server no longer holds a readable copy, so it echoes back the token the request presented (it was just validated, so it is necessarily the right one).
- **PUT** `/register/{client_id}/` — returns the rotated value when `DCR_ROTATE_REGISTRATION_TOKEN_ON_UPDATE` is on.

So `_issue_registration_token()` now returns the raw token string instead of the model instance, and `_application_to_response()` takes that raw string.

The management endpoints need the presented token too, which means the auth helper has to hand back three things instead of one. Rather than widen its tuple — a `tuple[App, Token, str] | tuple[None, HttpResponse, None]` cannot be narrowed by a type checker, since testing the first element for `None` says nothing about the third — `_get_application_from_registration_token()` becomes `_authenticate_registration_request()`, returning either a frozen `_AuthenticatedRegistration` dataclass or the `HttpResponse` to send. A single `isinstance(auth, HttpResponse)` check narrows all three values at once, so no type-narrowing asserts are needed and an `AssertionError` cannot surface as a 500 on an auth path if a later refactor breaks the invariant. The error returns stay single-value, so nothing in the helper reformats. mypy is clean on this file.

Uniqueness is unaffected: `AccessToken.token` is a plain `TextField` and the unique constraint lives on `token_checksum`, which is derived from the raw value.

## Tests

New `TestDCRRegistrationTokenStorage` in `tests/test_dcr_views.py`, covering the full lifecycle with the setting on: the stored column is blank while the checksum matches, the token returned at registration still authenticates against the management endpoint and is echoed back by `GET`, and a rotated `PUT` token is both returned and usable while also being redacted at rest.

Two of the three fail on `master` for the right reason (the raw token found sitting in the column); the third is a regression guard on the response threading.

`tests/test_bcp_rfc9700.py` gains direct coverage of `set_token_value()` in both modes, matching the existing coverage of the validator method it was extracted from.

Full suite passes locally: `1074 passed, 1 skipped`.

## Docs

`docs/security.rst` already stated that enabling the setting means "a database read no longer discloses usable tokens" — the code simply did not match for this token. Added a short note that registration access tokens are covered, and that `GET` echoes the presented value as a consequence.

Note that `set_token_value()` is public, so it now appears on the Models page via `automodule`.

## Notes

Found while auditing 3.4.0 before adopting it.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -188,6 +188,11 @@ By default DOT stores access and refresh tokens in cleartext (alongside a SHA-25
 token hash, so a database read no longer discloses usable tokens. Existing cleartext
 tokens are left in place and age out as they expire or rotate.
 
+This covers the RFC 7592 registration access token as well. Since the server keeps no
+readable copy, that token is returned in full only by the registration response that
+mints it; the management endpoint's ``GET`` echoes back whichever token the request
+presented.
+
 .. warning::
    Hashed token storage is incompatible with the refresh-token grace period, which
    must return a previously issued (cleartext) token from the database. When
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -189,9 +189,10 @@ token hash, so a database read no longer discloses usable tokens. Existing clear
 tokens are left in place and age out as they expire or rotate.
 
 This covers the RFC 7592 registration access token as well. Since the server keeps no
-readable copy, that token is returned in full only by the registration response that
-mints it; the management endpoint's ``GET`` echoes back whichever token the request
-presented.
+readable copy, a client only receives that token from a response that mints one: the
+registration response, and — when ``DCR_ROTATE_REGISTRATION_TOKEN_ON_UPDATE`` is
+enabled — the management ``PUT`` that rotates it. A management ``GET``, or a ``PUT``
+without rotation, echoes back whichever token the request presented.
 
 .. warning::
    Hashed token storage is incompatible with the refresh-token grace period, which
--- a/oauth2_provider/views/dynamic_client_registration.py
+++ b/oauth2_provider/views/dynamic_client_registration.py
@@ -7,6 +7,7 @@ RFC 7592 — GET/PUT/DELETE /register/{client_id}/
 
 import hashlib
 import json
+from dataclasses import dataclass
 from datetime import datetime, timedelta
 from datetime import timezone as dt_timezone
 from typing import Any
@@ -346,6 +347,21 @@ class DynamicClientRegistrationView(View):
         return JsonResponse(response_data, status=201)
 
 
+@dataclass(frozen=True)
+class _AuthenticatedRegistration:
+    """A management request that presented a valid registration access token.
+
+    ``presented_token`` is the raw token from the ``Authorization`` header. It is
+    carried alongside the row because under ``COMPLIANT_BCP_RFC9700_TOKEN_STORAGE``
+    the ``token`` column is blank, so the request itself is the only remaining
+    source of the value the RFC 7592 read response has to echo back.
+    """
+
+    application: AbstractApplication
+    registration_token: AbstractAccessToken
+    presented_token: str
+
+
 @method_decorator(csrf_exempt, name="dispatch")
 @method_decorator(login_not_required, name="dispatch")
 class DynamicClientRegistrationManagementView(View):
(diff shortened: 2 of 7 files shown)


oauth2_provider/views/dynamic_client_registration.py
```diff
@@ -218,26 +223,36 @@ def _issue_registration_token(application, user):
...
         user=user,
-        token=generate_client_secret(),
         expires=expires,
         scope=oauth2_settings.DCR_REGISTRATION_SCOPE,
     )
-    return token
+    # Assigning ``token`` directly would store a reusable credential in cleartext even
+    # when the deployment asked for hashed-at-rest storage; the validator is what
+    # applies the redaction consistently for every other token this library issues.
+    oauth2_settings.OAUTH2_VALIDATOR_CLASS()._set_token_value(token, raw_token)
```
dopry: _set_token_value() is a private method of OAuth2Validator. We shouldn't be calling it from outside of OAuth2Validator or a class that extends it. We should consider expanding the interface or adding a helper that can be used in _set_token_value.  IIRC, the OAUTH2_VALIDATOR_CLASS() is documented as extending oauthlib.oauth2.RequestValidator, so being able to extend and type the interface may mean updating the docs or establishing a new base class here.  what do you think is a good solution?

oauth2_provider/views/dynamic_client_registration.py
```diff
@@ -218,26 +223,36 @@ def _issue_registration_token(application, user):
...
+    token.save()
+    return token, raw_token
```
dopry: the returned token doesn't seem to be used anywhere. Is there a reason for returning it?  It would be nice to see this streamlined.

docs/security.rst
```diff
@@ -180,6 +180,11 @@ By default DOT stores access and refresh tokens in cleartext (alongside a SHA-25
 token hash, so a database read no longer discloses usable tokens. Existing cleartext
 tokens are left in place and age out as they expire or rotate.
 
+This covers the RFC 7592 registration access token as well. Since the server keeps no
+readable copy, that token is returned in full only by the registration response that
```
dopry: the token is also returned on the PUT rotation...

tests/test_dcr_views.py
```diff
@@ -951,3 +952,82 @@ def test_register_get_put_delete(self):
...
+        response = self.client.get(self.management_url, **_bearer(self.registration_token))
+        assert response.status_code == 200
+
+    def test_get_returns_the_presented_token(self):
+        """RFC 7592 §3 puts the registration access token in the read response, and the
+        server no longer holds a readable copy, so it echoes what the client presented."""
+        response = self.client.get(self.management_url, **_bearer(self.registration_token))
+        assert response.json()["registration_access_token"] == self.registration_token
+
+    def test_rotated_token_is_returned_and_usable(self):
```
dopry: We should also verify the revoked token no longer works. could just be an assertion in this test and updated name to avoid all the setup

oauth2_provider/views/dynamic_client_registration.py
```diff
@@ -430,14 +461,13 @@ def put(self, request, client_id, *args, **kwargs):
 
             if oauth2_settings.DCR_ROTATE_REGISTRATION_TOKEN_ON_UPDATE:
                 user = application.user
-                new_token = _issue_registration_token(application, user)
+                _, raw_token = _issue_registration_token(application, user)
```
dopry: I'm not sure how I feel about raw_token here shadowing the raw_token on line 437.  It feels ambiguous... 
maybe something like 
```python 
application, registration_token, presented_token = (
    self._get_application_from_registration_token(request, client_id)
)
response_token = presented_token

# ...

if oauth2_settings.DCR_ROTATE_REGISTRATION_TOKEN_ON_UPDATE:
    _, response_token = _issue_registration_token(application, application.user)
    registration_token.delete()

return JsonResponse(_application_to_response(application, response_token, request))
```
would be more self documenting?

dopry (changes requested):
This is a solid first pass... Aside from the inline items there is one other small items I'd like to see updated.

 Can you type the methods and functions whose signatures we're touching? I want to start working toward having typing in place, but doing it incrementally so it's not a massive lift.

dopry:
@pencil do you think you'll be able to take a pass on this before next week? I'm planning to cut 3.4.1 next week.

pencil:
Yeah, fair, the view shouldn't be reaching into the validator.

Maybe just pull it into a `set_token_value()` helper in `models.py`, next to `TokenChecksumField` where the `_raw_token` protocol already lives? Then `_set_token_value` delegates to it and the view calls the helper. Nothing new on the interface that way, so no base class or oauthlib docs to sort out.

Addressed and pushed the rest of your findings.

dopry:
@pencil I'm down with the helper and the proposed placement.  could you rebase when you implement that?

PR #1804, opened 2026-07-30
fix(oidc): make RP-initiated logout idempotent for orphaned id_token_hint

### LINKED ISSUE #1287: Error if multiples RP ask to logout the same user
**Describe the bug**

A user is logged with OIDC on 2 separate RPs.
If he is redirected to the rp-initiated logout view for RP1, all his tokens well be revoked (and deleted for id/access tokens)
If RP2 also redirects the used to the rp-initiated logout using the `id_token_hint` parameter, the request will be rejected because no `IDToken` will be found (since it was previously deleted).

**To Reproduce**

See above

**Expected behavior**

The specs have :
> If any of the validation procedures defined in this specification fail, any operations requiring the information that failed to correctly validate MUST be aborted and the information that failed to validate MUST NOT be used. Note that because RP-Initiated Logout Requests are intended to be idempotent, it is explicitly not an error for an RP to request that a logout be performed when the OP does not consider that the End-User is logged in with the OP at the requesting RP.

> As described in [Section 3](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RedirectionAfterLogout), when the OP detects errors in the RP-Initiated Logout request, the OP MUST not perform post-logout redirection to an RP. Beyond that, the OP has discretion on what information to display to the End-User in the resulting page at the OP and what actions to enable the End-User to perform next. It MAY display an error message. It MAY ask the End-User whether to log out of the OP.

> Note that giving the End-User the opportunity to log out may have security benefits, especially in kiosk scenarios. The End-User initiating a logout action at the RP may expect to be completely logged out, including from the OP. Not giving the End-User the opportunity to log out at the OP and leaving the End-User logged in would likely violate the End-User's security expectations about being completely logged out.

We should not raise an error if the user is already logged out.
- if we have a session then we can find the user and proceed with the logout
- if not we don't have to do anything (no identified user, no session, no token) so we don't have anything to do (see similar case in #1280)

**Version**

2.3.0

**Additional context**

ArockiaRajamanickam:
Fixes #1287.

## The problem

`_load_id_token()` handled `IDToken.DoesNotExist` in the same `except` as `JWException` and `JWTExpired`, so all four cases returned `(None, None)` and `validate_logout_request_user()` raised `InvalidIDTokenError` for all of them. But a verified token whose IDToken row is gone is not the same thing as a token that could not be verified.

That case is reached on any normal multi-RP logout. When the first RP logs the End-User out, `do_logout()` deletes their ID Tokens, so the second RP's `id_token_hint` now points at an IDToken that no longer exists. The second RP gets HTTP 400 and never receives its `post_logout_redirect_uri`.

The spec quoted in the issue is explicit that this must not be an error:

> Note that because RP-Initiated Logout Requests are intended to be idempotent, it is explicitly not an error for an RP to request that a logout be performed when the OP does not consider that the End-User is logged in with the OP at the requesting RP.

## The change

`_load_id_token()` now splits verification from the row lookup and returns `(None, claims)` when the token verified but its IDToken is gone, so callers can distinguish the two. `validate_logout_request_user()` keys the error on `claims` rather than on `id_token`, and returns both.

Nothing about verification is relaxed. A token that fails to deserialize, has no resolvable key, fails signature verification or is expired still returns `(None, None)` and is still rejected. `_validate_claims()` still enforces `iss`.

Two things needed care so that dropping the row does not lose a check:

- **`client_id` matching** previously used `id_token.application.client_id`. With no row, the RP is resolved from the verified `aud` claim via the existing `_get_client_by_audience()`, and failing to resolve it counts as a mismatch. So `client_id` is still validated rather than skipped.
- **`post_logout_redirect_uri`** is validated against the requesting Application. `get_request_application()` takes the claims so it can recover that Application the same way, instead of falling through to `None` and rejecting a legitimate RP's redirect.

`token_user` stays `None` in this case, so `must_prompt()` behaves exactly as it does for a request with no `id_token_hint`: an End-User who still has an OP session is prompted rather than silently logged out.

## Behaviour change

`test_rp_initiated_logout_get_revoked_id_token` asserted the old 400, and now asserts 200. I'd rather flag that plainly than bury it: `IDToken.revoke()` is `self.delete()`, so that test was already exercising exactly this case, and the assertion encoded the behaviour the issue reports as wrong. Callers who relied on a 400 for an orphaned hint will now see the prompt-or-logout path instead.

## Tests

`test_rp_initiated_logout_second_rp_after_logout` is new and reproduces the reported scenario end to end: log out via one RP, then present the now-orphaned hint from a second RP and assert it gets its redirect. On `main` it fails with `assert 400 == 302`; the updated revoked-token test fails with `assert 400 == 200`. Both pass with the change.

Full suite is 1069 passed, 1 skipped. `ruff check` and `ruff format --check` are clean, and the RP-Initiated Logout section of `docs/oidc.rst` now documents the idempotency behaviour and what is still enforced.

I used an AI assistant while working on this. The reproduction, the two failing-without-the-fix runs and the reasoning about which checks had to be preserved are mine.

### DIFF
--- a/docs/oidc.rst
+++ b/docs/oidc.rst
@@ -170,10 +170,12 @@ This feature has to be enabled separately as it is an extension to the core stan
 
 Logout requests are idempotent, as the specification requires. An ``id_token_hint`` that verifies but
 whose ID Token is no longer stored is therefore not treated as an error: this is the case once another
-RP has already logged the same End-User out, since that deletes their ID Tokens. Such a request is
-handled as though no ``id_token_hint`` had been supplied, so the End-User is prompted if they still
-have a session with the OP. An ``id_token_hint`` that cannot be verified is still rejected, and a
-``client_id`` given alongside it is still required to match the RP the ID Token was issued for.
+RP has already logged the same End-User out, since that deletes their ID Tokens. The End-User is
+prompted if they still have a session with the OP, just as they are for a request carrying no
+``id_token_hint`` at all. The requesting RP is still identified, from the ID Token's ``aud`` claim, so
+a ``post_logout_redirect_uri`` is still validated against it. An ``id_token_hint`` that cannot be
+verified is still rejected, and a ``client_id`` given alongside it is still required to match the RP
+the ID Token was issued for.
 
 
 Setting up OIDC enabled clients
--- a/docs/oidc.rst
+++ b/docs/oidc.rst
@@ -168,6 +168,13 @@ This feature has to be enabled separately as it is an extension to the core stan
        # ... any other settings you want
    }
 
+Logout requests are idempotent, as the specification requires. An ``id_token_hint`` that verifies but
+whose ID Token is no longer stored is therefore not treated as an error: this is the case once another
+RP has already logged the same End-User out, since that deletes their ID Tokens. Such a request is
+handled as though no ``id_token_hint`` had been supplied, so the End-User is prompted if they still
+have a session with the OP. An ``id_token_hint`` that cannot be verified is still rejected, and a
+``client_id`` given alongside it is still required to match the RP the ID Token was issued for.
+
 
 Setting up OIDC enabled clients
 ===============================
--- a/docs/oidc.rst
+++ b/docs/oidc.rst
@@ -170,7 +170,8 @@ This feature has to be enabled separately as it is an extension to the core stan
 
 Logout requests are idempotent, as the specification requires. An ``id_token_hint`` that verifies but
 whose ID Token is no longer stored is therefore not treated as an error: this is the case once another
-RP has already logged the same End-User out, since that deletes their ID Tokens. The End-User is
+RP has already logged the same End-User out, since that deletes their ID Tokens, and equally for an ID
+Token that was deliberately revoked or that this OP never issued. The End-User is
 prompted if they still have a session with the OP, just as they are for a request carrying no
 ``id_token_hint`` at all. The requesting RP is still identified, from the ID Token's ``aud`` claim, so
 a ``post_logout_redirect_uri`` is still validated against it. An ``id_token_hint`` that cannot be
--- a/oauth2_provider/views/oidc.py
+++ b/oauth2_provider/views/oidc.py
@@ -189,16 +189,34 @@ def _load_id_token(token):
             check_claims = None
         jwt_token = jwt.JWT(key=key, jwt=token, check_claims=check_claims)
         claims = json.loads(jwt_token.claims)
-
-        # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
-        # same user.
-        # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
-        # attribute on the IDToken Object later on.
-
-        return IDToken.objects.get(jti=claims["jti"]), claims
-
-    except (JWException, JWTExpired, IDToken.DoesNotExist):
+    except (JWException, JWTExpired):
+        # The token could not be verified.
         return None, None
+
+    # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
+    # same user.
+    # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
+    # attribute on the IDToken Object later on.
+
+    try:
+        return IDToken.objects.get(jti=claims["jti"]), claims
+    except IDToken.DoesNotExist:
+        # The token was verified but is no longer stored, which means the End-User is not logged in with
+        # the OP at the requesting RP. This happens once another RP has logged the same user out, as
+        # `do_logout()` deletes their ID Tokens. The verified claims are still returned so that callers
+        # can tell this apart from a token that could not be verified at all.
+        return None, claims
+
+
+def _get_application_from_claims(claims):
+    """
+    Resolves the Application an ID Token was issued for from its `aud` claim.
+
+    This is used when the corresponding IDToken is no longer stored. `_load_id_token()` has already
+    verified the token's signature with this Application's key, so `aud` can be relied upon.
+    """
+    validator = oauth2_settings.OAUTH2_VALIDATOR_CLASS()
+    return validator._get_client_by_audience(claims.get("aud", []))
 
 
 def _validate_claims(request, claims):
@@ -324,29 +342,42 @@ class RPInitiatedLogoutView(OIDCLogoutOnlyMixin, FormView):
     def validate_logout_request_user(self, id_token_hint, client_id):
         """
         Validate the an OIDC RP-Initiated Logout Request user
+
+        `(id_token, claims)` is returned. `claims` are the verified claims of `id_token_hint`, if one was
+        given. `id_token` is the stored IDToken that `id_token_hint` refers to; it is `None` when that
+        IDToken is no longer stored, meaning the End-User is not logged in with the OP at that RP.
         """
 
         if not id_token_hint:
-            return
+            return None, None
 
         # Only basic validation has been done on the IDToken at this point.
         id_token, claims = _load_id_token(id_token_hint)
 
-        if not id_token or not _validate_claims(self.request, claims):
+        if not claims or not _validate_claims(self.request, claims):
             raise InvalidIDTokenError()
 
         # If both id_token_hint and client_id are given it must be verified that they match.
         if client_id:
-            if id_token.application.client_id != client_id:
+            # When the IDToken is no longer stored, the requesting RP is recovered from the verified
+            # `aud` claim so that this check is still enforced. Not being able to resolve it counts as
+            # a mismatch.
+            application = id_token.application if id_token else _get_application_from_claims(claims)
+            if application is None or application.client_id != client_id:
                 raise ClientIdMissmatch()
 
-        return id_token
+        return id_token, claims
 
-    def get_request_application(self, id_token, client_id):
+    def get_request_application(self, id_token, client_id, claims=None):
         if client_id:
             return get_application_model().objects.get(client_id=client_id)
         if id_token:
             return id_token.application
+        if claims:
+            # `id_token_hint` was verified but its IDToken is no longer stored. The requesting RP is
+            # recovered from the verified `aud` claim so that `post_logout_redirect_uri` is still
+            # validated against the Application that asked for the logout.
+            return _get_application_from_claims(claims)
 
     def validate_logout_request(self, id_token_hint, client_id, post_logout_redirect_uri):
         """
@@ -360,8 +391,8 @@ class RPInitiatedLogoutView(OIDCLogoutOnlyMixin, FormView):
         will be validated against each other.
         """
 
-        id_token = self.validate_logout_request_user(id_token_hint, client_id)
-        application = self.get_request_application(id_token, client_id)
+        id_token, claims = self.validate_logout_request_user(id_token_hint, client_id)
+        application = self.get_request_application(id_token, client_id, claims)
         self.validate_post_logout_redirect_uri(application, post_logout_redirect_uri)
 
         return application, id_token.user if id_token else None
--- a/oauth2_provider/views/oidc.py
+++ b/oauth2_provider/views/oidc.py
@@ -197,16 +197,34 @@ def _load_id_token(token):
             check_claims = None
         jwt_token = jwt.JWT(key=key, jwt=token, check_claims=check_claims)
         claims = json.loads(jwt_token.claims)
-
-        # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
-        # same user.
-        # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
-        # attribute on the IDToken Object later on.
-
-        return IDToken.objects.get(jti=claims["jti"]), claims
-
-    except (JWException, JWTExpired, IDToken.DoesNotExist):
+    except (JWException, JWTExpired):
+        # The token could not be verified.
         return None, None
+
+    # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
+    # same user.
+    # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
+    # attribute on the IDToken Object later on.
+
+    try:
+        return IDToken.objects.get(jti=claims["jti"]), claims
+    except IDToken.DoesNotExist:
+        # The token was verified but is no longer stored, which means the End-User is not logged in with
+        # the OP at the requesting RP. This happens once another RP has logged the same user out, as
+        # `do_logout()` deletes their ID Tokens. The verified claims are still returned so that callers
+        # can tell this apart from a token that could not be verified at all.
+        return None, claims
+
+
+def _get_application_from_claims(claims):
+    """
+    Resolves the Application an ID Token was issued for from its `aud` claim.
+
+    This is used when the corresponding IDToken is no longer stored. `_load_id_token()` has already
+    verified the token's signature with this Application's key, so `aud` can be relied upon.
+    """
+    validator = oauth2_settings.OAUTH2_VALIDATOR_CLASS()
+    return validator._get_client_by_audience(claims.get("aud", []))
 
 
 def _validate_claims(request, claims):
@@ -332,29 +350,42 @@ class RPInitiatedLogoutView(OIDCLogoutOnlyMixin, FormView):
     def validate_logout_request_user(self, id_token_hint, client_id):
         """
         Validate the an OIDC RP-Initiated Logout Request user
+
+        `(id_token, claims)` is returned. `claims` are the verified claims of `id_token_hint`, if one was
+        given. `id_token` is the stored IDToken that `id_token_hint` refers to; it is `None` when that
+        IDToken is no longer stored, meaning the End-User is not logged in with the OP at that RP.
         """
 
         if not id_token_hint:
-            return
+            return None, None
 
         # Only basic validation has been done on the IDToken at this point.
         id_token, claims = _load_id_token(id_token_hint)
 
-        if not id_token or not _validate_claims(self.request, claims):
+        if not claims or not _validate_claims(self.request, claims):
             raise InvalidIDTokenError()
 
         # If both id_token_hint and client_id are given it must be verified that they match.
         if client_id:
-            if id_token.application.client_id != client_id:
+            # When the IDToken is no longer stored, the requesting RP is recovered from the verified
+            # `aud` claim so that this check is still enforced. Not being able to resolve it counts as
+            # a mismatch.
+            application = id_token.application if id_token else _get_application_from_claims(claims)
+            if application is None or application.client_id != client_id:
                 raise ClientIdMissmatch()
 
-        return id_token
+        return id_token, claims
 
-    def get_request_application(self, id_token, client_id):
+    def get_request_application(self, id_token, client_id, claims=None):
         if client_id:
             return get_application_model().objects.get(client_id=client_id)
         if id_token:
             return id_token.application
+        if claims:
+            # `id_token_hint` was verified but its IDToken is no longer stored. The requesting RP is
+            # recovered from the verified `aud` claim so that `post_logout_redirect_uri` is still
+            # validated against the Application that asked for the logout.
+            return _get_application_from_claims(claims)
 
     def validate_logout_request(self, id_token_hint, client_id, post_logout_redirect_uri):
         """
@@ -368,8 +399,8 @@ class RPInitiatedLogoutView(OIDCLogoutOnlyMixin, FormView):
         will be validated against each other.
         """
 
-        id_token = self.validate_logout_request_user(id_token_hint, client_id)
-        application = self.get_request_application(id_token, client_id)
+        id_token, claims = self.validate_logout_request_user(id_token_hint, client_id)
+        application = self.get_request_application(id_token, client_id, claims)
         self.validate_post_logout_redirect_uri(application, post_logout_redirect_uri)
 
         return application, id_token.user if id_token else None
--- a/oauth2_provider/views/oidc.py
+++ b/oauth2_provider/views/oidc.py
@@ -237,10 +237,18 @@ def _get_application_from_claims(claims: dict) -> Application | None:
     """
     Resolves the Application an ID Token was issued for from its `aud` claim.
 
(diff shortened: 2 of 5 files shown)


oauth2_provider/views/oidc.py
```diff
@@ -189,16 +189,34 @@ def _load_id_token(token):
...
+    # attribute on the IDToken Object later on.
 
+    try:
         return IDToken.objects.get(jti=claims["jti"]), claims
+    except IDToken.DoesNotExist:
+        # The token was verified but is no longer stored, which means the End-User is not logged in with
+        # the OP at the requesting RP. This happens once another RP has logged the same user out, as
+        # `do_logout()` deletes their ID Tokens. The verified claims are still returned so that callers
+        # can tell this apart from a token that could not be verified at all.
+        return None, claims
```
dopry, line 233: This introduces a third return state — `(None, claims)` for "verified but no longer stored" — but the docstring above (line 161-166) still documents only two:

> A tuple (IDToken, claims) is returned. ... If loading failed (None, None) is returned.

The whole contract of this PR is that callers must now distinguish `(None, None)` from `(None, claims)`, and every caller has to get that distinction right. Leaving it implicit in the docstring is the wrong place to leave it. Something like:

> `(IDToken, claims)` is returned when the token verifies and is still stored. `(None, claims)` is returned when it verifies but its IDToken is no longer stored, meaning the End-User is not logged in with the OP at the requesting RP. `(None, None)` is returned when the token could not be verified at all.

---

oauth2_provider/views/oidc.py
```diff
@@ -189,16 +189,34 @@ def _load_id_token(token):
...
+        return None, None
 
-        # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
-        # same user.
-        # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
-        # attribute on the IDToken Object later on.
+    # Assumption: the `sub` claim and `user` property of the corresponding IDToken Object point to the
+    # same user.
+    # To verify that the IDToken was intended for the user it is therefore sufficient to check the `user`
+    # attribute on the IDToken Object later on.
```
dopry: Moving this block out of the `try` is right, but it now sits above a path where there *is* no IDToken object — so "it is therefore sufficient to check the `user` attribute on the IDToken Object later on" no longer describes both outcomes below it.

Worth noting explicitly here that **not** deriving `token_user` from the `sub` claim in the orphaned case is deliberate, because it's the load-bearing security property of this PR and it's invisible to a reader. With `token_user` left `None`, `must_prompt()` returns `True` whenever the End-User still has an OP session, which is exactly what [RP-Initiated Logout §2](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) mandates:

> the OP MUST ask the End-User this question ... if the supplied ID Token does not belong to the current OP session with the RP

Resolving the user from `sub` would skip the prompt without any evidence of a session, which would turn a leaked stale ID Token into a silent logout. A future reader "tidying this up" needs to know that's a trap.

---

oauth2_provider/views/oidc.py
```diff
@@ -189,16 +189,34 @@ def _load_id_token(token):
...
-        return None, None
+
+def _get_application_from_claims(claims):
+    """
+    Resolves the Application an ID Token was issued for from its `aud` claim.
+
+    This is used when the corresponding IDToken is no longer stored. `_load_id_token()` has already
+    verified the token's signature with this Application's key, so `aud` can be relied upon.
+    """
+    validator = oauth2_settings.OAUTH2_VALIDATOR_CLASS()
```
dopry: The helper is sound, and I want to record *why* since the docstring only asserts it: this resolves through the same `_get_client_by_audience()` that `_get_key_for_token()` (`oauth2_provider/oauth2_validators.py:1472`) already used to fetch the verification key. So it necessarily returns the same `Application` whose key verified the signature — `aud` really can be relied on here. A sentence saying that ("the same helper that selected the verification key, so this is the Application the signature was verified against") would save the next reader the trip through the validator.

Two smaller things:

- **Type annotations.** `AGENTS.md` asks for annotations on any signature you're already modifying, and this one is brand new. Note that the return is the *swappable* application model, so `Application` is not quite right as an annotation — `get_application_model()` is what callers see. `AbstractApplication | None` is probably the honest one.
- `claims.get("aud", [])` — the `[]` default is unreachable in practice, since `_get_key_for_token()` returns `None` when `aud` is absent and `_load_id_token()` bails at line 177. Harmless, just don't expect coverage on it.

---

oauth2_provider/views/oidc.py
```diff
@@ -324,29 +342,42 @@ def validate_post_logout_redirect_uri(self, application, post_logout_redirect_ur
...
 
         # If both id_token_hint and client_id are given it must be verified that they match.
         if client_id:
-            if id_token.application.client_id != client_id:
+            # When the IDToken is no longer stored, the requesting RP is recovered from the verified
+            # `aud` claim so that this check is still enforced. Not being able to resolve it counts as
+            # a mismatch.
+            application = id_token.application if id_token else _get_application_from_claims(claims)
+            if application is None or application.client_id != client_id:
                 raise ClientIdMissmatch()
```
dopry, line 407: **The `else` arm here is the point of the change, and nothing in the suite executes it.**

Because it's a ternary, coverage.py marks line 365 as covered as soon as *any* request with a `client_id` reaches it — but `_get_application_from_claims(claims)` only runs when `id_token` is `None`, and no existing test combines a revoked/orphaned `id_token_hint` with a `client_id` (I grepped: zero). So the reported coverage is green over an untested branch.

I ran both outcomes and they behave correctly — matching `client_id` gives 200, mismatched gives 400 — so this is about locking the behaviour in, not a defect:

```python
@pytest.mark.django_db(databases="__all__")
def test_rp_initiated_logout_orphaned_id_token_matching_client_id(logged_in_client, oidc_tokens, rp_settings):
    validator = oauth2_settings.OAUTH2_VALIDATOR_CLASS()
    validator._load_id_token(oidc_tokens.id_token).revoke()
    rsp = logged_in_client.get(
        reverse("oauth2_provider:rp-initiated-logout"),
        data={"id_token_hint": oidc_tokens.id_token, "client_id": oidc_tokens.application.client_id},
    )
    assert rsp.status_code == 200


@pytest.mark.django_db(databases="__all__")
def test_rp_initiated_logout_orphaned_id_token_missmatch_client_id(logged_in_client, oidc_tokens, rp_settings):
    validator = oauth2_settings.OAUTH2_VALIDATOR_CLASS()
    validator._load_id_token(oidc_tokens.id_token).revoke()
    rsp = logged_in_client.get(
        reverse("oauth2_provider:rp-initiated-logout"),
        data={"id_token_hint": oidc_tokens.id_token, "client_id": "not-the-right-one"},
    )
    assert rsp.status_code == 400
```

The first fails on `master` with `assert 400 == 200` and passes here; the second passes on both, which is the point — it pins that the `client_id` check was preserved rather than skipped.

Separately, `application is None` on line 366 is effectively unreachable for the same reason as the `aud` default above: verification already required that Application to exist. Fine as defensive code.

---

oauth2_provider/views/oidc.py
```diff
@@ -324,29 +342,42 @@ def validate_post_logout_redirect_uri(self, application, post_logout_redirect_ur
     def validate_logout_request_user(self, id_token_hint, client_id):
         """
         Validate the an OIDC RP-Initiated Logout Request user
+
+        `(id_token, claims)` is returned. `claims` are the verified claims of `id_token_hint`, if one was
+        given. `id_token` is the stored IDToken that `id_token_hint` refers to; it is `None` when that
+        IDToken is no longer stored, meaning the End-User is not logged in with the OP at that RP.
```
dopry: Both public method contracts on this view change here, and neither is flagged for downstream users:

- `validate_logout_request_user()` returned `IDToken | None`, now returns a 2-tuple.
- `get_request_application()` gained a third positional parameter, and is called positionally at line 395.

`RPInitiatedLogoutView` is documented as subclass-friendly (`docs/settings.rst:599` advertises overriding it), so a downstream override of either method breaks on upgrade — `get_request_application()` with a `TypeError`, `validate_logout_request_user()` with a silent tuple-unpacking bug that's much nastier to trace. That belongs in `CHANGELOG.md` under `### Changed`, and arguably in `docs/upgrade.rst` alongside the other override-breaking changes.

Adding type annotations to these two signatures while you're here would also satisfy `AGENTS.md` and make the new tuple contract self-documenting.

---

tests/test_oidc_views.py
```diff
@@ -743,15 +743,46 @@ def test_rp_initiated_logout_get_id_token(logged_in_client, oidc_tokens, rp_sett
...
+    assert rsp.status_code == 302
+    assert not is_logged_in(logged_in_client)
+
+    # The second RP now presents its own, now orphaned, id_token_hint.
+    rsp = logged_in_client.get(
+        reverse("oauth2_provider:rp-initiated-logout"),
+        data={"id_token_hint": id_token_hint, "post_logout_redirect_uri": "http://example.org"},
+    )
+    assert rsp.status_code == 302
+    assert rsp["Location"] == "http://example.org"
```
dopry: **This test doesn't test a second RP.** The docstring says "Both RPs hold an ID Token for the same End-User", but both requests pass `oidc_tokens.id_token` — the same token, issued to the same Application. The setup the docstring describes is never built.

It also passes partly for an uninteresting reason: after the first logout the user is no longer authenticated, so `must_prompt()` short-circuits at line 412 (`if not self.request.user.is_authenticated: return False`) before the orphaned-token handling carries much weight.

`oidc_non_confidential_tokens` (`tests/conftest.py:291`) already gives you a genuinely distinct application with its own ID Token. I ran this version — it passes on this branch and fails on `master` with `assert 400 == 302`, i.e. it actually reproduces #1287:

```python
@pytest.mark.django_db(databases="__all__")
def test_rp_initiated_logout_second_rp_after_logout(
    logged_in_client, oidc_tokens, oidc_non_confidential_tokens, rp_settings
):
    assert oidc_tokens.application != oidc_non_confidential_tokens.application

    # First RP logs the End-User out. This deletes all of their ID Tokens.
    rsp = logged_in_client.get(
        reverse("oauth2_provider:rp-initiated-logout"),
        data={"id_token_hint": oidc_tokens.id_token, "post_logout_redirect_uri": "http://example.org"},
    )
    assert rsp.status_code == 302
    assert not is_logged_in(logged_in_client)

    # The second RP's own id_token_hint is now orphaned by the first RP's logout.
    rsp = logged_in_client.get(
        reverse("oauth2_provider:rp-initiated-logout"),
        data={
            "id_token_hint": oidc_non_confidential_tokens.id_token,
            "post_logout_redirect_uri": "http://other.org",
        },
    )
    assert rsp.status_code == 302
    assert rsp["Location"] == "http://other.org"
```

Using a different `post_logout_redirect_uri` for the second RP is deliberate — it proves the redirect was validated against the *second* Application, recovered via `aud`, rather than passing by coincidence.

Minor: there's a stray blank line before the closing `"""` on line 766.

---

docs/oidc.rst
```diff
@@ -168,6 +168,13 @@ This feature has to be enabled separately as it is an extension to the core stan
        # ... any other settings you want
    }
 
+Logout requests are idempotent, as the specification requires. An ``id_token_hint`` that verifies but
+whose ID Token is no longer stored is therefore not treated as an error: this is the case once another
+RP has already logged the same End-User out, since that deletes their ID Tokens. Such a request is
+handled as though no ``id_token_hint`` had been supplied, so the End-User is prompted if they still
+have a session with the OP. An ``id_token_hint`` that cannot be verified is still rejected, and a
+``client_id`` given alongside it is still required to match the RP the ID Token was issued for.
```
dopry: Good addition — the idempotency behaviour genuinely needed documenting.

One correction: "handled as though no ``id_token_hint`` had been supplied" is stronger than what the code does. The RP is still identified from the verified `aud` claim, so `post_logout_redirect_uri` is validated against that Application. A genuinely hint-less request with a `post_logout_redirect_uri` and no `client_id` would hit `InvalidOIDCClientError` at `views/oidc.py:326` instead. The two paths differ in exactly the way an operator reading this would care about.

Suggest narrowing the claim to the prompting behaviour, which is the part that really is identical:

```suggestion
Logout requests are idempotent, as the specification requires. An ``id_token_hint`` that verifies but
whose ID Token is no longer stored is therefore not treated as an error: this is the case once another
RP has already logged the same End-User out, since that deletes their ID Tokens. The End-User is
prompted if they still have a session with the OP, just as they are for a request carrying no
``id_token_hint``. The requesting RP is still identified from the ID Token's ``aud`` claim, so a
``post_logout_redirect_uri`` is still validated against it. An ``id_token_hint`` that cannot be
verified is still rejected, and a ``client_id`` given alongside it is still required to match the RP
the ID Token was issued for.
```

---

dopry (commented):
## Summary

The fix is correct, well-targeted and spec-justified, and it closes a real bug — I reproduced the failure against `master` before reviewing the diff. What's missing is release hygiene and a test that actually builds the scenario its own docstring describes.

### What I verified locally

Checked out `9317b7a`, installed `.[dev]`, ran with `DJANGO_SETTINGS_MODULE=tests.settings`:

| Check | Result |
|---|---|
| Full suite on PR head | 1098 passed, 1 skipped |
| `ruff check` / `ruff format --check` | clean (185 files) |
| Two **genuinely distinct** RPs (the #1287 scenario) | **fails on `master` with `assert 400 == 302`, passes here** |
| Orphaned hint + matching `client_id` | fails on `master` with `assert 400 == 200`, passes here |
| Orphaned hint + **wrong** `client_id` | 400 on both — check preserved, no regression |

### Design and spec assessment

The root-cause analysis holds. `do_logout()` deletes *all* of the user's access tokens and their `IDToken` rows across every application, so the moment one RP logs the End-User out, every other RP's `id_token_hint` is orphaned. Rejecting those is a spec violation — RP-Initiated Logout §2 is explicit:

> because RP-Initiated Logout Requests are intended to be idempotent, it is explicitly not an error for an RP to request that a logout be performed when the OP does not consider that the End-User is logged in with the OP at the requesting RP.

The security posture is preserved, and for a non-obvious reason worth stating: in the orphaned path `token_user` stays `None`, so `must_prompt()` returns `True` whenever the End-User still has an OP session — exactly what §2 mandates. A stale leaked ID Token therefore buys an attacker nothing beyond what a hint-less logout request already allows. Deliberately *not* resolving `token_user` from the `sub` claim is the right call; I've left an inline note asking for that to be written down, since it's invisible to a future reader.

`_get_application_from_claims()` is sound: it resolves through the same `_get_client_by_audience()` that `_get_key_for_token()` already used to select the verification key, so it necessarily returns the Application the signature was verified against.

### Should address before merge

Inline comments cover the code-level items. Two checklist items from the PR template aren't done:

1. **`CHANGELOG.md` is not updated.** This is a user-relevant change on two counts: the bug fix itself (`### Fixed`), and a behaviour change — an `id_token_hint` whose ID Token was revoked used to return 400 and now returns 200/302 (`### Changed`). Operators who relied on the strict behaviour need to see that. There's an `## [unreleased]` section ready for it.
2. **Author is not in `AUTHORS`.**

Also flagged inline: the new second-RP test doesn't use a second RP; the `client_id`-with-orphaned-token branch has no test; `_load_id_token()`'s docstring doesn't describe the new third return state; the two changed public method signatures break downstream overrides without a note; and missing type annotations on the touched signatures per `AGENTS.md`.

### Nits

- The `Application` is now resolved up to three times per request across three separate queries (once in `_get_key_for_token()`, then again in `validate_logout_request_user()` and/or `get_request_application()`). Not hot-path enough to block, but threading the already-resolved Application through would be cleaner.
- Thanks for calling out the `test_rp_initiated_logout_get_revoked_id_token` assertion flip in the description rather than burying it. `IDToken.revoke()` is `self.delete()`, so that test was already exercising this exact case and the old assertion encoded the behaviour #1287 reports as wrong. Agreed with the change; it just needs to reach the CHANGELOG.

---

dopry:
@ArockiaRajamanickam  I'd like to get this merged sooner rather than later. I'm opening up work on 3.5 and I'm hoping to release it sooner rather than later. I'd appreciate it if you could turn around the updates quickly. If you don't have time let me know and I'll get it wrapped.

oauth2_provider/views/oidc.py
dopry: Addressed in `47b7911`. The docstring now spells out all three outcomes and says why the distinction matters to callers:

```
A tuple `(IDToken, claims)` is returned, with three possible outcomes:

- `(IDToken, claims)` when the token verified and its IDToken is still stored.
- `(None, claims)` when the token verified but its IDToken is no longer stored, which means the
  End-User is not logged in with the OP at the requesting RP.
- `(None, None)` when the token could not be verified at all.

Callers must distinguish the last two: the second is not an error, because RP-Initiated Logout
requests are idempotent, while the third must be rejected.
```

I also annotated the signature — `def _load_id_token(token: str) -> tuple[AbstractIDToken | None, dict | None]:` — since this PR changes the function's contract, so `AGENTS.md` applies to it as much as to the methods below.

Resolving.

---

oauth2_provider/views/oidc.py
dopry: Addressed in `47b7911`. The existing sentence is now qualified with "when there is one", and the reasoning is written down directly above the branch it protects:

```
# When the IDToken is gone there is no such object to check, and the user is deliberately *not*
# resolved from the `sub` claim instead. Leaving it unresolved keeps `token_user` as `None`, which
# makes `must_prompt()` prompt an End-User who still has an OP session, as the specification
# requires when the ID Token does not belong to the current OP session with the RP. Resolving the
# user from `sub` here would skip that prompt without any evidence of a session, and so would turn
# a leaked stale ID Token into a silent logout.
```

Resolving.

---

oauth2_provider/views/oidc.py
dopry: Addressed in `47b7911`. The docstring now carries the reasoning rather than just asserting the conclusion:

```
This is used when the corresponding IDToken is no longer stored. `aud` can be relied upon because
this resolves through the same `_get_client_by_audience()` helper that `_get_key_for_token()`
already used to select the key the token's signature was then verified with. The Application
returned here is therefore the one that signature was verified against.
```

**On the annotation, my suggestion was wrong and I went the other way.** I said `Application` "is not quite right ... `AbstractApplication | None` is probably the honest one" — but `views/oidc.py:45` already binds `Application = get_application_model()` at module level, so within this module `Application` *is* the swapped concrete class, not the built-in one. `-> Application | None` is therefore both accurate and consistent with the rest of the file, and it needs no extra import. That's what's committed. (`AbstractIDToken` genuinely did need importing, and is.)

The `claims.get("aud", [])` default is left as-is, as unreachable-but-harmless.

Resolving.

---

oauth2_provider/views/oidc.py
dopry: Addressed in `47b7911` — both tests added, as `test_rp_initiated_logout_orphaned_id_token_matching_client_id` and `test_rp_initiated_logout_orphaned_id_token_missmatch_client_id`. I added `assert is_logged_in(logged_in_client)` to each as well, since in both cases the End-User's OP session must survive: the first is a prompt, not a logout, and the second is a rejection.

Re-verified against the rebased base (`e621e47`) rather than trusting the earlier run:

- matching `client_id` — fails on master with `assert 400 == 200`, passes here.
- mismatched `client_id` — 400 on both, which is the point of including it.

So the `else` arm now has execution behind it in both directions, not just a green coverage line.

`application is None` on the line below is still unreachable and still left as defensive code.

Resolving.

---

oauth2_provider/views/oidc.py
dopry: Addressed in `47b7911`. Both signatures are annotated, both carry a note aimed at subclassers, and there's a `### Changed` entry:

```
* #1287 `RPInitiatedLogoutView.validate_logout_request_user()` now returns an `(id_token, claims)`
  tuple rather than the `IDToken` alone, and `RPInitiatedLogoutView.get_request_application()` takes a
  third `claims` argument. Subclasses overriding either method must be updated.
```

`get_request_application()` also gained a docstring, since it had none and the new `claims` argument is the non-obvious part.

**Correcting myself on `docs/upgrade.rst`:** I suggested this "arguably" belongs there too. That's wrong, and following it would have put work in the wrong place. Checking the file's history, its per-release sections are written by the release commit — `Upgrading to 3.4.1` was added by `chore: 3.4.1 release (#1820)`, not by any of the PRs it describes. So `CHANGELOG.md` is the correct and only home for this at PR time, and consolidating it into `upgrade.rst` is the release manager's call when 3.5 goes out. Sorry for the detour.

Resolving.

---

tests/test_oidc_views.py
dopry: Addressed in `47b7911`. The test now takes `oidc_non_confidential_tokens` alongside `oidc_tokens`, asserts the two applications really are distinct, and sends the second RP to its own registered `http://other.org` — so the redirect passing proves it was validated against the second Application recovered from `aud`, rather than against the first. Stray blank line in the docstring is gone.

Worth recording how I checked it, because my first attempt at this was invalid: I ran the new test file against the unfixed base by stashing my changes first, which silently ran the *old* test file and produced a meaningless pass. Redone properly by copying the new test file onto `e621e47`, the four RP-logout tests split exactly as they should:

| Test | On base `e621e47` | On `47b7911` |
|---|---|---|
| `second_rp_after_logout` | fails, `assert 400 == 302` | passes |
| `get_revoked_id_token` | fails, `assert 400 == 200` | passes |
| `orphaned_id_token_matching_client_id` | fails, `assert 400 == 200` | passes |
| `orphaned_id_token_missmatch_client_id` | passes | passes |

The last one passing on both is deliberate — it's the guard that the `client_id` check wasn't dropped along with the row, so it should be green either side of the fix.

Full suite on the rebased head: 1350 passed, 1 skipped. `ruff check` and `ruff format --check` clean.

Resolving.

---

docs/oidc.rst
dopry: Addressed in `47b7911` — suggestion applied, with one wording tweak ("The requesting RP is still identified, from the ID Token's `aud` claim") to keep the sentence readable.

The Read the Docs build on the new head is green, and the section renders as intended: https://django-oauth-toolkit--1804.org.readthedocs.build/en/1804/oidc.html

Resolving.

---

PR #1811, opened 2026-08-20
fix(models): associate Application validation errors with their fields

### LINKED ISSUE #1343: Error messages in Application form are not associated with their fields.
Model/Application.clean is used to validate the Application model. When Errors are raised from Model.clean() There are non-specific and not associated with a field which can make it hard to understand what is really going on. 

image

is because the origin is not allowed in a field below...  This is very non-specific.

We should look at using [Model.full_clean](https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model.full_clean) and moving field specific validation to field validators.

dopry:
Fixes #1343

## Description of the Change

`Application.clean()` raised every failure as a non-field `ValidationError`, so the application forms (the registration/edit views and the Django admin) showed messages such as *"allowed origin URI Validation error. invalid_scheme: http://example.com"* at the top of the form with no indication of which input was at fault — the exact complaint in #1343 — and each save reported only the first problem found.

`clean()` now collects its errors keyed by the field they belong to and raises them together:

| problem | field |
| --- | --- |
| rejected redirect URI, or empty `redirect_uris` for a grant type that requires one | `redirect_uris` |
| non-https / otherwise rejected CORS origin | `allowed_origins` |
| RS256 without `OIDC_RSA_PRIVATE_KEY`; HS256 with a public grant or client | `algorithm` |
| HS256 with no client secret, or with a stored secret that is already hashed | `client_secret` |
| HS256 while `hash_client_secret` is still set | `hash_client_secret` |

Consequences:

* The shipped `application_form.html` already renders `field.errors`, so each message now appears next to the offending input, and a single submit reports every problem instead of one per round trip.
* `ValidationError.message_dict` is keyed by those field names, so callers of `Application.full_clean()` — dynamic client registration, CIMD, `createapplication` — now surface the field name alongside the message. The message texts themselves are unchanged.
* Django raises `ValueError` when model validation names a field a form omits, which a custom form with a narrower `Meta.fields` can trigger (e.g. omitting `redirect_uris` while keeping `authorization_grant_type`). `ApplicationForm.add_error()` re-keys those to non-field errors — the pre-existing behavior — so such forms keep working.

The remaining piece of #1343 was moving this validation off non-field errors; the HS256/hashed-secret live form warning noted in the issue thread is already in place.

## Checklist

- [x] unit-test added
- [x] documentation updated
- [ ] tests/app/idp updated to demonstrate new features — n/a, this changes the built-in application form/admin behavior
- [ ] tests/app/rp updated to demonstrate new features — n/a

Tests: `tests/test_models.py` gains coverage that each error names its field, that unrelated problems are collected into one `message_dict`, and that every invalid URI in a multi-line field gets its own message; `tests/test_application_views.py` gains a `TestApplicationFormFieldErrors` class asserting the errors land on the form fields (and render inside the field's control group), plus the narrower-form fallback to non-field errors. Full suite, ruff, sphinx-lint and the docs build pass locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_017m1dpRYGScha2uf7badQyz)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   non-standard and breaks interoperability with spec-compliant clients. It is scheduled for
   removal in 4.0.
 ### Changed
+* #1343 `Application.clean()` now reports its validation errors per field instead of as
+  non-field errors, and reports all of them at once instead of stopping at the first
+  problem. The application forms (the built-in registration/edit views and the Django
+  admin) render each message next to the offending input — a rejected redirect URI on
+  `redirect_uris`, a non-https CORS origin on `allowed_origins`, an unusable algorithm on
+  `algorithm`, and the HS256 client-secret conflicts on `client_secret` /
+  `hash_client_secret`. `ValidationError.message_dict` is keyed by those field names, so
+  callers of `Application.full_clean()` (including dynamic client registration and CIMD)
+  now surface the field name alongside the message. A custom `ModelForm` that omits one of
+  those fields still gets the message as a non-field error, provided it subclasses
+  `oauth2_provider.forms.ApplicationForm`.
 * The `AccessToken` and `RefreshToken` admins now invalidate tokens through a **"Revoke selected"**
   action instead of raw delete (delete is disabled on those two admins). A raw delete of an access
   token left its bound refresh token behind (`RefreshToken.access_token` is `SET_NULL`) — an orphan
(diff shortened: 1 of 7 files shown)


dopry:
CI on this PR is red for a reason outside the diff: pytest-django **4.14.0** renamed the `settings` fixture's `SettingsWrapper.finalize()` to `Settings._finalize()`, so `tests/conftest.py` raises `AttributeError` at teardown of every test using the `oauth2_settings` fixture. It reproduces identically on `master` (725 errors there, 736 here — the 11 extra are this PR's new tests, which use the same fixture), and the single `test_checks_fail_when_router_crosses_databases` failure in the multi-DB cell is a knock-on of the settings overrides never being disabled.

Fix is in #1813. Once that lands I'll merge `master` in here so CI re-runs against it.

For the record, with that fix applied locally this branch is fully green: `1079 passed, 1 skipped` on pytest-django 4.12.0 and 4.14.0 alike.

---

PR #1830, opened 2026-08-21
fix: time-bound the introspection POST and correct its logging

### LINKED ISSUE #1827: bug: resource-server introspection POST has no timeout, so a stalled remote AS can pin workers indefinitely
## Summary

`OAuth2Validator._get_token_from_authentication_server()` calls the remote RFC 7662 introspection endpoint with no timeout:

https://github.com/django-oauth/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L620

```python
try:
    response = requests.post(introspection_url, data=body, headers=headers)
except requests.exceptions.RequestException:
    return None
```

`requests` has no default timeout, so this blocks until the OS gives up. Every request carrying a bearer token goes through this path when `RESOURCE_SERVER_INTROSPECTION_URL` is set, so an authorization server that accepts the connection and then stalls will hold one worker per inbound request until the pool is exhausted — an availability failure in the resource server caused entirely by a third party.

## Why this stands out

It is the only outbound HTTP call in the library that is not time-bounded. Everything else added since 3.4 is:

| call site | bound |
|---|---|
| CIMD document fetch | `CIMD_FETCH_TIMEOUT_SECONDS` (default 5) |
| client-assertion JWKS fetch | `CLIENT_ASSERTION_JWKS_FETCH_TIMEOUT_SECONDS` (default 5) |
| `core/safe_fetch.py` | a single deadline budget shared across every DNS-resolved address, specifically so N addresses cannot hold a worker for N × timeout |

The introspection client predates that work and never got the same treatment.

## Suggested fix

Add a `RESOURCE_SERVER_INTROSPECTION_TIMEOUT_SECONDS` setting (default `5`, matching its siblings) alongside the existing `RESOURCE_SERVER_INTROSPECTION_{URL,CREDENTIALS,JWT_*}` family, and pass it through:

```python
response = requests.post(
    introspection_url,
    data=body,
    headers=headers,
    timeout=oauth2_settings.RESOURCE_SERVER_INTROSPECTION_TIMEOUT_SECONDS,
)
```

A connect/read tuple would be finer-grained, but a single total is consistent with the rest of the codebase and simpler to document.

Note that `safe_fetch` is not the right tool here even though it is stricter: the introspection URL is operator-configured rather than attacker-supplied, so SSRF hardening is not the concern, and `safe_fetch` is `GET`-only. Only the timeout gap needs closing.

Requires a `docs/settings.rst` entry and a test that a slow endpoint raises `requests.exceptions.Timeout` and is handled by the existing `RequestException` branch (returning `None`, i.e. token invalid).

## Provenance

Raised by the code-quality reviewer on #1765, which only relocates this code to `oauth2_provider/resource_server/validators.py`. Filed separately so that PR stays behavior-neutral.

### LINKED ISSUE #1828: bug: log.exception() misused in two resource-server paths (+ implicit return in the introspection client)
## Summary

Two resource-server call sites use `log.exception()` where it does not fit, plus one clarity fix in the same function. All are one-line changes, tracked together because a single PR touches all three.

### 1. Non-200 introspection response — not in an exception handler

https://github.com/django-oauth/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L627

```python
# Log an exception when response from auth server is not successful
if response.status_code != http.client.OK:
    log.exception(
        "Introspection: Failed to get a valid response "
        "from authentication server. Status code: {}, "
        "Reason: {}.".format(response.status_code, response.reason)
    )
    return None
```

This branch is reached on a perfectly ordinary HTTP response, not from an `except` block. `log.exception` sets `exc_info=True`, so with no active exception the handler appends a useless `NoneType: None` line to every one of these records — and, worse, if the caller happens to be inside an unrelated `except` further up the stack, it will attach *that* exception's traceback and make the log actively misleading.

Should be `log.warning`, and while touching the line, `%`-style deferred formatting rather than eager `.format()`:

```python
log.warning(
    "Introspection: Failed to get a valid response from authentication server. "
    "Status code: %s, Reason: %s.",
    response.status_code,
    response.reason,
)
```

### 2. `AccessToken.DoesNotExist` in `OAuth2ExtraTokenMiddleware` — expected condition logged as an error

https://github.com/django-oauth/django-oauth-toolkit/blob/master/oauth2_provider/middleware.py#L63

```python
except AccessToken.DoesNotExist as e:
    log.exception(e)
```

Here `exc_info` is legitimate — it *is* an exception handler — but the level is wrong. A bearer token that does not resolve is the normal outcome for an expired, revoked, or simply bogus token, and this middleware runs on every request carrying an `Authorization` header. The result is a full traceback per unauthenticated request: noisy logs, real formatting overhead on a hot path, and a trivially cheap way for an unauthenticated caller to inflate log volume.

`log.debug("Bearer token not found in the database")` is enough. Note also that `log.exception(e)` logs only the exception's `str()` as the message, which for `DoesNotExist` is an empty string — so today these records carry no message at all, just a traceback.

### 3. Implicit fall-through return in `_get_token_from_authentication_server()`

Same function as (1). Its last statement is the `if` at

https://github.com/django-oauth/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L640

```python
if "active" in content and content["active"] is True:
    ...
    return access_token
# <- falls off the end, returning None implicitly
```

so an introspection response with `"active": false` (or no `active` key at all) returns `None` by falling off the end, while every other failure path in the function returns `None` explicitly.

**The behavior is correct** — the caller does `if access_token and access_token.is_valid(scopes)`, so `None` means "token invalid", which is the right fail-closed outcome for an inactive token. But in a security decision this central, the most common rejection path should not be the one that is invisible. Add an explicit terminal return:

```python
    return access_token

log.debug("Introspection: token is not active")
return None
```

Purely a readability change, no behavior difference. Included here because it is inside the function (1) already modifies.

## Scope

Behavior is unchanged throughout: every path keeps its existing control flow. Worth a brief changelog note for (1) and (2), since operators filtering these records at `ERROR` level would stop seeing them.

## Provenance

(1) and (2) were raised by Copilot on #1765, (3) by the code-quality reviewer on the same PR. That PR only relocates this code to `oauth2_provider/resource_server/{validators,middleware}.py`, so all three were dismissed there to keep it behavior-neutral. See also #1827 for the missing timeout on the same introspection call.

yzxcj797:
Fixes #1827
Fixes #1828

## Summary

**Time-bound the introspection POST (#1827).** `_get_token_from_authentication_server()` was the only outbound HTTP call in the library without a deadline — `requests.post` with no `timeout` blocks until the OS gives up, so a stalled authorization server pinned one worker per bearer-token request. Adds `RESOURCE_SERVER_INTROSPECTION_TIMEOUT_SECONDS` (default `5`), matching the existing `CIMD_FETCH_TIMEOUT_SECONDS` / `CLIENT_ASSERTION_JWKS_FETCH_TIMEOUT_SECONDS` convention, plus a `docs/settings.rst` entry.

**Logging correctness (#1828), same area:**

- the non-200 branch of the introspection response is not inside an exception handler, so `log.exception()` appended a meaningless `NoneType: None` line to every record — and attaches an unrelated caller's traceback when invoked inside some outer `except`. Now `log.warning` with `%`-style deferred formatting instead of eager `.format()`;
- an introspection response with `"active": false` returned `None` by falling off the end of the function — now an explicit `return None`;
- `OAuth2ExtraTokenMiddleware` logged a full traceback for every request carrying a bearer token that does not resolve. That is the normal outcome for expired, revoked or bogus tokens (and `DoesNotExist`'s `str()` is empty, so the record had no message), so it is now a single `log.debug`.

## Testing

- `test_introspection_post_uses_configured_timeout` — asserts the POST carries the configured timeout;
- `test_introspection_inactive_token_returns_none` — pins the explicit return;
- updated `test_response_when_auth_server_response_not_200` to expect `WARNING` without the `NoneType: None` suffix (it previously pinned the buggy behavior);
- `test_invalid_bearer_token_logs_at_debug_level` — middleware stays below WARNING on unknown tokens.

Suite: `tests/test_oauth2_validators.py` 68 passed (baseline: 66 — the 5 pre-existing JWT-fixture setup errors are identical with and without this change, they need optional test deps absent here), `tests/test_oauth2_provider_middleware.py` 10 passed + 6 subtests.


dopry:
@yzxcj797 can you rebase this. I just landed a reorg PR that organizes code by authorization_server, client, resource_server. so things are structurally pretty different

dopry:
superceded by PRs on the correct base.

PR #1842, opened 2026-08-23
Add a deploy check for unreachable response types (#1756)

### LINKED ISSUE #1756: Deploy check for OAUTH2_RESPONSE_TYPES_SUPPORTED entries the server can never accept (response_type order-independence)
Follow-up from #1749 (RFC 9700 compliance gates), where Copilot flagged that `validate_response_type()` matches OIDC response types by exact string while the new discovery gating treats them as space-separated sets.

## Background

OIDC's [OAuth 2.0 Multiple Response Type Encoding Practices §4](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#terminology) says a `response_type` containing spaces is compared as an **order-independent set** — `"token id_token"` and `"id_token token"` are equivalent.

DOT (and oauthlib underneath it) do not honor that today, and the fix Copilot suggested — normalizing in `OAuth2Validator.validate_response_type()` — would be dead code:

- oauthlib's `AuthorizationEndpoint.create_authorization_response` routes requests by **exact-string** lookup of `response_type` in its handler registry, which contains only the canonical orderings (`"code"`, `"token"`, `"id_token"`, `"id_token token"`, `"code token"`, `"code id_token"`, `"code id_token token"`, `"none"`).
- An unregistered permutation such as `"token id_token"` falls through to the default handler (the authorization-code grant), which rejects anything that isn't `"code"` with `unsupported_response_type` — **before** DOT's validator can influence routing.

So this is pre-existing behavior, unchanged by #1749: the default `OAUTH2_RESPONSE_TYPES_SUPPORTED` advertises exactly the canonical orderings oauthlib registers, so out of the box discovery matches acceptance. The residual foot-gun is an operator hand-writing a permutation (e.g. `"token id_token"`) into `OAUTH2_RESPONSE_TYPES_SUPPORTED` / `OIDC_RESPONSE_TYPES_SUPPORTED`: discovery then advertises a value the server will always reject. (The #1749 set-based *filtering* is already robust to permutations, so the BCP gates still filter such entries correctly.)

## Proposed follow-up

1. **Deploy check (cheap, this repo):** add a `manage.py check --deploy` warning flagging any entry in `OAUTH2_RESPONSE_TYPES_SUPPORTED` / `OIDC_RESPONSE_TYPES_SUPPORTED` that is not one of the canonical orderings oauthlib's endpoint registry accepts, with a hint giving the canonical form. This closes the discovery-vs-acceptance inconsistency without any behavior change.
2. **(Optional, bigger) true order-independence:** make acceptance order-independent per the OIDC spec, either upstream in oauthlib's dispatch table or by canonicalizing `response_type` (sort the space-separated set) in DOT before dispatch. This is a behavior change (previously-rejected permutations become accepted) and should be considered separately from the check.

Refs: #1749.

aviseth:
Fixes #1756

## Description of the Change

Adds a `manage.py check --deploy` warning (`oauth2_provider.W013`) for entries in
`OAUTH2_RESPONSE_TYPES_SUPPORTED` / `OIDC_RESPONSE_TYPES_SUPPORTED` that the
authorization endpoint can never accept. This is item 1 of the issue only — item 2
(true order-independent dispatch) is a behavior change and is left alone.

oauthlib's `AuthorizationEndpoint` routes by exact-string lookup of `response_type`
in its handler registry, which contains only the canonical orderings. OIDC Multiple
Response Type Encoding Practices §4 defines a multi-valued `response_type` as an
order-independent set, so an operator may reasonably hand-write `"token id_token"`
into a discovery list. That exact string is not registered, so the request falls
through to the authorization-code handler and is rejected with
`unsupported_response_type` — while discovery keeps advertising it as supported.

The check reports that discrepancy at configuration time. Nothing about what the
endpoints accept changes.

Two details worth calling out:

- **The accepted set is derived, not hardcoded.** It is read off the configured
  server's registry via `get_oauthlib_core().server.response_types`, so it is correct
  for `oauthlib.oauth2.Server` (`code`/`none`/`token`), for `oauthlib.openid.Server`
  (the eight OIDC orderings), and for a custom `OAUTH2_SERVER_CLASS` — and it cannot
  drift from oauthlib. If the server or validator is not constructible at check time
  the check returns no messages rather than failing the run.
- **`OIDC_RESPONSE_TYPES_SUPPORTED` is only consulted when `OIDC_ENABLED` is `True`**,
  since the OIDC discovery document that advertises it is not served otherwise, and
  the non-OIDC server registers none of the `id_token` response types — checking it
  unconditionally would emit warnings about a list nothing reads.

A permuted entry gets a hint naming the canonical ordering to use instead; an entry
with no registered equivalent gets a hint listing what the configured server accepts.

## Checklist

- [x] unit-test added
- [x] documentation updated

The last two do not apply: a deploy-time system check has no runtime surface for the
idp/rp demo apps to exercise.


aviseth:
Small process note rather than a ping about the review itself: the `Test` check on this PR is required, but the workflow is sitting in "waiting for approval" because I am a first-time contributor here, so it has not run and cannot report. Merging shows as blocked on a check that is physically unable to complete until someone approves the run.

No rush on the review — just flagging it, since from the outside the PR looks like it is failing CI when it has simply never been allowed to start. Happy to rebase if that would help kick it off.

victorwon2001 (commented):
The new check can turn a malformed setting into a crash instead of reporting a system-check message. I reproduced this on `90cfb061` with:

```python
OAUTH2_PROVIDER = {"OAUTH2_RESPONSE_TYPES_SUPPORTED": [1]}
```

`manage.py check --deploy` raises `AttributeError: 'int' object has no attribute 'split'` at `frozenset(response_type.split())`. Since this function is validating configuration, could it handle non-string entries before canonicalizing them and add a regression test for that case? A `checks.Error` would give operators an actionable result without aborting the entire check command.

dopry:
I ran a multi-agent review of this PR (four independent review dimensions, every finding then reproduced by two adversarial verifiers against oauthlib 3.3.1 and a live Django test setup). The check's detection logic is correct in every scenario tested — default installs emit no W013, the accepted set is right for both `oauthlib.oauth2.Server` and `oauthlib.openid.Server`, W013 is free and next in sequence, and the tests are order-independent. Four fixable items survived verification, all about wording/robustness rather than the core logic:

**1. "Always rejected with `unsupported_response_type`" is wrong for permutations containing `code`** (warning message, docstring, `docs/settings.rst`, CHANGELOG). oauthlib's authorization-code grant gates with a *containment* test (`'code' not in request.response_type`), so a permuted entry like `"token code"` or `"id_token code token"` passes that gate and instead fails the validator's exact-string `validate_response_type` table → the client gets **`unauthorized_client`**. With a custom validator that treats the value as a set, the request isn't rejected at all: it completes as a plain code flow (302 with only a `code`). Since 7 of the 8 non-canonical orderings contain `code`, the stated error code is wrong in most of the cases the check targets. The check *firing* is still right — suggest rewording to "never served as advertised: rejected with `unsupported_response_type` or `unauthorized_client`, depending on the entry."

**2. A non-string list entry crashes `check --deploy`.** The `try/except` only wraps server construction; `response_type.split()` is unprotected, so `["code", 123]` raises an uncaught `AttributeError` and Django's check runner doesn't catch exceptions from checks. An `isinstance` guard lets the entry fall through to the generic W013 warning instead.

**3. The check builds the whole backend and the bare `except` can silently disable it.** `get_oauthlib_core()` also instantiates `OAUTH2_BACKEND_CLASS`, which contributes nothing to `server.response_types` — and with the deprecated `JSONOAuthLibCore` under warnings-as-errors CI, its `DeprecationWarning` becomes an exception the bare `except` swallows, so the check returns `[]` even with a bogus entry configured. Constructing validator + server directly (as `get_oauthlib_core` itself does) avoids this. (The bare `except` itself is fine — verifiers confirmed narrowing it would crash on legitimate custom-server configs.)

**4. Interplay with `COMPLIANT_BCP_RFC9700_IMPLICIT_GRANT`.** With the gate on, `bcp_filter_response_types` drops implicit entries from both discovery documents order-independently — so a permuted implicit entry like `"token id_token"` is never advertised, yet W013 still says it is, and the hint points at `'id_token token'`, which is equally filtered and rejected under the gate. Relevant since the gates flip to `True` in 4.0. Suggest skipping entries the filter would drop.

Also worth a test: the `except → return []` fallback for unconstructible custom server classes is an explicit contract of this PR but has no coverage (lines 387–391 unexercised).

If useful, commit [`c9b240b`](https://github.com/django-oauth/django-oauth-toolkit/commit/c9b240b) on the `claude/django-oauth-toolkit-pr-review-xenuv1` branch implements all of the above (stacked on this PR's head, +79/−18, full suite green) and can be cherry-picked — but feel free to address these your own way instead.

---

aviseth:
Thanks both — all four items are addressed in b66352e. None of them changed when W013 fires, only what it says and how robustly it gets there.

**1. The error code in the message was wrong for most entries.** @dopry is right: `unsupported_response_type` only holds for entries without `code`. oauthlib's authorization-code grant gates on containment (`'code' not in request.response_type`), so a permuted entry containing `code` passes that gate and instead fails the validator's exact-string `validate_response_type` lookup — the client gets `unauthorized_client`. Since seven of the eight non-canonical orderings contain `code`, the old wording was wrong in most of the cases the check targets. Reworded in the warning, the docstring, `docs/settings.rst` and the CHANGELOG to "never served as advertised: rejected with `unsupported_response_type` or `unauthorized_client`, depending on the entry."

**2. Non-string entries crashed `check --deploy`** — @victorwon2001's repro was exact. `response_type.split()` sat outside the `try`, and Django's check runner does not catch exceptions raised by a check, so `OAUTH2_RESPONSE_TYPES_SUPPORTED: [1]` aborted the whole command. A non-string entry now falls through to the generic W013 warning, which is the actionable result you asked for. I kept it a `Warning` rather than an `Error` so it stays consistent with the rest of this check; happy to make it an `Error` if you'd rather malformed config be fatal. Regression test: `test_a_non_string_entry_is_reported_rather_than_raised`.

**3. The bare `except` could silently disable the check.** Confirmed. `get_oauthlib_core()` also instantiates `OAUTH2_BACKEND_CLASS`, which contributes nothing to `server.response_types`, and `JSONOAuthLibCore` raises its `DeprecationWarning` from the constructor — under warnings-as-errors that became an exception the `except` swallowed, returning `[]` even with a bogus entry configured. The check now builds validator + server directly, exactly as `get_oauthlib_core()` does internally. I left the bare `except` as-is per your verifiers' finding that narrowing it breaks legitimate custom-server configs. Regression test: `test_the_check_survives_a_deprecated_backend_under_warnings_as_errors`, which fails on the previous commit.

**4. `COMPLIANT_BCP_RFC9700_IMPLICIT_GRANT` interplay.** Also confirmed, and it mattered more than it first looks: with the gate on, `bcp_filter_response_types()` drops implicit entries from both discovery documents order-independently, so a permuted implicit entry is never advertised and W013 was asserting something false — and its hint pointed at a canonical ordering that is equally filtered. Those entries are now skipped. Hybrid entries contain `code` and so are not implicit, meaning the gate does not drop them and they are still unreachable and still reported; `test_hybrid_entries_are_still_reported_under_the_bcp_gate` pins that boundary so the skip can't widen by accident.

Verification: 1533 passed, 3 skipped on `tests/` with oauthlib 3.3.1 / Django 6.1, ruff clean. The three tests covering items 2–4 were each confirmed failing against the previous commit before the fix went in.

One thing I did not do: the `except → return []` fallback still has no test of its own. It needs a server class that constructs but has no `response_types`, and I wasn't sure whether you'd want that shape pinned or left free. Say the word and I'll add it.

aviseth:
Added the missing fallback test in ff63641 — that was the one loose end from my last comment, and it is also the two lines Codecov flagged.

`test_an_unusable_server_class_disables_the_check_rather_than_failing_it` covers both shapes: a custom `OAUTH2_SERVER_CLASS` that raises from `__init__`, and one that builds but has no `response_types`. Both return no messages with a bogus entry configured, which is the documented contract — the check goes quiet rather than taking down `manage.py check`.

1534 passed, 3 skipped, ruff clean.

PR #1847, opened 2026-08-30
fix(rfc9207): add iss to authorization error redirects too

### LINKED ISSUE #1846: RFC 9207: MUST include iss on error response
**Describe the bug**

[RFC 9207 §2](https://www.rfc-editor.org/rfc/rfc9207.txt):

> In authorization responses to the client, **including error responses**,
  an authorization server supporting this specification MUST indicate
  its identity by including the iss parameter in the response.

I believe that `_add_iss_to_redirect` is called for success responses and not for error response

**To Reproduce**

```python
# repro_iss.py
"""RFC 9207: `iss` is missing from authorization ERROR responses.

django-oauth-toolkit 3.4.1, COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS = True.
"""
import django

settings.configure(
    DEBUG=True, SECRET_KEY="repro", ALLOWED_HOSTS=["*"], ROOT_URLCONF=__name__, USE_TZ=True,
    DATABASES={"default": {"ENGINE": "django.db.backends.sqlite3", "NAME": ":memory:"}},
    INSTALLED_APPS=[
        "django.contrib.auth", "django.contrib.contenttypes",
        "django.contrib.sessions", "oauth2_provider",
    ],
    MIDDLEWARE=[
        "django.contrib.sessions.middleware.SessionMiddleware",
        "django.contrib.auth.middleware.AuthenticationMiddleware",
    ],
    TEMPLATES=[{
        "BACKEND": "django.template.backends.django.DjangoTemplates", "APP_DIRS": True,
        "OPTIONS": {"context_processors": ["django.contrib.auth.context_processors.auth"]},
    }],
    OAUTH2_PROVIDER={
        "SCOPES": {"read": "Read scope"},
        "PKCE_REQUIRED": False,
        "COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS": True,
    },
)

django.setup()

from django.urls import include, path

urlpatterns = [path("o/", include("oauth2_provider.urls", namespace="oauth2_provider"))]

from urllib.parse import parse_qs, urlparse

from django.contrib.auth import get_user_model
from django.core.management import call_command
from django.test import Client
from oauth2_provider.models import get_application_model

call_command("migrate", run_syncdb=True, verbosity=0)

user = get_user_model().objects.create_user("alice", password="pw")
app = Application.objects.create(
    name="repro", user=user,
    client_type=Application.CLIENT_PUBLIC,
    authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
    redirect_uris="https://client.example/cb",
    skip_authorization=False,
)


def authorize(allow):
    client = Client()
    client.login(username="alice", password="pw")
    response = client.post("/o/authorize/", {
        "client_id": app.client_id,
        "redirect_uri": "https://client.example/cb",
        "response_type": "code",
        "scope": "read",
        "state": "xyz",
        "allow": "True" if allow else "False",
    })
    location = response.headers.get("Location", "")
    return location, parse_qs(urlparse(location).query)


for label, allow in (("approve (success)", True), ("deny  (error)  ", False)):
    location, params = authorize(allow)
    print(f"{label}  iss present: {str('iss' in params):5}  {location}")

```

```console
$ uv run --no-project --with django --with 'django-oauth-toolkit==3.4.1' python3 repro_iss.py
django-oauth-toolkit 3.4.1
COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS = True

approve (success)  iss present: True   https://client.example/cb?code=XgeT0…&state=xyz&iss=http%3A%2F%2Ftestserver%2Fo
deny  (error)      iss present: False  https://client.example/cb?error=access_denied&state=xyz

RESULT: BUG - iss omitted from the error response
```

Same with `--with 'django-oauth-toolkit @ git+https://github.com/django-oauth/django-oauth-toolkit@master'`

**Expected behavior**

True and True

**Version**

3.4.1

**Additional context**

Looks like the fix wouldn't be too hard. Let me know if you want a PR.

nityanand123gupta:
Fixes #1846

## Description of the Change

[RFC 9207 §2](https://www.rfc-editor.org/rfc/rfc9207.txt) requires the `iss` authorization-response parameter on **every** authorization response, including error responses — but it was only ever added on the success path.

`OAuthLibCore.create_authorization_response` calls `_add_iss_to_redirect` after a successful `Location` header is built, gated by `COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS`. `AuthorizationServerViewMixin.error_response`, however, builds its redirect URL directly (`redirect_uri + separator + oauthlib_error.urlencoded`) with no `iss` injection at all — so e.g. a resource owner denying access produces an error redirect missing `iss`, even with the setting enabled.

The fix reuses the same `_add_iss_to_redirect` helper in `error_response`, under the same setting gate, only when there's an actual `redirect_uri` to redirect to (a `FatalClientError` has none — it renders an error page instead of redirecting, since the `redirect_uri` itself couldn't be trusted, so there's nothing to add `iss` to in that case).

## Checklist

- [x] PR only contains one change
- [x] unit-test added
- [ ] documentation updated (no docs describe `error_response`'s URL-building directly; the existing RFC 9207 docs already describe the general behavior this now correctly extends to errors)
- [x] `CHANGELOG.md` updated
- [ ] tests/app/idp updated (not a new feature, no IdP demo app changes needed)
- [ ] tests/app/rp updated (same as above)

## Testing

- Added `test_iss_added_to_authorization_error_redirect` and `test_iss_omitted_from_authorization_error_redirect_by_default` to `tests/test_bcp_rfc9700.py`, mirroring the existing success-path tests but with `allow=False` (user denies) to hit the error-redirect path.
- Verified the new "added" test fails without the fix (`git stash` on `mixins.py` alone) and passes with it restored.
- Full suite: `pytest tests/` → 1522 passed, 3 skipped (3 known pre-existing failures unrelated to this change — timezone/import-compat flakiness confirmed present on a clean checkout before my change too).
- `ruff check` and `ruff format --check` are clean on the changed files.

### DIFF
--- a/oauth2_provider/authorization_server/views/mixins.py
+++ b/oauth2_provider/authorization_server/views/mixins.py
@@ -8,8 +8,10 @@ endpoints. It builds on the shared
 
 from django.http import HttpRequest
 
+from oauth2_provider.core.backends_oauthlib import _add_iss_to_redirect
 from oauth2_provider.core.exceptions import FatalClientError
 from oauth2_provider.core.views import OAuthLibCoreMixin
+from oauth2_provider.settings import oauth2_settings
 
 
 class AuthorizationServerViewMixin(OAuthLibCoreMixin):
@@ -95,9 +97,21 @@ class AuthorizationServerViewMixin(OAuthLibCoreMixin):
         redirect_uri = oauthlib_error.redirect_uri or ""
         separator = "&" if "?" in redirect_uri else "?"
 
+        url = redirect_uri + separator + oauthlib_error.urlencoded
+
+        # RFC 9207 §2 requires `iss` on *every* authorization response, including error
+        # responses -- not just successful ones (see _add_iss_to_redirect's other call
+        # site in backends_oauthlib.create_authorization_response). Only add it when
+        # there's an actual redirect_uri to redirect to: a FatalClientError has none (it
+        # renders an error page instead of redirecting, since the redirect_uri itself
+        # couldn't be trusted), and there's nothing to add `iss` to in that case.
+        if redirect_uri and oauth2_settings.COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS:
+            issuer = oauth2_settings.oauth2_authorization_server_issuer(self.request)
+            url = _add_iss_to_redirect(url, issuer)
+
         error_response = {
             "error": oauthlib_error,
-            "url": redirect_uri + separator + oauthlib_error.urlencoded,
+            "url": url,
         }
         error_response.update(kwargs)
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -150,3 +150,4 @@ Apoorv Darshan
 Wojtek
 Rafa Audibert
 ArockiaRajamanickam
+nityanand123gupta
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -170,6 +170,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is unchanged in every case, only subclassing and patching are affected.
 
 ### Fixed
+* #1846 The `iss` authorization-response parameter (RFC 9207, gated by
+  `COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS`) is now added to error redirects as well as
+  successful ones. RFC 9207 §2 requires it on every authorization response, but it was only
+  ever added on the success path in `create_authorization_response`; an error redirect (e.g.
+  the resource owner denying access) built its `Location` header directly in
+  `AuthorizationServerViewMixin.error_response` with no `iss` injection at all.
 * #1828 Two resource-server paths no longer log at the wrong level. A non-200 introspection
   response is an ordinary response, not an exception, so it is logged with `log.warning` instead
   of `log.exception` — the latter appended a meaningless `NoneType: None` line to every such
--- a/tests/test_bcp_rfc9700.py
+++ b/tests/test_bcp_rfc9700.py
@@ -182,6 +182,42 @@ class TestPkcePlainGate(TestCase):
         # RFC 9207: the issuer (matching the metadata issuer) is echoed on the redirect.
         self.assertIn("iss=http%3A%2F%2Ftestserver%2Fo", response["Location"])
 
+    def test_iss_added_to_authorization_error_redirect(self):
+        # RFC 9207 §2 requires `iss` on error responses too, not just successful ones.
+        self.oauth2_settings.COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS = True
+        self.client.login(username="co", password="123456")
+        response = self.client.post(
+            reverse("oauth2_provider:authorize"),
+            data={
+                "client_id": self.application.client_id,
+                "response_type": "code",
+                "redirect_uri": "https://example.org/cb",
+                "scope": "read",
+                "state": "abc",
+                "allow": False,
+            },
+        )
+        self.assertEqual(response.status_code, 302)
+        self.assertIn("error=access_denied", response["Location"])
+        self.assertIn("iss=http%3A%2F%2Ftestserver%2Fo", response["Location"])
+
+    def test_iss_omitted_from_authorization_error_redirect_by_default(self):
+        self.client.login(username="co", password="123456")
+        response = self.client.post(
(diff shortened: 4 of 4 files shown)


oauth2_provider/authorization_server/views/mixins.py
```diff
@@ -8,8 +8,10 @@
 
 from django.http import HttpRequest
 
+from oauth2_provider.core.backends_oauthlib import _add_iss_to_redirect
```
dopry, line 11: the _ indicates a private member. Hoist this to a public helper in core and leave a compatibility shim in place with a deprecation warning for external users, and update all internal code to use the core version.

oauth2_provider/authorization_server/views/mixins.py
```diff
@@ -95,9 +97,21 @@ def error_response(self, error, **kwargs):
         redirect_uri = oauthlib_error.redirect_uri or ""
         separator = "&" if "?" in redirect_uri else "?"
 
+        url = redirect_uri + separator + oauthlib_error.urlencoded
+
+        # RFC 9207 §2 requires `iss` on *every* authorization response, including error
+        # responses -- not just successful ones (see _add_iss_to_redirect's other call
+        # site in backends_oauthlib.create_authorization_response). Only add it when
+        # there's an actual redirect_uri to redirect to: a FatalClientError has none (it
```
dopry, line 105: The comment says "a FatalClientError has none [no redirect_uri]" — that isn't true. OAuthLibCore.create_authorization_response raises FatalClientError(error=error, redirect_uri=credentials["redirect_uri"]) ([core/backends_oauthlib.py:194](https://github.com/django-oauth/django-oauth-toolkit/blob/0f9d49f4abe00bc5c605ed52cf42cd96a3ed2ae2/oauth2_provider/core/backends_oauthlib.py#L194)), and OAuthToolkitError.__init__ copies it onto the oauthlib error.  I suspect the intent is to say it shouldn't redirect, but the comment is ambiguous.  There isn't a need here to mention the success path... that's useful context in the PR.. but doesn't necessarily add value to the code.

oauth2_provider/authorization_server/views/mixins.py
```diff
@@ -95,9 +97,21 @@ def error_response(self, error, **kwargs):
...
+        # renders an error page instead of redirecting, since the redirect_uri itself
+        # couldn't be trusted), and there's nothing to add `iss` to in that case.
+        if redirect_uri and oauth2_settings.COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS:
```
dopry, line 108: redirect_uri is truthy on fatal errors too, so this isn't the right gate. Flow: create_authorization_response raises FatalClientError(error=error, redirect_uri=credentials["redirect_uri"]) (backends_oauthlib.py:194), and on the GET path oauthlib sets redirect_uri on the error itself — in both cases it's the URI that failed validation. So this branch adds iss to the error-page context url for fatal errors, where we deliberately return 400 with no Location (RFC 6749 §4.1.2.1: "MUST NOT automatically redirect the user-agent to the invalid redirection URI").

RFC 9207 §2 applies to "authorization responses to the client" — on the fatal path no response is returned to the client, so no iss is owed there.

Gate on the redirect decision instead:
```
redirect = not isinstance(error, FatalClientError)
if redirect and redirect_uri and oauth2_settings.COMPLIANT_BCP_RFC9700_AUTHZ_RESPONSE_ISS:
```
and reuse redirect for the return value below

dopry (changes requested):
docs/security.rst:155 and docs/settings.rst:646 still say iss is added "to the authorization response" — please add "including error responses" to both. 

Optional but suggested: 
* handle_no_permission() (base.py:519-521) hand-builds the prompt=none → login_required error redirect and bypasses this fix — with the setting on it still emits 302 …?error=login_required&state=abc with no iss. Same spec basis, same two-line injection (or better, a shared error-redirect helper). Fine to defer, but let's file the follow-up before merge so #1846 isn't closed on a partial fix.

* Consider one CHANGELOG sentence noting OIDC Core §3.1.2.6 says error responses "SHOULD NOT" carry other parameters, and RFC 9207's later, specific MUST (plus the advertised authorization_response_iss_parameter_supported flag and the opt-in setting) is why iss is added anyway.

dopry:
@nityanand123gupta  when do you think you'll get a chance to address the PR feedback?

Issue #1843, opened 2026-08-24
3.4.1 update not functional on Oracle


msmitherdc:
## Description

3.4.1's fix for #1687 (PR #1780) introduced a new query in
`OAuth2Validator._save_bearer_token` that combines `select_for_update()`
with an implicit `LIMIT 1` via `.first()`:

https://github.com/django-oauth/django-oauth-toolkit/blob/3.4.1/oauth2_provider/oauth2_validators.py#L1011-L1015

```python
previous_access_token = (
    AccessToken.objects.select_for_update()
    .filter(source_refresh_token=refresh_token_instance)
    .first()
)
```
On Oracle, SELECT ... FOR UPDATE combined with a row-limiting clause has
never been supported by Django's Oracle backend
(django.db.backends.oracle.features.DatabaseFeatures.supports_select_for_update_with_limit = False), so this raises on every refresh_token grant request that goes
through this code path:

django.db.utils.NotSupportedError: LIMIT/OFFSET is not supported with select_for_update on this database backend.
Full traceback:

File ".../oauth2_provider/oauth2_validators.py", line 1014, in _save_bearer_token
    .first()
File ".../django/db/models/query.py", line 1189, in first
    for obj in queryset[:1]:
File ".../django/db/models/sql/compiler.py", line 1611, in execute_sql
    sql, params = self.as_sql()
django.db.utils.NotSupportedError: LIMIT/OFFSET is not supported with select_for_update on this database backend.
This is a regression of a previously-fixed issue
The exact same anti-pattern (select_for_update() + an implicit LIMIT)
was reported against RefreshToken in #866 back in 2021 and fixed in #906
by switching to a plain .get() (no limit) instead of
.filter().first(). #1780 reintroduces the same pattern against
AccessToken.

Suggested fix
AccessToken.source_refresh_token is a OneToOneField, so the filter can
only ever match zero or one row — .get() is a drop-in, limit-free
replacement for .filter(...).first():

```
try:
    previous_access_token = AccessToken.objects.select_for_update().get(
        source_refresh_token=refresh_token_instance
    )
except AccessToken.DoesNotExist:
    previous_access_token = None
```
Environment
django-oauth-toolkit 3.4.1
Django 6.1, django.db.backends.oracle (Oracle 19c)
Reproduced via Django's test client against a refresh_token grant with REFRESH_TOKEN_REUSE_PROTECTION triggering the rotation path.
Versions affected
[...]

PR #1850, opened 2026-09-03
previous_access_token use get rather than first

msmitherdc:
Fixes #
#1843
## Description of the Change
3.4.1's fix for https://github.com/django-oauth/django-oauth-toolkit/issues/1687 (PR https://github.com/django-oauth/django-oauth-toolkit/pull/1780) introduced a new query in
OAuth2Validator._save_bearer_token that combines select_for_update()
with an implicit LIMIT 1 via .first():

https://github.com/django-oauth/django-oauth-toolkit/blob/3.4.1/oauth2_provider/oauth2_validators.py#L1011-L1015

previous_access_token = (
    AccessToken.objects.select_for_update()
    .filter(source_refresh_token=refresh_token_instance)
    .first()
)

AccessToken.source_refresh_token is a OneToOneField, so the filter can
only ever match zero or one row — .get() is a drop-in, limit-free
replacement for .filter(...).first():
## Checklist

- [x] unit-test added
- [ ] documentation updated


dopry:
Overall the fix looks right, but we're missing a reproduction environment. We need tooling to support both local verification and ci verification, in line with what we already do for postgres and mysql to keep the cross db testing consistent.

this is a rough sketch of what goes into setting up a db env for testing and ci IIRC. 

- docker-compose.oracle.yml, the [gvenzl/oracle-free:23-slim-faststart](https://github.com/gvenzl/oci-oracle-free) image looks like a reasonable candidate, CI run time is something we try to keep tight, so it may make sense to use the existing FREEPDB1 database to minimize startup time.  
-  docker/oracle-init - if necessary. I don't know what goes into setting up oracle, you can look at the postgres and mysql variants for guidance.
- tests/oracle_settings.py
- add oracle to the tox.ini
   -  ora23 database factor
   - add test environments ex) py310-dj42-ora23, py312-dj52-ora23, and py314-dj60-ora23
   - add migration environments ex) migrations-dj{42,52,60}-ora23
 - add coverage to test.yml
    - update discovery to output an oracle matrix alongside postgres/mysql.
    - oracle-standalone job
    - add oracle to final success job deps. 
    
 We can forgoe the oracle-pr  variants. We've overloaded PR in the naming of DB environment to mean Primary/Replica. This could either be a precedent PR or included in this one. I'm open to either approach.

msmitherdc:
I plan to add all that, i just need to setup a github runner to connect to our RDS oracle to be available for test runs

dopry:
I'm not in love with depending on an externally managed self-hosted runner for oracle testing.  It doesn't provide infrastructure for maintainers to do the testing locally, and if the runner is offline CI for the project will fail without direct recourse for maintainers to address.  Could this be done in docker containers and with docker compose like we do with postgres and mysql?  I know the  oracle containers tend to be massive multi GB images...  I'd be open to this approach for a bit, but with the caveat it's a short term solution until we can find a more accessible setup for oracle testing.

dopry:
could gvenzl/oracle-xe give us a compatible v21 in docker compose?

PR #1854, opened 2026-09-17
Set `request.user` to `AnonymousUser` for userless tokens in Ninja

brianhelba:
## Description of the Change
An access token may have no user (e.g. with the `client_credentials` grant). `HttpOAuth2` was setting `request.user = None` for these, so any endpoint reading `request.user.is_authenticated` failed with a 500.

This was raised during review of the original PR, where falling back to `AnonymousUser` was the agreed direction:
https://github.com/django-oauth/django-oauth-toolkit/pull/1646#discussion_r2928607812

However, the fix was never applied before that PR was superseded by #1727, so it was lost in the transition.

Since `request.user` can no longer be `None` here, also drop the defensive handling of that in `OAuth2UserOrClientRateThrottle`.

## Checklist

- [x] unit-test added
- [x] documentation updated


brianhelba:
@dopry Please take a look.

PR #1859, opened 2026-09-20
fix(cimd): provision RS256 signing algorithm for CIMD applications

### LINKED ISSUE #1853: CIMD applications don't get algorithm set
**Describe the bug**

Applications created via CIMD don't get an algorithm set.

When I am trying to use DOT to authenticate clients are coming in via FastMCP's `RemoteAuthProvider`, the authentication flow eventually fails at:

```
django.core.exceptions.ImproperlyConfigured: This application does not support signed tokens
```

**To Reproduce**

(The following can all run on localhost)

1. Set up DOT as a basic oauth2/oidc provider with CIMD enabled.
2. Configure FastMCP with `RemoteAuthProvider` pointing to DOT.
3. Add MCP to Claude Code and try to authenticate

**Expected behavior**

Authentication for the newly created CIMD client should complete successfully and not fail when it tries to get the access token.

**Version**

3.4.1

**Additional context**

This seems similar to the issue described here:

https://github.com/gts360/django-oauth-toolkit-dcr/issues/6

I am not certain if it is a good idea to just set an algo or if that has any security implications that I am unaware off.

I am also a bit confused how this is all supposed to work in the end, because it seems like the whole flow will cause the id_token will end up at Claude Code and not at the MCP server, so I am not sure how the MCP server is supposed to identify the user, but I guess that is an orthogonal problem out of scope for DOT.

anushamukka9:
Client Initiated Metadata Documents provision public applications, but the upsert never set `algorithm`, so any OIDC flow that mints an id_token for a CIMD client fails with "This application does not support signed tokens" (#1853).

CIMD clients are always public, which rules out HS256 (it needs a client secret as the HMAC key, and model validation forbids it for public clients). This sets the algorithm to RS256 when the server has `OIDC_RSA_PRIVATE_KEY` configured - the only viable signing setup. The choice is driven by server capability, not by anything in the fetched document, so a hostile or confused metadata document can't influence it. Servers without an RSA key keep the previous behavior (no algorithm) rather than failing metadata validation outright.

Adds two regression tests (algorithm set with a server key; left unset without one). Full CIMD suite passes (108 tests); ruff check and format clean.

Fixes #1853


cfra:
Verified this fixes #1853 for me. :+1:

